PR #252, opened 2014-06-05
Bind openfl.external.ExternalInterface#call to proper this in html5 backend.

ioga:
This change makes ExternalInterface#call in html5 backend work in a similar way to flash backend.

In flash, whenever you call ExternalInterface.call('ns.object.method'), the method will be called with ns.object as 'this'. 
In html5 currently this would be just 'window'.

### DIFF
--- a/backends/html5/openfl/external/ExternalInterface.hx
+++ b/backends/html5/openfl/external/ExternalInterface.hx
@@ -28,6 +28,11 @@ class ExternalInterface {
 		
 		var callResponse:Dynamic = null;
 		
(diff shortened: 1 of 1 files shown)


roughee:
I am getting an error "Cannot read property 'bind' of undefined"
    ```Object.Method.instance().get("myVar")```
ends up 
    ```Object.Method.instance().get("myVar").bind(Object.Method.instance())```

instance() is :
    
    if (!Method.privateInstance) {
        Method.privateInstance = new Method();
    }
    return Method.privateInstance;

Issue #297, opened 2014-07-29
[HTML5] WordWrap not working


starburst997:
Can't get wordwrap to work on TextField on HTML5

Same as this issue on openfl-html5: https://github.com/openfl/openfl-html5/issues/17

Workaround: manually add new line (\n) to text.

jgranick:
Word wrap is working with current versions :smile:

PR #397, opened 2014-11-18
Textfield autoSize and wordWrap fixes (for html5-canvas)

T1mL3arn:
About wordWrap
https://github.com/openfl/openfl-html5/issues/17
https://github.com/openfl/openfl/issues/297
About autoSize
http://www.openfl.org/archive/community/programming-haxe/textfield-autosize-center-broken/

And now it is fixed!

Demo project is here https://github.com/T1mL3arn/TextField-fixes-demo

How it looks in Flash

in flash

How fix looks in Opera and Firefox

opera firefox new code

How it looks without fixes (Firefox throws error, Opera is not)

opera firefox old code

Additionally, in the process of correcting some of the tests have been updated. 
And little fix for text field length getter.

There is also a small "glitch" with changing the text for text field with autoSize=RIGHT/CENTER (I made animation in demo project). If i understood, this occurs because html5 uses requestAnimationFrame method. But text field gets right values for width and x, and after next call of requestAnimationFrame the field will be in the right place.

Also, important note: this fixes works only for simple-formated text, so html text has old behavior!

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasTextField.hx
+++ b/openfl/_internal/renderer/canvas/CanvasTextField.hx
@@ -125,12 +125,7 @@ class CanvasTextField {
 			var negativeSize:Bool = textField.width <= 0 || textField.height <= 0;
 			var noAutoSize:Bool = textField.autoSize == TextFieldAutoSize.NONE;
 			
-			if((textField.__text == null || textField.__text == "") && !textField.background && !textField.border || negativeSize && noAutoSize) {
-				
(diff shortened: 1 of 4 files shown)


jgranick:
I'm running your demo, and most cases seem to be fixed now :smile:

Text has been an uphill battle -- the trick has been a solution that translates the core logic to all our platforms, and doesn't lock too much into the platform-specific renderer code. I think we have it, but it could polish a bit more.

PR #409, opened 2014-11-30
Added Mouse_OVER & MOUSE_OUT support to HMTL5

zorend:
Added Mouse_OVER & MOUSE_OUT support to HMTL5 target by tracking the hit tested object.


jgranick:
Added, thank you! :smile:

PR #426, opened 2014-12-16
Fix negative scale on matrix

starburst997:
To reproduce bug:

var matrix = new Matrix();
matrix.scale( -0.25, 0.25 );
bmp.transform.matrix = matrix;


starburst997:
https://github.com/openfl/openfl/pull/468 seems like a better solution

PR #434, opened 2014-12-23
Somehow fix drawTiles when shape has transformations (scaleX, scaleY), non-zero center point or non-zero x or y coordinates.

restorer:
Without fix:
drawtiles-bug
With fix:
drawtiles-fixed

I use Tilesheet::drawTiles() for my particle engine. Unfortunately, original drawTiles (for html5 target) works correctly only when no scaleX or scaleY applied to parent shape, no center point and shape has x == 0 and y == 0.

This patch is more hack than fix, but it works. Known problem: re-calculated bounds width and height is not correct sometimes, but it doesn't affect visual appearance.

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasGraphics.hx
+++ b/openfl/_internal/renderer/canvas/CanvasGraphics.hx
@@ -633,7 +633,7 @@ class CanvasGraphics {
 								if (rect != null && rect.width > 0 && rect.height > 0 && center != null) {
 									
 									context.save ();
-									context.translate (tileData[index], tileData[index + 1]);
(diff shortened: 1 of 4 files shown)


jgranick:
I need to take a look at this, I made the decision with OpenFL 3 that drawTiles should be affected by the parent transform, it didn't before, but it just makes more sense if it is. This was done for GL rendering, but hasn't been applied for canvas or DOM rendering yet, sounds like that's what you're after in this case

restorer:
I close this PR because I lost this branch and can't merge with upstream.
New PR opened - https://github.com/openfl/openfl/pull/692

PR #435, opened 2014-12-23
- Improved BitmapData.getRGBAPixels speed

soywiz:
- Added Memory.selectTemp that allows using Memory.select without interfering with other code using Memory (the openfl-native project should be updated too)

### DIFF
--- a/openfl/Memory.hx
+++ b/openfl/Memory.hx
@@ -84,6 +84,17 @@ class Memory {
 		
 	}
 	
+	static public function selectTemp (bytes:ByteArray, callback: Void -> Void):Void {
+		var old = gcRef;
+		try {
+			select(bytes);
+			callback();
+			select(old);
+		} catch (e:Dynamic) {
+			select(old);
+			throw e;
+		}
+	}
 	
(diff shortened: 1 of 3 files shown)


soywiz:
Sorry for the delay. I tried that commit in order to get a speedup specially in neko. But later found that neko is reading bytes one by one. So probably that won't be a huge speedup.
Probably the best way to get it faster on neko too would be having critical functions like: premultiply, depremultiply and component swizzling being implemented in C++ and added to the external interface. Maybe with good c++ compilers it would even use SIMD instructions, and would be really fast. With other targets, if they are reading all four bytes at once, and have a good JITter and optimizer, it should be faster.

Regarding with memory API: In flash player you can get and set the memory domain, here you can just set it. If you are using it just in a game, there is no problem. But if you are using it in two different places. For example for audio mixing, and fast-writting to a buffer to send to the gpu, or just decoding an image o decompressing a file, and you mix it. You could end with problems. So probably before doing stuff you should backup the previous memory domain, and then restore it when you are done. So if a library initially doesn't use it and doesn't cause problems with your game, and later tries to optimize stuff, that wouldn't break things.

PR #451, opened 2015-01-04
Fixed -next's BitmapData fill color to work the same as flash and -legac...

gunnbr:
Fixed -next's BitmapData fill color to work the same as flash and -legacy.

Added unit tests to verify the proper behavior of getPixels(), which helped me discover a couple of problems (one of which Joshua already fixed. ;)

The change to BitmapData.hx matches identical functionality in the _v2 BitmapData.hx.

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -161,7 +161,14 @@ class BitmapData implements IBitmapDrawable {
 			this.height = height;
 			rect = new Rectangle (0, 0, width, height);
 			
-			if (!transparent) {
+			if (transparent) {
+
+				if ((fillColor & 0xFF000000) == 0) {				
+					fillColor = 0;
+				}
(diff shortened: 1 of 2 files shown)


gunnbr:
Note that this won't pass until the fix for #207 gets in.

This test also has a failure on neko I need to figure out. (We create a BitmapData with a fillcolor of `0x80112233`, but read back `0x00112233` on neko.) I'm looking for a fix for this.

gunnbr:
I fear the unit tests may still fail on this due to a difference in the way neko legacy works. Then again, I guess I'm not sure if we run unit tests on that or not. Looking on Travis to see...

gunnbr:
Ack. Yeah, this is failing due to that test. Unfortunately, I'm not at that computer at the moment, but I'll try to make a quick patch to disable that test.

PR #465, opened 2015-01-09
Fixed value correction of percentLoaded never applied

Zarmeck:
The code was creating an unused and anonymous boolean instead of assigning the maximum value to percentLoaded.
A small oversight in what looks like dead code.


jgranick:
Thanks! :smile:

PR #474, opened 2015-01-17
SoundJS version update to 0.6. Added IOErrorEvent dispatching to Sound w...

dimanux:
...hile using load method of Sound.


jgranick:
Does this release of SoundJS handle missing audio files better than before?

dimanux:
Hi Joshua,

Sure, I couldn't find any fileerror event in old version of the library.

Thanks!

PR #476, opened 2015-01-18
Html5 support for Tilesheet.TILE_BLEND_ADD and lineStyle alpha parameter 

benmerckx:
When compiling my game to html5 these options were missing.


jgranick:
Thank you :)

PR #477, opened 2015-01-19
implemented a simple drawPath function

shadowofsoul:
using the existent functions, has been implemented a drawPath function. winding is not supported. simple but at least works in most case scenario.


shadowofsoul:
my bad, copied wrong the file, i will submit a new one

PR #478, opened 2015-01-19
implemented a simple drawPath function

shadowofsoul:
using the existent functions, has been implemented a drawPath function. winding is not supported


jgranick:
Thank you! :)

PR #487, opened 2015-01-27
Fix smoothing param for draw() method.

T1mL3arn:
Really?
```Haxe
if (!bool_value)
    object.field = false;
```
And even without this, smoothing is not properly used inside the draw(). Now it is fixed.

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -506,14 +506,6 @@ class BitmapData implements IBitmapDrawable {
 				renderSession.context = cast buffer.__srcContext;
 				renderSession.roundPixels = true;
 				
-				if (!smoothing) {
-					
-					untyped (buffer.__srcContext).mozImageSmoothingEnabled = false;
-					untyped (buffer.__srcContext).webkitImageSmoothingEnabled = false;
-					buffer.__srcContext.imageSmoothingEnabled = false;
-					
-				}
-				
 				var matrixCache = source.__worldTransform;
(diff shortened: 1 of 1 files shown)


T1mL3arn:
Oh, travis failed with this:
```Local results server appears to hang so test reporting was cancelled.```
What am I doing wrong?

CaptainN:
The original code seems to be doing this:

- set the canvas smoothing properties to match `smoothing` bool
- draw using the canvas
- reset the canvas smoothing properties back to `true` if `smoothing` was false. There's an ambiguity there of the kind I don't usually prefer - were the canvas smoothing properties true before they were set to false? The original code assumes so.

Your patch is leaving the canvas smoothing properties `false` which may be affecting any other code that is using the canvas instance.

Note: I haven't looked any further than your patch.

jgranick:
I believe the code is assuming the default is a smoothing canvas. If you had 1000 objects that are all smoothed (which is often the default) it will never set the smoothing attribute, increasing render performance. However, I suppose the downside to this is that if you are using smoothing=false, it will set twice each time.

Perhaps we can maintain a cache of some kind, whether the object has smoothing enabled or not, I'd prefer to touch the DOM as little as possible

CaptainN:
- it looks like IE and opera are missing prefixes:
    - mozImageSmoothingEnabled
    - webkitImageSmoothingEnabled
    - msImageSmoothingEnabled
    - oImageSmoothingEnabled

jgranick:
I think Opera works without a prefix, I guess the IE prefix works in IE 11 only, perhaps this was written against IE 10 at the time

PR #494, opened 2015-02-05
Added missing method to easily reset a Float32Array buffer

kiroukou:
We were missing that very useful method (which used to exist?).
Sounds useful (maybe with @nocompletion)


jgranick:
I can add this in _v2 (since it's hidden) but I expect to not have hidden things like this long-term in the new implementations -- expecting a replacement to the typed arrays for Lime 2 / OpenFL "next" based on abstracts that should perform a lot better :smile:

kiroukou:
Any good other solution would be perfect to us :)
Meanwhile i though that even if quite hidden, that method would be quite
useful for'some people.
Thanks'for'the merge

PR #500, opened 2015-02-12
Conflict security change

kiroukou:
Hello,

This change might sounds stupid, but it prevents an haxe compiler problem when the Main class started with a "create" package.

"Type.getClassFields (::APP_MAIN::))"   was replaced by  "Type.getClassFields (create.Main))"  and so conflicted with create method.

Thomas


jgranick:
Just solved this a different way, give it a try and let me know if it works for you :smile:

PR #508, opened 2015-02-18
Improved performance(memory usage) when drawing large shapes

misohena:
Test code:

```
package;
import openfl.display.Sprite;
class Main extends Sprite {
    public function new () {
        super ();
        var R = 20000; // (2*20000)^2 = 1.6GB!
        graphics.beginFill(0x0000ff);
        graphics.drawCircle(stage.stageWidth/2, stage.stageHeight/2+R, R);
        graphics.endFill();
    }
}
```

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasGraphics.hx
+++ b/openfl/_internal/renderer/canvas/CanvasGraphics.hx
@@ -1,6 +1,7 @@
 package openfl._internal.renderer.canvas;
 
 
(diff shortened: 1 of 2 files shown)


ibilon:

jgranick:
I do not think we can write directly to the final canvas (for performance reasons), since that would require redrawing constantly.

However, I think we should definitely consider doing an optimization that is more direct when doing something simple like a standard rectangle fill.

Thank you for your help :smile:

PR #519, opened 2015-02-25
Add mouse wheel event support

misohena:


drezabek94:
I'm going to port this fix over to Stage.hx (since Application.hx was merged with the Stage), in addition to adding MouseEvent.ctrlKey, .altKey, and .shiftKey support.

PR #520, opened 2015-02-26
Fix MouseEvent.buttonDown

misohena:
If any one of the buttons is pressed, MouseEvent.buttonDown becomes true.


jgranick:
Thanks! It looks like Flash uses the left (primary) mouse button only for the "buttonDown" variable, at least according to the documentation. I simplified to only watch it instead of all three

Issue #528, opened 2015-03-07
MouseEvent's ctrlKey, altKey, and shiftKey flags are always false


drezabek94:
I noticed that openfl.events.MouseEvent.__create doesn't set the ctrlKey, altKey, and shiftKey flags. I am willing to contribute a fix, so if anyone has any advice on how to do so, please let me know.

Unless someone has a better idea, I'll probably add ctrlKey, altKey, and shiftKey flags to openfl.display.Application, and then modify onKeyDown and onKeyUp to set those flags, and finally modify onMouse and MouseEvent.__create to pass those flags along.

PR #530, opened 2015-03-08
Implemented key modifiers and MOUSE_WHEEL

drezabek94:
Implemented ctrlKey, altKey, and shiftKey modifiers for mouse events.
Ported @misohena's mouse wheel event code from Application.hx to
Stage.hx.
This should fix #519, #528


jgranick:
Thank you

I decided to add this a bit differently, but the spirit of what both of you did is in the code. Thank you so much for your help! :smile:

Issue #533, opened 2015-03-12
Bitmap.clone() does not work on html5


T1mL3arn:
Sample code
```Haxe
var bmd1 = new BitmapData(100, 100, false, 0xAA0000);
var bmd2 = bmd1.clone();

var bmp1 = new Bitmap(bmd1);
var bmp2 = new Bitmap(bmd2);

bmp1.y = bmp2.y = 100;
bmp1.x = 100;
bmp2.x = 220;

bmd1.fillRect(new Rectangle(0, 50, 100, 20), 0x00AA00);

addChild(bmp1);
addChild(bmp2);
```
Flash result
![img1](https://cloud.githubusercontent.com/assets/9349164/6616029/6c6b6c6c-c8f5-11e4-9b83-cf6709d3b26b.PNG)
Html5-canvas result
![img2](https://cloud.githubusercontent.com/assets/9349164/6616031/75f12fd8-c8f5-11e4-9308-2af679e19891.PNG)

PR #535, opened 2015-03-15
Realised bitmap.clone() via copyPixels() #533

T1mL3arn:
Related to this #533 . Fix for html5.


jgranick:
Thanks! I just got this fixed in the Lime ImageBuffer implementation -- it was copying by reference, not doing a deep copy :wink:

PR #536, opened 2015-03-17
Android needs precision highp.

player-03:
At least in my tests. I also tried iOS but didn't see any issues there. Maybe it's something to do with Android's implementation of OpenGL?

The issue I'm seeing is, when I try to render objects far away from the origin (when the camera is also far from the origin), their position becomes unreliable, causing objects to appear to vibrate. Presumably, these large numbers are getting rounded thanks to the low precision.

An alternative would be to make a static String variable ("AGLSLParser.glslHeader" or something), and add that to the header. Then if you find you need a specific precision, you can set it yourself.

### DIFF
--- a/openfl/_internal/aglsl/AGLSLParser.hx
+++ b/openfl/_internal/aglsl/AGLSLParser.hx
@@ -20,7 +20,7 @@ class AGLSLParser {
 		var body:String = "";
 		var i:Int = 0;
 		
-		#if html5
+		#if (html5 || android)
 		header += "precision highp float;\n";
 		#end
 		
--- a/openfl/_internal/aglsl/AGLSLParser.hx
+++ b/openfl/_internal/aglsl/AGLSLParser.hx
@@ -20,8 +20,14 @@ class AGLSLParser {
 		var body:String = "";
 		var i:Int = 0;
 		
-		#if (html5 || android)
+		#if html5
 		header += "precision highp float;\n";
+		#elseif (android || ios)
+		if (desc.header.type == "vertex") {
+			
+			header += "precision highp float;\n";
+			
+		}
 		#end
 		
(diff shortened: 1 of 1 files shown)


delahee:
Fyi precision highp float is not always available. There is a flag i don't
remember to do this adequately.

You problem also depend on your no near / far value. You can compress them
to improve z fight / precision jitter.

Gl.

player-03:
According to the language specification, highp must be available to vertex shaders (and according to various others, it's recommended for attributes like position).

I'll change my code to specify highp only for vertex shaders.

delahee:
Oh cool it makes sense.

2015-03-17 6:56 GMT+01:00 player-03 <notifications@github.com>:

> According to the language specification, highp must be available to vertex
> shaders (and according to various others, it's recommended for attributes
> like position).
>
> I'll change my code to specify highp only for vertex shaders.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openfl/openfl/pull/536#issuecomment-82125594>.
>



-- 
David Elahee

player-03:
Is there some reason this hasn't been accepted yet?

jgranick:
I thought we used "mediump" in all our shaders, for performance, and that "highp" failed on some devices.

Perhaps the vertex distinction makes a difference, and perhaps the issue here is that this is used for Stage3D content? Or is this being used by the standard renderer?

player-03:
I know it's used for Stage3D, but I don't know about the standard renderer. And you'd have been the person I'd ask if I wanted to find out...

If you're asking whether I'd be willing to change the `if` statement so that this only applies to Stage3D shaders, the answer is yes. I'm not sure how I'd get that information, though. It doesn't look like `Description` has any information about Stage3D.

PR #547, opened 2015-03-22
fix TextField compile error for haxe 3.2 rc2

eliasku:
https://github.com/openfl/lime/issues/380


jgranick:
Forgot to close this after making the fix, thank you again! :smile:

PR #562, opened 2015-03-27
Speedup (dramatically) canvas BitmapData rendering in DOM

restorer:
Speedup (dramatically) non-image BitmapData (when __canvas is used) rendering in DOM mode

### DIFF
--- a/openfl/_internal/renderer/dom/DOMBitmap.hx
+++ b/openfl/_internal/renderer/dom/DOMBitmap.hx
@@ -81,18 +81,24 @@ class DOMBitmap {
 			}
 			
 			DOMRenderer.initializeElement (bitmap, bitmap.__canvas, renderSession);
+			bitmap.__changeId = bitmap.bitmapData.__changeId - 1;
 			
 		}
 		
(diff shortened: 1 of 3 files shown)


jgranick:
The underlying Lime `Image` type does have a "dirty" flag, perhaps we could use that.

I see a potential problem with multiple bitmaps sharing a single `BitmapData`, perhaps what we need is to use the `image.dirty` flag to check for needing to synchronize, and then perhaps we need to have an integer value that increments every time the `BitmapData` is synch'ed, then the `Bitmap` can check if the `BitmapData` version ID matches, and if not, update it's canvas element

restorer:
Oh, totally forgot that multiple bitmaps can share a single `BitmapData`.
Just committed a fix.

restorer:
> The underlying Lime Image type does have a "dirty" flag, perhaps we could use that

At first attempt I also try to use "dirty" from lime, but it turned out that it is not suitable (I don't remember exact reason). That's why I added "__changeId".

ibilon:

PR #566, opened 2015-03-30
Guard html5 section against double include.

restorer:
This prevents double-include of soundjs, when project depends on openfl,
and also it depends on third-party library, which also depends on openfl.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -79,7 +79,7 @@
 			
 		</section>
 		
-		<section if="html5">
+		<section if="html5" unless="openfl-html5">
 			
 			<set name="openfl-html5" />
 			<haxedef name="openfl-html5" />


jgranick:
Perhaps we should filter in the tools, and only embed one set of dependencies?

I'm concerned about doing the above, because what if you include OpenFL, then a conflicting library, then OpenFL, we want the last included library to force precedence, but it does make sense to keep the includes tidy :wink:

restorer:
I just realized that `ArrayHelper.concatUnique()` / `ArrayHelper.addUnique()` actually didn't work as expected for assets, dependencies, icons, etc... because it compare object references instead of comparing by it actual contents.

I think about 2 methods of fix:

1. if value is a class instance, get it fields using `Type.getInstanceFields()` and than compare with other value using `Reflect.field()`
2. add `Comparable` interface with `equals()` method, implement this interface for Asset, Dependency, etc..., and in `ArrayHelper` if value is instance of `Comparable` call `equals()` methods instead of compare by reference. As a bonus we can remove `key` parameter from `ArrayHelper.concatUnique()` by implementing `Comparable` for `Haxelib` and comparing just `name` in `equals()`

What do you think?

jgranick:
Fixed :smile:

https://github.com/openfl/lime/commit/72558165c1224edac4994a5b270d8bd6e9e6b1ca

PR #569, opened 2015-04-02
Fixes issue lime#395

mrcdk:
Fixes this issue https://github.com/openfl/lime/issues/395


mrcdk:
It doesn't really fix it, I'll try to find a better fix. Don't merge!

PR #604, opened 2015-04-17
getRGBAPixels to use copyChannel which is much faster

tocsick:
Stepping through every pixel in Haxe is very, very slow. Why not use the native lime function lime_bitmap_data_copy_channel which does exactly that but much faster?

My project required loading of 8 2048x2048 textures which took about 25 seconds. Switching to this implementation it now takes about a second.


jgranick:
Thanks! I just manually pulled in the first commit, as the second commit seemed to undo your work somehow. I just added the quick compile fix and it's all in, thank you, this is a great idea.

PR #607, opened 2015-04-17
Fixupsidedownrender :)

larsiusprime:
This fixes upside down results from BitmapData.draw(). It uses matrices to do the transformations.


mrcdk:
oh, I had almost that but the problem was that I got inverted stuff, I didn't think of cloning the matrix (or revert it back) xD good job!

Issue #608, opened 2015-04-17
BitmapData.fillRect before BitmapData.draw() causes inconsistent visuals


larsiusprime:
Flash (Correct) :

![fillrectflash](https://cloud.githubusercontent.com/assets/739064/7211371/c529cc0e-e51f-11e4-96ea-06461fdf1090.png)

Neko+Next (Incorrect) :

![fillrectneko](https://cloud.githubusercontent.com/assets/739064/7211375/cfc17ef0-e51f-11e4-826a-2264f88cd314.png)

Code:
```Haxe
public function new () {
        
        super ();
        
        
        var data = new BitmapData(100, 100, true);
        data.fillRect(new Rectangle(0, 0, 50, 50), 0xFFFF0000);
        data.fillRect(new Rectangle(0, 50, 50, 50), 0xFF0000FF);
        data.fillRect(new Rectangle(50, 0, 50, 50), 0xFF00FF00);
        data.fillRect(new Rectangle(50, 50, 50, 50), 0xFFFFFF00);
        
        var thing = new BitmapData(50, 50, true);
        thing.fillRect(new Rectangle(0, 0, 50, 50), 0xFF000000);
        thing.fillRect(new Rectangle(2, 2, 46, 46), 0x00000000);
        thing.fillRect(new Rectangle(10, 10, 30, 30), 0xFFFFFFFF);
        
        var bitmap = new Bitmap (data);
        addChild (bitmap);
        
        bitmap.x = (stage.stageWidth - bitmap.width) / 2;
        bitmap.y = (stage.stageHeight - bitmap.height) / 2;
        
        var matrix:Matrix = new Matrix();
        matrix.translate(25, 25);
        
        var data2 = new BitmapData(100, 100, true);
        data2.draw(data);
        data2.draw(thing,matrix);
        
        data2.fillRect(data2.rect, 0xFF808080);
        data2.draw(thing, matrix);
        
        var bitmap2 = new Bitmap (data2);
        addChild(bitmap2);
        
        bitmap2.x = bitmap.x + bitmap.width + 10;
        bitmap2.y = bitmap.y;
    }
```

larsiusprime:
Any ideas @mrcdk ?

larsiusprime:
I have done a deep dive into the fillRect() logic, all the way down to the c++ code, and verify it is correct. If you remove the last draw() command, both targets draw the grey square correctly. It is only after you draw ON TOP OF THAT, that neko will revert to the colored background.

[...]

PR #609, opened 2015-04-17
BitmapData: fix for #608

larsiusprime:


jgranick:
Thanks, we criss-crossed paths :smile:

PR #614, opened 2015-04-21
pull out GLTextField to a shared class

james4k:
Allows other renderers to make use of the same TextField graphics code.


james4k:
A note on the CI failure:

There is a test failure for `neko -Ddisable-cffi`, but openfl:master has the same test failure. The difference is that during this build, the tools threw an exception. Not sure of the cause.

```
Called from ? line 1
Called from ApplicationMain.hx line 122
Called from ApplicationMain.hx line 58
Called from /usr/lib/haxe/std/neko/_std/Sys.hx line 118
Uncaught exception - std@sys_exit
```

james4k:
Just pinging. This should be ready to review/merge.

PR #615, opened 2015-04-22
BitmapData: getColorBoundsRect

larsiusprime:


larsiusprime:
Closing this one as it has some unnecessary code now RE: drawSelf and I accidentally opened a second one so we'll go with that instead XD
https://github.com/openfl/openfl/pull/634/files

PR #620, opened 2015-04-24
Font embedding path fix.

bendmorris:
Fixes a "file not found" error when embedding fonts. This fix was previously applied to embedData (https://github.com/openfl/openfl/pull/261, issue https://github.com/openfl/lime-tools/issues/87) but not embedFont.

Should solve https://github.com/HaxePunk/HaxePunk/issues/360


vance:
awsomes

PR #621, opened 2015-04-24
Fixed crash when doubleClickEnabled is true

tocsick:
__lastClickTime is never set so it crashes when used in the double click detection code.


jgranick:
Thanks!

PR #622, opened 2015-04-24
Fixing Texture uploads from BitmapData, and a TextFormat type issue.

Type1J:
This fixes

https://github.com/away3d/away3d-core-openfl/issues/40
https://github.com/openfl/lime/issues/409

, and many problems that I've been having with blending when rendering to textures. BitmapData is assumed to be premultiplied with alpha, and because OpenFL doesn't, this makes sure that that happens.

One improvement would be to check the lime Image for it's premultipled property, but the Image is private, so maybe that should be exposed on BitmapData, or @:access-ed.

The kerning property is correct in OpenFL 3.x, but legacy/lime-hybrid gives me type errors, so I fixed that as well.

This patch fixes a ton for me, so please merge it.

### DIFF
--- a/openfl/_legacy/text/TextFormat.hx
+++ b/openfl/_legacy/text/TextFormat.hx
@@ -13,7 +13,7 @@ package openfl._legacy.text; #if openfl_legacy
 	public var font:String;
 	public var indent:Null<Float>;
 	public var italic:Null<Bool>;
-	public var kerning:Null<Float>;
+	public var kerning:Null<Bool>;
 	public var leading:Null<Float>;
 	public var leftMargin:Null<Float>;
 	public var letterSpacing:Null<Float>;
--- a/openfl/display3D/textures/Texture.hx
+++ b/openfl/display3D/textures/Texture.hx
@@ -51,9 +51,9 @@ class Texture extends TextureBase {
 	
 	
 	public function uploadFromBitmapData (bitmapData:BitmapData, miplevel:Int = 0):Void {
-		
+
 		// TODO: Support upload from UInt8Array directly
-		
+
 		#if openfl_legacy
 		var p = BitmapData.getRGBAPixels (bitmapData);
 		#elseif js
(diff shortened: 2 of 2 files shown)


jgranick:
Thanks!

I do not have a good test case for Stage3D content, so I'm sort of making this change, but I realized that OpenFL legacy will NOT use premultiplied alpha, so we need to be careful not to mulitply in that sense.

I also made a tweak that (I think) will improve performance, because the `image.premultiplied = true` operation goes through a C++ method to quickly hit all the pixels, which (in Neko especially) performs faster

https://github.com/openfl/openfl/blob/master/openfl/display3D/textures/Texture.hx#L53-L82

Please take a look and see if this still works for you. I already committed the above

Type1J:
I'm pretty sure that the change you listed would work. If it's faster, then maybe that should be the way to go. I'd still like the uploadFromUInt8Array() function in this pull, though, since that's about as fast as you could make it.

jgranick:
I already pulled, this was just a tweak to it :smile:

Type1J:
Awesome! Thanks!

Type1J:
Hmm... It looks like that's not working. The blends reverted after the new changes. I'll look into it, and I'll make another pull request when I get it fixed.

Type1J:
Found it. I was using lime-hybrid because I'm rendering TextFields to textures. 3.0 didn't do that correctly last time I checked, but maybe we should just make the jump and fix the TextFields, since everything else seems to work better in 3.0.

Type1J:
Actually, there's quite a few things that need to be fixed because there are different blend issues as well.

After thinking about it, anytime a texture is uploaded from a BitmapTexture, it's assumed to be premultiplied, so the default for legacy should probably be to do the premultiply here.

Type1J:
Also, this is probably a good project to test quickly:

https://github.com/dimumurray/AlphaTest

It's from https://github.com/away3d/away3d-core-openfl/issues/40

Issue #559, opened 2015-03-25
Inconsistent behaviour graphics.lineTo html5


MatthijsKamstra:
code below creates a zigzag in flash/neko

```
		var _line:Sprite = new Sprite();
		_line.graphics.lineStyle(1, 0x000000, 1, true);
		_line.graphics.moveTo( 120, 120);
		_line.graphics.lineTo( 175, 250);
		_line.graphics.lineTo( 230, 80);
		_line.graphics.lineTo( 360, 160);
		this.addChild(_line);
```
flash/neko
![screen shot 2015-03-25 at 12 05 25](https://cloud.githubusercontent.com/assets/907219/6823375/a5b60f62-d2e7-11e4-8c5e-797d5316dc45.png)

But it closes the line in html5
![screen shot 2015-03-25 at 12 05 18](https://cloud.githubusercontent.com/assets/907219/6823376/a5b7f3e0-d2e7-11e4-961d-3a24906c5384.png)

jcward:
Hmm, I'm having the opposite problem now, where lines are being automatically closed in Flash, but not in linux/neko. I did try the snippet above, and it works fine. But my snippet below is inconsistent. Seems like the [Flash graphics API](http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/Graphics.html) says that paths with a fill will auotmatically close, likely the difference here.

```haxe
package;

import openfl.display.Shape;
import openfl.display.Sprite;


class Main extends Sprite {
	public function new () {
		super ();

    function add_shape(x,y)
    {
      var sh = new Shape();		
      var height = 55;
      sh.graphics.beginFill(0xff4444, 0.01);
      sh.graphics.drawCircle(0, 0, height/3);
      sh.graphics.lineStyle(3,0xff4444,0.7);
      sh.graphics.beginFill(0xff4444, 0.3);
      sh.graphics.moveTo(-height/5, -height/5);
      sh.graphics.lineTo( height/5, -height/5);
      sh.graphics.lineTo(        0,  height/6);
      sh.x = x; sh.y = y;
      addChild(sh);
    }

    add_shape(15, 15);
    add_shape(15, 55);
	}
}
```

Flash:
![image](https://user-images.githubusercontent.com/2192439/29153727-68332abe-7d4c-11e7-8d4e-4ce13e102758.png)

Neko:
[...]

Issue #420, opened 2014-12-09
drawing api force closePath for HTML5 target


postite:
this line
https://github.com/openfl/openfl/blob/master/openfl/_internal/renderer/canvas/CanvasGraphics.hx#L107
force the lineTo to be closed even if it's not requested.

bkorcsog:
This is a common scenario for drawing graphics in Flash. The lineTo doesn't close until you close by Drawing to the end point.

Here's an example from wonderful:

http://wonderfl.net/c/q8jR

But in this case they use the drawPath command rather than a sequence of moveTo/lineTo/curveTo events.

PR #628, opened 2015-04-26
Made canvas renderer more consistent with Flash

tocsick:
This commit makes the rendering to canvas a lot more consistent with Flash. See example: http://i.imgur.com/EhkQk68.png 

I started doing this to fix issue #559 but kept going to make it as consistent as possible. Specifically, if no fill is set it won't close paths but also now you can change line style between lineTo commands and have it mirror that of Flash.

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasGraphics.hx
+++ b/openfl/_internal/renderer/canvas/CanvasGraphics.hx
@@ -32,59 +32,316 @@ class CanvasGraphics {
 	private static var TAN22 = 0.4142135623730950488016887242097;
 	
 	private static var bounds:Rectangle;
-	private static var hasFill:Bool;
-	private static var hasStroke:Bool;
-	private static var inPath:Bool;
 	private static var inversePendingMatrix:Matrix;
 	private static var pendingMatrix:Matrix;
-	private static var positionX:Float;
-	private static var positionY:Float;
-	private static var setFill:Bool;
+	
+	private static var hasStroke : Bool;
+	private static var hasFill:Bool;
+	
+	private static var bitmapFill:BitmapData;
+	private static var bitmapRepeat:Bool;
+	
+	private static var graphics:Graphics;
+	
+	private static var fillCommands:Array<DrawCommand>;
+	private static var strokeCommands:Array<DrawCommand>;
 	
 	#if (js && html5)
 	private static var context:CanvasRenderingContext2D;
 	private static var pattern:CanvasPattern;
 	#end
-	
-	
-	private static function beginPath ():Void {
 		
+	private static function playCommands( commands : Array<DrawCommand>, stroke : Bool = false )
+	{
 		#if (js && html5)
-		if (!inPath) {
+		
+		bounds = graphics.__bounds;
+		
+		var offsetX = bounds.x;
+		var offsetY = bounds.y;
+		
+		var positionX : Float = 0;
+		var positionY : Float = 0;
+		
+		var closeGap : Bool = false;
+		var startX : Float = 0;
+		var startY : Float = 0;
+									
+		for (command in commands) {
+					
+			switch (command) {
+									
+				case CubicCurveTo (cx1, cy1, cx2, cy2, x, y):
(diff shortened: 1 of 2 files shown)


tocsick:
Also fixes #420

jgranick:
Looking at your example image, the changes look great. I just tried with a project I'm working with here, and I noticed that the renderer changes do not seem to handle holes inside of fills. For example, when using static SWF text (which comes through as shape commands), "o", "p" and other letters with a hole in the center look filled-in. For some reason, some circular masks I have aren't working properly either.

Are you set up with the SWF library? Would it be hard to try putting some static text in a SWF and reproduce what I'm talking about?

tocsick:
Thanks for testing, I'll have a look later on today and see if I can reproduce / fix the issue.

Hasufel:
@tocsick do you still have the haxe code of the test drawing you mirrored there in png by any chance? If so, could you post it somewhere please?

tocsick:
@Hasufel Sure, it just grabbed snippets from the Internet it's not a comprehensive test or anything...

https://gist.github.com/tocsick/0d5c3fe1a4c928d95feb

Hasufel:
@tocsick thank you. It's perfect as a simple reference test for shape/drawing consistency across new platforms.

tocsick:
@jgranick Okay reproduced the issue and located the problem. See: http://i.imgur.com/tjkI4Pp.png

Rebased and squashed.

jgranick:
Awesome, looks fine here, definitely looks better by your example, we can use this and continue to make improvements if we see anything else. Thanks! :+1:

tocsick:
#420 and #559 can be closed :)

jgranick:
@tocsick Is there any chance you might be interested in helping look at the Cairo software renderer? It works a lot like canvas (many browsers use Cairo under the hood, makes sense!) and in your test above, I have already gotten consistent results with Flash.

Cairo can be used on all the native platforms, so it means that we could have a renderer that's consistent with HTML5 and Flash in those places, we could then mix with the GL renderer to balance quality and performance

Let me know :smile:

tocsick:
I'd love to help. What's the current status and how can I get involved?

jgranick:
Lime has almost complete bindings to the Cairo library. This means we can write code against it, without touching C++. For example, this should look familiar:

https://github.com/openfl/openfl/blob/master/openfl/_internal/renderer/cairo/CairoGraphics.hx

If you use the latest Lime and OpenFL (or both from the source), you can use `openfl test neko --window-hardware=false` to force software, and use Cairo instead of OpenGL.

## Current Status

Using your gist, Cairo looks better (!) than Flash Player in the rounded rectangle, the rest is identical. In practice, it's not quite perfect, but I'm confident it can work. Mozilla, Webkit, even the Linux Flash Player use Cairo.

I bet we could use Cairo for Graphics rendering, then later hook that to a GL texture -- bitmapData.draw, (possibly) cacheAsBitmap, and non-hardware blend modes could all go through Cairo. Consistency in rendering between Flash, HTML5 (which already is quite great now!) and native would be awesome -- and really doors for Flash IDE integration, and continuing to push the envelope.

## What Works

Shapes, Bitmaps, scrollRects and masks (mostly) a lot of the Graphics API and DisplayObjectContainers seem to work

## What Doesn't

bitmapFills don't seem right, drawTiles or drawTriangles is probably related. The position on scaling a scrollRect object is off

The Graphics class is mostly there, but in some edge cases, it seems like the paths get a little narrow?

I think Canvas is implemented using Cairo in a few places, so using it seems naturally similar to the HTML5 Canvas API, with a few differences.

jgranick:
Oh, and if you aren't on Slack chat, would be happy to talk more there if you like :smile:

tocsick:
Cairo is awesome! Love the idea of using it for graphics rendering with the OpenGL renderer. 

I've got the Cairo renderer working pretty consistent with bitmap fills and strokes as well as gradient fills and strokes now. Check it out: http://i.imgur.com/MUh4i8D.png

I'm going to build a comprehensive graphics test pattern now to try and cover more cases. I will post it here for everyone when I'm done. Do you want me to start a new issue for the Cairo renderer?

Never used Slack chat before, how do I get on board?

Issue #636, opened 2015-04-27
BitmapData.draw() when passed a TextField to draw doesn't always work.


Type1J:
This problem was found when porting an Away3D app. Please try looking at the Basic_View stats panel for an example.

Type1J:
I think that this is due to Stage3D not restoring the GL state after a render.

Type1J:
The Stage3D not restoring the state problem may be only some of the issue, but from what I've observed, it may be the sole contributor of this issue.

This is the normal Basic_View Away3D example with the modification that the FPS is passed the color green. The white rectangles are where the green characters should be.
![image](https://cloud.githubusercontent.com/assets/413028/7389128/0cbe1882-ee39-11e4-8454-5ab0e59a0973.png)

This is the same app with

    _view.render();

in the frame callback commented out.
![image](https://cloud.githubusercontent.com/assets/413028/7389221/f648e2ca-ee39-11e4-9290-deaea58d23a0.png)

Although this TextField doesn't render into a BitmapData, this sort of behavior is what I was seeing.

Type1J:
Just to confirm, I changed the render function to

	private static var frame_counter = 0;
	private function _onEnterFrame(e:Event):Void
	{
		_plane.rotationY += 1;
		++frame_counter;
		if (frame_counter < 300) {
		  return;
		}
		if (frame_counter > 600) {
		  return;
		}
		_view.render();
	}

For 3 seconds, I see the white screen with the correct green text.
Then, I see the plane spinning with the incorrect white blocks in place of the text.
3 seconds later, the screen goes white, and the incorrect white blocks render on a white background (you can't see them, but they're there).

I originally said that this was Stage3D not setting all of the GL state back to any values that may have changed, but that seems like it would be a heavy task to keep track of changed GL state just to restore it later.
This could also be viewed as the TextField (spicifically, the openfl._internal.renderer.opengl.utils.GraphicsRenderer) not setting all of the required GL state for its own rendering.

I haven't figured out what state we're talking about, yet. If anybody knows, then please point me in the right direction.

Type1J:
I found that the problem occurs when GL.useProgram (); is called and it isn't set back to the OpenFL renderer program. I'm looking into where that was initially set.

Type1J:
It seems that the problem is with openfl/_internal/renderer/opengl/utils/ShaderManager.hx

[...]

PR #641, opened 2015-04-29
Put BlendModeManager and ShaderManager in a sane state after OpenGLView renders.

Type1J:
For https://github.com/openfl/openfl/issues/636


xerosugar:
Is using the git version of openfl enough in order to get it to work w/ babylonhx? or what else do I need to do to make it use #641?

Issue #594, opened 2015-04-12
Blurry Bitmap rendering on half-pixel positions with -Dnext


Gama11:
```haxe
package;

import openfl.Assets;
import openfl.display.Bitmap;
import openfl.display.Sprite;

class Main extends Sprite
{
	public function new()
	{
		super();
		
		var bitmap = new Bitmap(Assets.getBitmapData("assets/test.png"));
		bitmap.y = 0.5;
		addChild(bitmap);
	}
}
```

This graphic:

![test](https://cloud.githubusercontent.com/assets/2620907/7106728/9db43b82-e14b-11e4-81b7-9b81fd9ea5a4.png)

isn't rendered correctly whenever the position is set to a half-pixel value:

![](http://i.imgur.com/8HriQGS.png)

Not sure if it's strictly limited to that, but anything else I tried (`0.501`, `0.499` etc) looked fine. This also seems to affect `y` way more than `x` (for this particular image anyway), although with `x = 0.5` it also isn't 100% correct:

![](http://i.imgur.com/jJV9mOM.png)

`-Dlegacy` doesn't have this issue.

Tested on neko and windows with Haxe 3.1.3, OpenFL 3.0.1 and Lime 2.3.1.

sruloart:
That seems too specific, like a missing rounding somewhere...

mrcdk:
I can't reproduce the error on latest git

larsiusprime:
Let me try to reproduce as well.

Gama11:
You could also just open the flixel debugger, that's where I discovered the issue originally:

![](http://i.imgur.com/BlwsX1j.png)

larsiusprime:
Tested with neko/release and windows/release with latest stuff (openfl,lime,etc) from Git and latest Haxe RC (3.2 something). 

Can't reproduce, it just renders correctly.

larsiusprime:
WAIT A MINUTE.

I wonder if this is related to this series of ephiphenomenal, hard-to-reproduce things I'm seeing in Defender's Quest (but only for my tester and not me):
https://github.com/larsiusprime/tdrpg-bugs/issues/271

larsiusprime:
This seems to be cropping up in next for me on various flixel demos, especially anything that uses flixel-ui:

Legacy:
![blendmodes_legacy](https://cloud.githubusercontent.com/assets/739064/7143205/bbdf3a9a-e2a3-11e4-9c0e-d1b902e82e77.png)

Next:
[...]

PR #642, opened 2015-04-29
Fix blurry bitmaps

mrcdk:
Fixes issue #594  (hopefully)

this also comes with:
- Optimized shaders
- Added `Bitmap.pixelSnapping` support (not fully)
- `BitmapData` should default to nearest filtering.


mrcdk:
Let's wait until we have confirmation that this fixes it before merging :sweat_smile:

larsiusprime:
You rang?

jgranick:
I just did some testing, and it seems good, if we find ways to improve, we can refine from here :smile:

Issue #646, opened 2015-04-30
Next graphics - getBounds wrt parent (OpenFLNextShapeTest #4)


jcward:
Test case: https://github.com/jcward/OpenFLNextShapeTest

`text.getBounds(this)` returns top-left (instead of proper text .x and .y position) in Next, red outline should be around text.

[<img src="https://raw.githubusercontent.com/jcward/OpenFLNextShapeTest/master/compare.png" width=1168>](https://raw.githubusercontent.com/jcward/OpenFLNextShapeTest/master/compare.png)

jcward:
Tested with latest source, html5 target. This problem still exists, and another problem is that the text is mis-positioned:

[<img src="https://raw.githubusercontent.com/jcward/OpenFLNextShapeTest/master/openfl_next_flash_html5.png">](https://raw.githubusercontent.com/jcward/OpenFLNextShapeTest/master/openfl_next_flash_html5.png)

larsiusprime:
The text thing I might be able to solve.

larsiusprime:
Alrighty:

**Flash**
x = 86.5
y = 3.4
w = 201.7
h = 62.5

**Legacy**
x = 87
y = 5
width = 201
height = 62

**Next**
x = 0
y = 0
width = 201.68
height = 62.569

Looks like Next has properly good width/height values, closer than legacy even. So let me see if I can deduce a proper X/Y for this. I tightened up the renderer a while back so I should be able to figure this out...

larsiusprime:
Oh wow this is a *simple* fix I think. I think you can just plug x,y in to the x,y for that rectangle instead of 0,0: 
https://github.com/openfl/openfl/blob/master/openfl/text/TextField.hx#L1009

larsiusprime:
https://github.com/openfl/openfl/pull/647

larsiusprime:
Try grabbing that branch yourself and see if it fixes everything?

larsiusprime:
I now get these values:

x = 83
y = -0.79
width = 201.68
height = 62.569

So the x,y values don't match up with flash/legacy, although they're close. That said, VISUALLY they look right. Also I ran a test on flash and legacy, and found that the x,y values of the TextField itself exactly matched the x,y values from getBounds, so any discrepancy in x,y values lies elsewhere.

larsiusprime:
**Neko-next**

Before:
[...]

PR #647, opened 2015-04-30
TextField: make the rectangle have the correct x,y coordinates.

larsiusprime:


larsiusprime:
One second, we need to also take care of the matrix transformations:
https://github.com/openfl/openfl/issues/646#issuecomment-97816823

PR #665, opened 2015-05-16
Enhanced Cairo graphics renderer

tocsick:
Pull request for my work on the Cairo renderer.

Already very consistent with flash: http://i.imgur.com/TDRPlvG.png
Runs BunnyMark at low FPS: http://i.imgur.com/XZBGx6g.jpg

### DIFF
--- a/openfl/_internal/renderer/cairo/CairoGraphics.hx
+++ b/openfl/_internal/renderer/cairo/CairoGraphics.hx
@@ -11,7 +11,10 @@ import openfl._internal.renderer.RenderSession;
 import openfl.display.BitmapData;
 import openfl.display.CapsStyle;
 import openfl.display.DisplayObject;
+import openfl.display.GradientType;
 import openfl.display.Graphics;
+import openfl.display.InterpolationMethod;
+import openfl.display.SpreadMethod;
 import openfl.geom.Matrix;
 import openfl.geom.Point;
 import openfl.geom.Rectangle;
@@ -31,8 +34,6 @@ class CairoGraphics {
 	private static var SIN45 = 0.70710678118654752440084436210485;
 	private static var TAN22 = 0.4142135623730950488016887242097;
 	
-	private static var bitmapFill:BitmapData;
-	private static var bitmapRepeat:Bool;
 	private static var bounds:Rectangle;
 	private static var cairo:Cairo;
 	private static var fillCommands:Array<DrawCommand>;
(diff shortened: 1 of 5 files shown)


tocsick:
Wrote a quick text renderer using FreeType's decompose feature. http://i.imgur.com/76WBV4l.png

Still needs some work; lime's binding doesn't support cubic curves properly. If I add cubic support it will look better but I'm guessing at this point it might break something else. Is the way to go or should I be doing it like the OpenGL renderer and using bitmaps?

jgranick:
It looks cubic curves are being approximated using a quadratic:

https://github.com/openfl/lime/blob/f77e9caef401650618845198cb3b13d847b10645/project/src/text/Font.cpp#L242

In the legacy renderer, I think this was the right behavior (cubic curves used to be ignored entirely, which was fun for certain fonts :wink:)

Perhaps now we should preserve cubic curves, since the renderers do have cubicCurveTo support (or at least I think they do?). If they don't, the approximation to a quadratic curve should occur there, not during a decompose

jgranick:
We could keep the bitmap images and use the `TextFieldGraphics` class like GL does, but I suppose that using vectors to render does have certain value to it, particularly if text sizes get enormous

tocsick:
Sorry, I should have said not to merge yet. I just wanted a place to post current work and discuss the changes. 

I was wrong about the cubic curve thing. I just had a mistake in my code. It looks great now http://i.imgur.com/5TWleCD.png

In any case, you are right that lime should preserve the information and let the approximation or proper renderering take place later. I will post a pull request in lime for that.

jgranick:
Things are look much better. I've noticed three things in testing, some rounded rectangles (not sure it's actually a roundRect call, or just a rounded rectangle by geometry) end up looking like a triangle instead (one short side, one long side), I have some masked graphics that are not appearing (might be something with a matrix) or some word wrapping issues with text, but that's to be expected -- text will just continue to evolve (and hopefully we find ways to unify the TextField implementation further for this)

tocsick:
Thanks, I'll keep working on it and hopefully fix those up soon.

The TextField vector implementation is a drop in replacement for the bitmap one. It just sends draw commands to the graphics object. Problem is the OpenGL renderer isn't great at drawing shapes so text is illegible.

jgranick:
Huh, the GL text is supposed to go through Freetype (which renders each requested glyph to a bitmap), before doing a drawTiles render, based on the tilesheet created

jgranick:
However, the best implementation (other than legacy) right now is the HTML5 one, which still uses direct canvas calls, and needs to be made a bit more universal (so wordWrap and other features can be shared)

tocsick:
Yes the GL renderer does that. What I meant was, my vector renderer can be dropped into the GL renderer and it works because it just makes calls to the graphics object. But the GL renderer isn't great at shapes so it isn't readable.

jgranick:
Oh, oh, I see, so we have the ability to render using shapes or using bitmaps. Although I'm tempted to stick with one implementation (to keep maintenance simpler), I wonder if we really should have both (like, say, if the glyph is larger than X points, then just render the shape, etc)

tocsick:
Cairo actually has built in support for FreeType. I was able to get it to render text very nicely yesterday with pixel hinting and it looks beautiful. Almost identical to flash. I will post examples later today. Once we have Cairo being used with the GL renderer I think we will only need this one text rendering path as we can cache to bitmap (or Cairo surface) anyway and it's the same result only nicer.

tocsick:
Here is a comparison of Flash text rendering and the Cairo text renderer. http://i.imgur.com/MsRm1A1.png

PR #680, opened 2015-05-23
Work on Cairo and HTML5 renderers [DON'T MERGE]

tocsick:
Fixes and performance tweaks.

See: https://github.com/tocsick/openfl-test-graphics for some test cases.


tocsick:
I started experimenting with mixing Cairo and OpenGL. Extremely inefficient (and buggy) at the moment but it works.

http://i.imgur.com/1mwtb6D.png

tocsick:
Okay, now we're talking. Getting higher FPS in test cases using this Cairo / OpenGL renderer then the current OpenGL one on its own. We now have fairly consistent rendering across: Flash, HTML5, OpenGL and Cairo.

larsiusprime:
:D

larsiusprime:
Question -- does this rendering path on HTML require webGL I assume?

jgranick:
I think the HTML5 adjustments here are to the default canvas renderer on HTML5, in order to handle WebGL, we will need to do a hybrid WebGL + canvas path rather than OpenGL + Cairo, but it should work similarly :smile:

larsiusprime:
Cool :) Basically the reason I ask is that Flixel's HTML5 support has kind of been in limbo for a while, and if/when the OpenFL HTML5 target is consistent with flash/cpp (or nearly so) I'd like to dive in sometime and see if we can bring it up to speed.

tocsick:
@jgranick is right, I was just working on the default canvas renderer. I've just pushed another commit though that shows that we can indeed do the same thing with WebGL. 

So now we have the following render paths that are all pretty much consistent (albeit not as performant as they could be with some more work):

* Flash
* Cairo
* OpenGL + Cairo
* Canvas
* WebGL + Canvas

I took a screenshot of each here: https://github.com/tocsick/openfl-test-graphics

Currently a call to drawTiles or drawTriangles forces the renderer to use OpenGL / WebGL and anything else gets offloaded to the Cairo / Canvas renderer. This means at the moment you can't mix those two calls with other draw calls. 

Interestingly, WebGL actually out peforms OpenGL significantly. There must be some big bottleneck somewhere in our implementation. 

I want to now unify the text layout engine so that Cairo and Canvas use the same code which will make maintenance simpler. And then of course implement some of the missing functionality (HTML text and ranges).

tocsick:
@jgranick, how should we handle drawTriangles on OpenGL and WebGL? Using 3d acceleration gives much better performance but Flash lets you mix those calls with other drawing operations. It would be very complicated to mix 3d acceleration with the cpu rendering in the graphics class.

I don't really see any other choice then to do exactly what Flash did and leave graphics entirely on the CPU and implement all the 3d acceleration in Stage3D which is rendered below the display list. Adobe obviously did it this way for a reason and this is probably why.

The drawTiles functionality could then sit on top of Stage3D instead of graphics and it would have a performance benefit on the Flash target as well.

jgranick:
I think that `drawTiles

tocsick:
Option 1 would be trickier but the best for the end user. Not having to concern themselves with lower level implementation details to achieve high performance. Things would just work as expected.

jgranick:
I agree, I feel that 1 would be better long-term, probably. Short-term, I think that flagging things for software (and explaining -- don't mix-and-match for HW acceleration) would be an okay way to keep a hybrid solution (while still keeping the implementation simpler)

tocsick:
:( wasn't ready yet. I implemented option 1 but can't merge now due to so many conflicts.

jgranick:
If you have a working branch, I can look at merging manually. I understand the shock, but before progressing further, I needed to do some reorganization.

The TextField, particularly, was threatening to get big and scary and too difficult to understand easily. I'm looking at bringing more functionality into the platform-specific internal TextField classes, over time, I think that we probably will arrive to more conventions (like "Backend"`.measureText`) implemented for DOM, Canvas, Cairo, etc.

I also reverted some changes regarding `CairoShape` and `TextField` but kept all the renderer improvements you made. I also disabled the scaling, there's a few problems that I'd be happy to discuss, did you get an invite for Slack? Maybe we could talk about it there?

PR #685, opened 2015-05-25
Fix incorrect textHeight

vroad:
LINE_HEIGHT + DESCENDER? Looks incorrect.

### DIFF
--- a/openfl/text/TextField.hx
+++ b/openfl/text/TextField.hx
@@ -2347,7 +2347,7 @@ class TextField extends InteractiveObject {
 		//TODO: might need robustness check for pathological cases (multiple format ranges) -- would need to change how line heights are calculated
 		var th = 0.0;
 		for (i in 0...numLines) {
-			th += __getLineMetric(i, LINE_HEIGHT) + __getLineMetric( i, DESCENDER );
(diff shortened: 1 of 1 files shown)


larsiusprime:
Did you check this against the text metrics demo?

tocsick:
He's right that LINE_HEIGHT + DESCENDER is incorrect, LINE_HEIGHT includes DESCENDER already (it was my fault, I was fiddling with it when playing with the text renderer). But the if statement following needs to be updated too. The whole thing should be:

    th += __getLineMetric( i, ASCENDER ) + __getLineMetric( i, DESCENDER );
    if ( i != numLines - 1 ) {
        th += __getLineMetric( i, LEADING );
    }

larsiusprime:
Cool :) Just double checking that the intuition (which of course sounds right) matches the expected visual results.

vroad:
TextMetrics is broken with/without that fix. I'm not sure why text is rendered here.

vroad:
My fix was correct? Then there might be other places which we still need to fix...

tocsick:
@vroad the if statement following the line you fixed needs to be updated to account for leading. At present it is removing non-existing leading on the last line.

vroad:
I've changed that too but TextMetrics is still broken.

PR #695, opened 2015-05-29
take the scrollRect into account when __hitTest

### LINKED ISSUE #694: scrollRect render problem on android and events on html5
Hi,

this report includes 2 bugs (that are related with scrollRect):

1) on html5, if you have a container with scrollRect, and a the content of it, goes into a negative space, anything that would cover it, will not fire any mouse or touch events

2) on android, same settings, content goes in a negative space, anything that would cover it, would not be rendered in the screen ( is not covered by the content itself, the clip of scrollRect works, but at the same time, "erase" all the elements that would be covered by it)

as a example, here is a red sprite of 100x100, inside a scrollRect container of 50x50. the position of the red is  y = -50. the green section is a button of 50x100. so, it means that the red section would cover the green button if the scrollRect didn't exist. in html5 is rendered correctly but events are screw up. in android, the render is incorrect and event's doesn't work either.

html5:

screenshot-16

android:

screenshot-19

here is the project for the test case:

https://drive.google.com/file/d/0B4ORAawuQOukczRPelVBdzV2VTA/view?usp=sharing

mrcdk:
Fixes #694 events

### DIFF
--- a/openfl/display/DisplayObjectContainer.hx
+++ b/openfl/display/DisplayObjectContainer.hx
@@ -696,8 +696,9 @@ class DisplayObjectContainer extends InteractiveObject {
 		
 		if (!visible || (interactiveOnly && !mouseEnabled)) return false;
 		
-		var i = __children.length;
+		if (scrollRect != null && !scrollRect.containsPoint(globalToLocal(new Point(x, y)))) return false;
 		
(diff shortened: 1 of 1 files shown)


shadowofsoul:
i can confirm that this works :)

PR #700, opened 2015-06-03
Allow user to set parameters of HXTelemetry config

jcward:
Allow the user to set parameters of the Telemetry config via a public static config accessor. Features of this implementation include:
- app title still set from Telemetry.hx, telemetry still instantiated from early in framework (Stage)
- error message if set too late (must be a static function initializer, before stage is constructed)

I guess the static initializer is questionable -- perhaps there is another way? How is app title parsed? I don't love the idea of cross-platform locating and parsing a .telemetry.cfg config file (like Flash -- this is one of the most irritating components of telemetry, imo, plus options may be different), or using compiler defines (might cause full recompile when not necessary.) But I'm happy to discuss.

### DIFF
--- a/openfl/profiler/Telemetry.hx
+++ b/openfl/profiler/Telemetry.hx
@@ -3,6 +3,7 @@ package openfl.profiler; #if !flash
 
 #if (cpp && hxtelemetry)
 import hxtelemetry.HxTelemetry;
+import hxtelemetry.HxTelemetry.Config;
 #end
 
 @:allow(openfl.display.Stage)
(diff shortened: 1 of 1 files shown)


jcward:
@jgranick - I want to document how this is done (and what works in legacy vs next.) I see this code in `Telemetry.hx`:

		#if !lime_legacy
		config = (Lib.application.config:Dynamic).telemetry;
		#else
		config = ApplicationMain.telemetryConfig;
		#end

I assume this means that, for legacy, we use the static initializer:

    (static function initTelemetry() { })(  ApplicationMain.telemetryConfig.allocations = true; etc; )

While for next we add to the project xml, e.g.:

     <config:telemetry host="192.168.1.17" allocations="true" />

I'll test this, but is this generally your recollection / intent?

(And note to self -- it looks like the above code was added in 3.3.0).

Thanks,
-Jeff

jgranick:
You should be able to use `<config />` for both, try it :smile:

jcward:
Oh, great, I'll try (last time I think I had `<config:hxtelemetry>` which I believe was the problem? Anyway, I'll try this afternoon.

PR #704, opened 2015-06-03
Improved preloader creation/extension

NemoStein:
The current preloader requires the user to extend `NMEPreloader` and override its methods, but the `NMEPreloader` have it's own, hard-coded, behavior.

This pull request changes `NMEPreloader` to extend `OpenFLPreloader`, which is a empty class (almost an interface that extends `Sprite`).
The only requirement to create a preloader now is 1) extend `OpenFLPreloader` and dispatch an `Event.COMPLETE` when you are done with the preloader.

### DIFF
--- a/openfl/display/Preloader.hx
+++ b/openfl/display/Preloader.hx
@@ -168,15 +168,10 @@ class Preloader extends LimePreloader {
 }
 
 
-#if tools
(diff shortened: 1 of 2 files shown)


ibilon:

PR #709, opened 2015-06-10
updated template. took out uselees space ;)

lewislepton:


templates/project/Source/Main.hx
```diff
@@ -1,19 +1,11 @@
...
 class Main extends Sprite {
 	
-	
 	public function new () {
-		
 		super ();
-		
-		
-		
+		//code goes here
```
Gama11: If you're going for minimal, might as well get rid of this comment too. ;)

lewislepton: very true ;)

PR #710, opened 2015-06-11
added "headerOnly" parameter to Assets.getBitmapData(). Only works wi…

larsiusprime:
…th PNG at the moment.

### DIFF
--- a/openfl/Assets.hx
+++ b/openfl/Assets.hx
@@ -80,7 +80,9 @@ class Assets {
 	 * @param	useCache		(Optional) Whether to use BitmapData from the cache(Default: true)
 	 * @return		A new BitmapData object
 	 */
-	public static function getBitmapData (id:String, useCache:Bool = true):BitmapData {
+	public static function getBitmapData (id:String, useCache:Bool = true, headerOnly:Bool = false):BitmapData {
+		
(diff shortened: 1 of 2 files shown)


MSGhero:
What does header only do?, and you left a trace in there.

larsiusprime:
Basically, there are times when you might want to know the width/height (and other metadata) about an image without having to load the whole thing into memory first. This is the foundation for that sort of feature (I expect Joshua to move the API around rather than pull this in directly)

MSGhero:
Isn't the image already in memory, or does cpp/whatever target handle that differently than flash?

larsiusprime:
Yes, cpp handles it differently. It loads the image from the file system on
demand.

On Thu, Jun 11, 2015 at 7:56 PM, Nick <notifications@github.com> wrote:

> Isn't the image already in memory, or does cpp/whatever target handle that
> differently than flash?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openfl/openfl/pull/710#issuecomment-111317774>.
>



-- 
www.fortressofdoors.com -- Games, Art, Design

jgranick:
Thanks! Solved this with a Lime API, here:

https://github.com/openfl/lime/pull/474

PR #717, opened 2015-06-16
Return true for MultiTouch.supportsTouchEvents

PVince81:
Not sure how to really detect this, but I think it might be fine to always return true for cpp targets which include:
- Windows + touchscreen
- Linux + touchscreen
- Android / mobile ?

(note: touch doesn't seem to work yet on Android, so might want to wait for this)

### DIFF
--- a/openfl/ui/Multitouch.hx
+++ b/openfl/ui/Multitouch.hx
@@ -155,6 +155,8 @@ class Multitouch {
 			return true;
 			
 		}
+		#elseif (cpp)
+		return true;
 		#end
 		
 		return false;


grapefrukt:
This is not fine. As best I can tell the Windows (and other) desktop platforms do not emit TouchEvents? 
How does this always being true make any sense?

PVince81:
Some might have laptop touchscreens. Also as far as I saw it is not easy to detect from the underlying lib to find out what to set this value to.
On Jul 10, 2015 3:33 PM, Martin Jonasson <notifications@github.com> wrote:This is not fine. As best I can tell the Windows (and other) desktop platforms do not emit TouchEvents? 
How does this always being true make any sense?

PR #762, opened 2015-07-24
Stop unneeded Clones

PaulGene:
Thanks to hxScout I noticed this was causing lots of GC


larsiusprime:
Notice any side effects or anything?

 It looks pretty safe since it doesn't modify the transform coming in, if __transform is perfectly encapsulated here and Travis doesn't catch anything it's probably safe to merge.

PR #768, opened 2015-07-28
Fixes text rendering & word wrapping.

larsiusprime:
I'm off by one vertical pixel in ascender metrics for the default text (times new roman) on windows, but can't track it down. The round() functions on LINE_HEIGHT, ASCENDER, and DESCENDER are about trying to dial that in to what flash/legacy does.


larsiusprime:
Summary:

1. Fix basic text rendering (was broken)
2. Implement correct word wrapping that matches flash & legacy (ie, break on whitespace as well as hyphens)
3. Cleanup and refactor potentially useful functions into openfl._internal.text.TextUtil class

PR #784, opened 2015-08-07
Text

larsiusprime:
Post merge catch-up


MSGhero:
Not gonna lie, I'm kinda scared to update openfl after all these changes.

larsiusprime:
Heh :)

Most of these changes just have to do with text rendering and affect the -Dnext branch only. I've been testing pretty rigorously against a lot of demos comparing flash output -Dnext and fixing the discrepancies. After this last set of changes, -Dnext text rendering is closer to flash than it's ever been.

That said we should probably make sure this is very well tested against as many text demos as we can before we put it out on haxelib.

PR #801, opened 2015-08-19
Implemented LineScaleMode

boblucas:
LineScaleMode.NONE can't be implemented consistently the way things are rendered now, so I aproximated it by averaging the x and y scale this works for the common case where scaleX and scaleY are equal, and acts reasonably otherwise.

### DIFF
--- a/openfl/_internal/renderer/cairo/CairoGraphics.hx
+++ b/openfl/_internal/renderer/cairo/CairoGraphics.hx
@@ -49,7 +49,8 @@ class CairoGraphics {
 	private static var pendingMatrix:Matrix;
 	private static var strokeCommands:Array<DrawCommand>;
 	private static var strokePattern:CairoPattern;
-	
+	private static var worldTransform:Matrix;
+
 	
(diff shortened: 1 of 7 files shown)


boblucas:
Is this PR too invasive or did I do something silly? If it's never going to be be merged(or the feedback required to do so is more effort to write than fixing it yourself) I'd rather just close it and make it an issue.

tienery:
It might help to provide a demo of it working so that it does not break anything else in the current branch. It's not that it's not good, just some people can be a bit cautious with contributed code, no matter how small the change.

boblucas:
Thanks for the tip, I'll have some time to make a demo tomorrow.

ibilon:

jgranick:
Since this was first written, we started redrawing graphics when scaled (with an optional define to disable the behavior). This solves a lot of the logic of this PR, in terms of making graphics "scale sensitive"

With this done, we really should use the line scale mode logic, so that we can determine whether to draw lines scaled up or not. I'm going to leave this open, and put it on the wishlist, to remember to revisit this :smile:

devezas:
Hi, is this going to be supported? I really need this for a port...

PR #803, opened 2015-08-20
blend modes: cleaning up a bit, enumerating missing modes, added supp…

larsiusprime:
…ort for lighten & darken.


larsiusprime:
The cairo changes might not really do anything at the moment.

I have verified that drawTiles now supports lighten & darken in flixel. The other blend modes I've added  constants for throughout don't do anything yet, they resolve to default in BlendModeManager for now.

PR #819, opened 2015-09-02
Add osx testing to travis

ibilon:
(hopefully I didn't mess the if)


ibilon:
ok so as expected something went wrong, I forgot about the before install section.

also this will increase by a lot the build time, osx builds take a while to start (one hour in and only 2/10 osx started), there must be a lot less box available than for linux

PR #830, opened 2015-09-12
Drawcommand refactor

larsiusprime:
This is the big one!

The old DrawCommand used dynamic enums which caused all *sorts* of GC-able allocations, as well as other performance bottlenecks. This fixes all that.

Summary:

1. Replace DrawCommand with DrawCommandBuffer
2. Use DrawCommandReader as an abstract view into DrawCommandBuffer
3. Refactor to re-use instance variables where possible to reduce allocations
4. Profit!

This works for Defender's Quest quite well, reduces overall memory use, increases performance, and reduces GC-able allocations.

That said, such a huge change definitely requires some solid testing before it's pulled in. I have made changes to CanvasGraphics too, for instance, and I've checked my code, but I've not run it extensively. We should run this against demos to make sure it lines up and that there aren't regressions.

Testing with HxScout, I got like 40% reduced memory, and decimated the GC allocations. YMMV.


MSGhero:
What's the difference between `splice` and `clear` on arrays? Last week, Joshua said `splice` allocs on Flash but doesn't in cpp, but is `clear` better?

larsiusprime:
Ah, I hadn't checked the flash side just yet. Basically yeah, splice avoids an alloc on cpp for sure, but I'm not sure about flash.

larsiusprime:
@MSGhero : oh I see now. As far as I know there is no clear() function on arrays in the flash api. The clear() command you're seeing for the DrawCommandBuffer is a custom function for that class.

MSGhero:
Ah whoops, ok.

openfl/display/Graphics.hx
```diff
@@ -244,7 +244,7 @@ import js.html.CanvasRenderingContext2D;
 	 */
 	public function beginGradientFill (type:GradientType, colors:Array<Dynamic>, alphas:Array<Dynamic>, ratios:Array<Dynamic>, matrix:Matrix = null, spreadMethod:Null<SpreadMethod> = null, interpolationMethod:Null<InterpolationMethod> = null, focalPointRatio:Null<Float> = null):Void {
 		
-		__commands.push (BeginGradientFill (type, colors, alphas, ratios, matrix, spreadMethod, interpolationMethod, focalPointRatio));
+		__commands.writeBeginGradientFill (type, cast colors, cast alphas, cast ratios, matrix, spreadMethod, interpolationMethod, focalPointRatio);
```
james4k: These casts don't seem safe, and a little surprised that it would build.

Since the public API takes Array<Dynamic>, I don't think we'll be adding any allocs by passing them as Array<Dynamic> all the way through.

openfl/display/DrawCommandType.hx
```diff
@@ -0,0 +1,29 @@
+package openfl.display;
+
+/**
+ * @author larsiusprime
+ */
+
+enum DrawCommandType {
```
james4k: DrawCommandType, DrawCommandBuffer, DrawCommandReader should all be internal APIs, however best that is done. OpenFL users should not use these.

openfl/display/DrawCommandReader.hx
```diff
@@ -0,0 +1,434 @@
...
+		
+	}
+
+}
+
+@:forward
+abstract BeginBitmapFillView (DrawCommandReader) {
+
+	public function new(d:DrawCommandReader)     { this = d; }
+	public var bitmap (get, never):BitmapData;   private function get_bitmap():BitmapData { return cast this.obj(0); }
```
james4k: These View property methods could be inlined, as well as probably all of the DrawCommandBuffer/Reader stuff that are one liners.

openfl/display/DrawCommandReader.hx
```diff
@@ -0,0 +1,434 @@
...
+	
+	public inline function obj (index:Int):Dynamic {
+		
+		return buffer.o[oPos + index];
+		
+	}
+
+}
+
+@:forward
```
james4k: Do we need `@:forward`?

openfl/_internal/renderer/canvas/CanvasGraphics.hx
```diff
@@ -88,7 +91,7 @@ class CanvasGraphics {
 	}
 	
 	
-	private static function createGradientPattern (type:GradientType, colors:Array<Dynamic>, alphas:Array<Dynamic>, ratios:Array<Dynamic>, matrix:Matrix, spreadMethod:Null<SpreadMethod>, interpolationMethod:Null<InterpolationMethod>, focalPointRatio:Null<Float>) {
+	private static function createGradientPattern (type:GradientType, colors:Array<Int>, alphas:Array<Int>, ratios:Array<Int>, matrix:Matrix, spreadMethod:Null<SpreadMethod>, interpolationMethod:Null<InterpolationMethod>, focalPointRatio:Null<Float>) {
```
james4k: Should `alphas:Array<Int>` be `alphas:Array<Float>`?

Although, see previous comment about continuing to use `Array<Dynamic>`.

openfl/display/DrawCommandReader.hx
larsiusprime: I think so, it wasn't compiling without it and someone suggested it and then it worked. TBH I have no idea what it means or does.

openfl/display/Graphics.hx
larsiusprime: I refactored this b/c I had an error with gradients, and along the way I tried to tighten up the types a bit... only to find out the public api is indeed dynamic for em. I think the actual glitch was a messed up variable in the reader anyway, so the typing was probably of no value anyway given it has to be dynamic in the end. I'll revert it in a bit.

openfl/_internal/renderer/canvas/CanvasGraphics.hx
larsiusprime: Yeah this is a mistake. Alphas is indeed a float type, but I'll just put em all back to being Dynamic.

openfl/display/DrawCommandType.hx
MSGhero: private enum with `@:allow`s?

james4k:
Does OpenFL have a fairly comprehensive renderer test suite of some kind that we can use here?

jgranick:
We have the munit tests (under openfl/tests) but this doesn't really cover Graphics, it covers DisplayObject and BitmapData more thoroughly

james4k:
Hmm, I wonder if we can get the final output image for each target to do basic checks on.

larsiusprime:
Do we have a facility in -Dnext to just dump the framebuffer to a BitmapData object? If so then we could just write the file out.

jcward:
> dump the framebuffer to a BitmapData object?
That'd be an interesting feature for lots of testing purposes.

larsiusprime:
@jcward: Most definitely. Flixel's bitmap log is an incredibly handy debugging feature, and being able to know with exact certainty what is being displayed on the screen from code would be incredibly useful. Especially in remote testing situations, as well as on our console efforts.

PR #834, opened 2015-09-13
StageAlignment and StageScaleMode

mumeka:
This is an implementation for StageAlign and StageScaleMode. It works, however we need to talk some issues, there can be better approaches.

(1):
In Flash you can set:
    ```stage.align = "";```
Which makes stage centered, however this behaviour is not documented. And we can not write this line in Haxe, it tells ```stage.align must be flash.display.StageAlign```.

We use StageAlign as an Enum in Openfl, so we can write this code:
    ```stage.align = null;```
But this code crashes in Flash version, it does not allow null.
To make it centered in Flash ouput we need to remove all lines which sets ```stage.align```

(2)
In Flash stage alignment and scaleMode changes view however, it does not edit ```stage.tranform```, changes are hidden.
a
b
c

(2.A.)
Image A uses ```stage.align = StageAlign.TopLeft;```
For Flash transform is:
```stage.transform = (1, 0, 0, 1, 0, 0)```
For Native Openfl with discussed code:
```stage.transform = (1, 0, 0, 1, 0, 0)```

(2.B.)
Image B uses ```stage.align = StageAlign.BottomRight;```
For Flash transform is:
```stage.transform = (1, 0, 0, 1, 0, 0)```
For Native Openfl with discussed code:
```stage.transform = (1, 0, 0, 1, 150, 150)```

(2.C.)
Image C uses ```stage.align = StageAlign.TopLeft; stage.scaleMode = StageScaleMode.SHOW_ALL;```
For Flash transform is:
```stage.transform = (1, 0, 0, 1, 0, 0)```
For Native Openfl with discussed code:
```stage.transform = (2, 0, 0, 2, 0, 0)```

(2.Final)
So this implementation is changing transform and changes are not hidden from developer. Maybe we can hide this changes from transform matrix if it is harmful. 

I will make more tests for this pull request, just wanted to receive ideas.

### DIFF
--- a/openfl/display/Stage.hx
+++ b/openfl/display/Stage.hx
@@ -195,7 +195,7 @@ class Stage extends DisplayObjectContainer implements IModule {
 	 * information, see the "Security" chapter in the <i>ActionScript 3.0
 	 * Developer's Guide</i>.</p>
 	 */
-	public var align:StageAlign;
+	public var align (get, set):StageAlign;
 	
 	/**
 	 * Specifies whether this stage allows the use of the full screen mode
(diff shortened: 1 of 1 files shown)


MSGhero:
(1) Do you think changing `StageAlign` to an `@:enum abstract(String) from String` would work here? There are some other enum classes that cause this kind of issue as well.

mumeka:
It still says ```String should be flash.display.StageAlign``` while compiling Flash target with ```@:enum abstract StageAlign(String) from String```.

You can write ```stage.align = ""``` in native flash with Adobe Flash CC. Because ```Stage.align``` is a String.

However, while Openfl is compiling Flash target, it is not accepting ```String``` but only ```StageAlign``` enum.

MSGhero:
The string you type in has to match the strings defined inside the abstract, and you probably don't have "" defined. Maybe try `to String` too, but I mean it's not that big of a deal.

mumeka:
Ready for merging. We can find a way to fix mentioned issues later, this implementation is better than nothing.

PR #842, opened 2015-09-19
Remove created OpenGL resources from list to avoid memory leak

vroad:
Currently, textures, vertex/index buffers, shader programs created by Context3D is not removed from list when they get disposed. This will slowly lead to memory leak if you create/dispose textures or buffers many times.

Using Array's remove method is not best for performance as it does linear search and creates new array, but I left it untouched for now.

### DIFF
--- a/openfl/display3D/Context3D.hx
+++ b/openfl/display3D/Context3D.hx
@@ -133,7 +133,7 @@ import openfl.Lib;
 	
 	public function createCubeTexture (size:Int, format:Context3DTextureFormat, optimizeForRenderToTexture:Bool, streamingLevels:Int = 0):CubeTexture {
 		
-		var texture = new CubeTexture (GL.createTexture (), size); // TODO use format, optimizeForRenderToTexture and streamingLevels?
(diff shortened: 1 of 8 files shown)


jgranick:
Perhaps the constructors for some of those classes (IndexBuffer, etc) should be private?

jgranick:
Also, do we need to consider looking at making the Context3D API buffered in the future, so we can take commands on an Event.ENTER_FRAME and push to GL as often as we need to? Thanks :smile:

vroad:
According to this, queuing commands from multiple threads and issuing them in a separate thread is good for performance, but it seems pretty complex to implement.
There might be no performance gain if we do that in single thread.

https://www.chromium.org/developers/design-documents/gpu-command-buffer

Constructors of Stage3D classes should be private to prevent users from inappropriately instantiating it, as you said.

vroad:
Next thing we could try to reduce driver overhead is reducing redundant state changes, though I doubt if it has huge effect.
We could also try new features from OpenGL ES 3.0 or above, but I'm still not familiar with that. To do that we should have separate Stage3D implementation for backward compatibility.

PR #861, opened 2015-10-01
Fix BitmapData.paletteMap

azrafe7:
openfl [next] only

This PR changes the logic inside `BitmapData.paletteMap()` to more closely follow as3 behaviour. Namely:
- use pixels from `sourceBitmapData`
- sum values from mapping arrays to compute the resulting color

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -1444,43 +1444,29 @@ class BitmapData implements IBitmapDrawable {
 		var sw:Int = Std.int (sourceRect.width);
 		var sh:Int = Std.int (sourceRect.height);
 		
-		var pixels = getPixels (sourceRect);
-		pixels.position = 0;
+		var pixels = sourceBitmapData.getPixels (sourceRect);
 		
-		var pixelValue:Int, r:Int, g:Int, b:Int, a:Int, color:Int, c1:Int, c2:Int, c3:Int, c4:Int;
+		var pixelValue:Int, r:Int, g:Int, b:Int, a:Int, color:Int;
 		
 		for (i in 0...(sh * sw)) {
 			
-			pixelValue = pixels.readUnsignedInt ();
+			pixelValue = pixels.readUnsignedInt();
 			
-			c1 = (alphaArray == null) ? pixelValue & 0xFF000000 : alphaArray[(pixelValue >> 24) & 0xFF];
-			c2 = (redArray == null) ? pixelValue & 0x00FF0000 : redArray[(pixelValue >> 16) & 0xFF];
-			c3 = (greenArray == null) ? pixelValue & 0x0000FF00 : greenArray[(pixelValue >> 8) & 0xFF];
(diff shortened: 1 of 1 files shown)


azrafe7:
I've adapted code from http://www.splashdust.net/2009/11/bitmapdata-and-palettemap/ to test differences between targets, and used the flash docs to find out how to (hopefully) correctly implement it.

I'd also add that, while the loop is now simplified and a tiny bit faster, the gap in performance with flash is still huge.

I think we need a fast way to access the raw bytes cross-platform for this kind of tricks/effects to be usable in real time. Not sure how to achieve/abstract that (pixel formats, byte buffers, etc.), but would be interested to hear some ideas about this.

The gist is at https://gist.github.com/azrafe7/549ac5e8f1afccd491a7

larsiusprime:
If you want to implement a "fast" version of this you'll probably need to dip down to the CFFI layer... but maybe there is a haxe-level fast direct bytes method available now.

@jgranick or @jcward would know.

azrafe7:
@larsiusprime: Thanks for the quick response.

Going through CFFI would probably work for cpp|neko, but not for html5 (where it seems that `setPixels` is slower than multiple `setPixel` f.e.). Furthermore there's endianness and pixel format issues to take into account I think (not mentioning canvas/webgl/etc.).

Anyway keep the discussion going, as I'd really like to help in improving OpenFl performance regarding pixel manipulation.

larsiusprime:
NP, sorry I don't know about how to do cross platform fast bytes right off the top of my head... I've implemented a few of these BitmapData functions myself in the past, and it would be swell if there was a fast way to access byte data.

azrafe7:
Yes, I seem to remember you implemented `threshold()` (right?), and back then using `Memory` gave impressive results. It seems to not be the case anymore nowadays though, so still eager to find out ways to improve the status quo.

larsiusprime:
That was way back, yeah, on the legacy codebase, and it was only effective for c++. But we're supposed to have native arrays or something like that now, and a lot of work has been done on byte-level access so maybe an API exists now.

jcward:
> NP, sorry I don't know about how to do cross platform fast bytes right off the top of my head...
> But we're supposed to have native arrays or something like that now, and a lot of work has been done on byte-level access so maybe an API exists now.
I don't know either, but I'm interested, for sure!

azrafe7:
@jgranick: this PR could be _slightly_ improved (performance-wise) by accessing bytes directly (i.e. instead of using wrappers, and leveraging conditional compilation), but I'm not sure about endianness as I can only test on win7.

What are your thoughts about this? Is it mergeable as it is?

azrafe7:
I know OpenFl goes through quite a few hoops to give a common interface for bitmap features, but for reference here's a [pure haxejs implementation](http://try.haxe.org/#AFD05) to test agains the gist for html5.

PS: I'm not trying to devalue your work (not at all!), just willing to give feedback and maybe lend a hand, if I'm able to!

ibilon:

PR #873, opened 2015-10-07
Fixed capabilities for tvOS (missing semicolon)

tanis2000:
This also fixes a problem with fonts on iOS 8.2+.
And it also fixes the problem of scaling on Retina displays.


ibilon:
The semicolon is already fixed,
if there's still a font issue with iOS please reopen a new pull request against the develop branch.

PR #697, opened 2015-06-02
Custom shaders

mrcdk:
This PR adds custom shader support for `DisplayObject.shader` and `Tilesheet.drawTiles()` on next.
It also adds support for `DisplayObject.cacheAsBitmap`

**Custom shaders aren't compatible with legacy, Flash or HTML5 canvas.**

I'll make a sample of how to use them later (I'm open to ideas :sweat_smile:)

PR #874, opened 2015-10-07
Shaders

mrcdk:
It's back! More info here: https://github.com/openfl/openfl/pull/697

### What does this PR add?
- Shaders:
    - Now, each `DisplayObject` has a `shader` property.
    - `drawTiles()` also has a new shader argument.
    - The shader will substitute the default shader used by OpenFL *for itself and each children independently* (same happens with `drawTiles()`). This means that the shader will be applied each object and not to the whole object. If you want to apply it as a whole, you can use `cacheAsBitmap`
- `cacheAsBitmap` support:
    - When using `cacheAsBitmap` it will create a cached copy of the `DisplayObject`. The function is the same as the one in Flash but there are a few key differences. In Flash, being a software renderer itself, make sense to apply `cacheAsBitmap` for complex vector graphics. In OpenFL, being a hardware renderer, we already "cache" the vector graphics because we use `cairo` (a vector graphics software renderer) to render the graphics into a texture. Of course, it looks that adding `cacheAsBitmap` support is a waste of time and effort, but, one thing that OpenFL has and Flash doesn't is shaders. Now it makes more sense to have this supported so you can apply a shader to a whole sprite (children included) at once. It also give us support to implement filters! :+1:
    - With `cacheAsBitmap` I've also added a few more properties to control it:
        - `cacheAsBitmapMatrix` will let you control the matrix applied to the cached bitmap when caching it.
        - `cacheAsBitmapSmooth` will let you control the smoothness of the cached bitmap. It only makes sense to modify this value if you are using `cacheAsBitmapMatrix`.
        - `cacheAsBitmapBounds` is a hack, explained later.
- Filters:
    - Added a few filters, they are more powerful than shaders but also slower. Like in Flash, using filters will force `cacheAsBitmap` to `true` on the `DisplayObject`.
    - Be careful using them, each `DisplayObject` with filters will, at least, create 2 copies of itself. Each copy will be a power of two texture because OpenGL ES 2.0 non-power of two texture support isn't the best and create problems. One copy will be the unmodified cached `DisplayObject` and one will be the filtered one. Adding more filters to the array will ping-pong between the two copies. Some filters will also add a third copy because they need the original version for the final composition (for example `DropShadowFilter`) 

### Known issues:
- Filters aren't 1:1 to Flash, some of them are close but they need some fine tuning in their values.
    - `ColorMatrixFilter`: Same output as Flash (as far as I know :sweat_smile:)
    - `BlurFilter`: It doesn't give the same output as Flash with the same settings, but tweaking a bit the `blurX`, `blurY` and `quality` it can come close to it.
    - `GlowFilter`: It has the same *problem* as `BlurFilter` with the blur settings. I don't really know what `strength` does,  `inner` isn't implemented and `knockout` just hides the original bitmap.
    - `DropShadowFilter`: It has the same *problems* as `GlowFilter`.
    - `ConvolutionFilter`: Only 3x3 matrices are accepted.
    - `ShaderFilter`: There is no much to say about this one, it just accepts a `Shader` and applies it to the cached bitmap.
- `cacheAsBitmap` + `drawTiles()` + `scrollRect` on children may create bigger textures than needed. The usual usage of `scrollRect` is to limit the view of an sprite to a certain boundaries, basically if you have a 100x100 sprite and apply a (0, 0, 100, 50) `scrollRect` you will only see the top-half part of the sprite. Because of that, when using `cacheAsBitmap` it would be silly to create a 100x100 texture with half of it empty, so I tried to get the smallest possible texture size to not waste memory. This may not be that big of a deal for you, but what if I told you that HaxeFlixel's Mode demo was creating a 16384x8192 texture when going full screen (1920x1080)? it should create a 2048x2048 texture. Mix that with filters (remember, two copies so two huge textures) and... well, let's say that my gpu drivers weren't happy... Anyway, while I have managed to reduce those huge texture sizes to a more normal and sensible ones, there are still problems... I've been trying to solve this issue since the original PR was made without any success... it may be the silliest thing on the world but I can't see it. As a hack, I've added `cacheAsBitmapBounds` where the you can force the bounds of the texture but it's hard to use and not ideal (I'm probably the only one that knows what values it should have :sweat_smile: )

As always, this PR changes a **LOOOOT** of things and in a lot of critical parts of the rendering pipeline. I can only test that much. I have tested all the openfl-samples without any issue but, again, expect regressions and problems.

I will make a few samples and push them to https://github.com/openfl/openfl-samples in the following days.


larsiusprime:
*Dayum*, son!

jcward:
Freakin sweet. I'll play with this later today.

mrcdk:
Just to give some more information, there are still optimizations to do here, but those optimizations need more internal and risky changes. For example, on top of my head:
- Cache and filter all the things at the start of the render loop. Right now they are cached/filtered as they are needed.
- Reuse framebuffers. Right now the framebuffers are tied to the `DisplayObject`

tanis2000:
Awesome!

johnhattan:
Oooh this new feature fills me with love.

jcward:
Hmm, doesn't work again lime 2.6.7:

`/openfl/display/Stage.hx:23: characters 7-23 : Type not found : lime.ui.Joystick`

lime in git appears to hace that file, pulling it now...

TheSHEEEP:
Sweeeeeet!

PR #876, opened 2015-10-08
Use the full request url in case it includes url variables.

bmfs:
Test case: https://gist.github.com/bmfs/ba27868115055c3c093e

In Flash it works properly
In HTML5, the Loader class strips the request url from any URL variables.

(In the native target, the test case won't work. However that is due to the curl request no having the followlocation flag set to true)

### DIFF
--- a/openfl/display/Loader.hx
+++ b/openfl/display/Loader.hx
@@ -429,16 +429,7 @@ class Loader extends Sprite {
 		
 		worker.doWork.add (function (_) {
 			
-			var path = request.url;
-			var index = path.indexOf ("?");
-			
-			if (index > -1) {
-				
-				path = path.substring (0, index);
-				
-			}
-			
-			BitmapData.fromFile (path, function (bitmapData) worker.sendComplete (bitmapData), function () worker.sendError (IOErrorEvent.IO_ERROR));
+			BitmapData.fromFile (request.url, function (bitmapData) worker.sendComplete (bitmapData), function () worker.sendError (IOErrorEvent.IO_ERROR));
 			
 		});
 		


jgranick:
Thanks! I've fixed this for the HTML5 case (native needs this to not fail in loading local files) though I found an issue that's Neko-specific (Loader regression) that I'm looking into, using your test. Hope to have that ready soon :smile:

jgranick:
I realized that my problem was I needed a clean build on this machine, Neko is running great, though the second image doesn't appear. If you don't mind checking it out, it seems it receives output like:

`<a href="https://mc-avatars.miniclipcdn.com/0-200.png">Found</a>.`

Rather than the image. Is this to do with the user agent of desktop versus HTML5, or is there some other issue? Thanks!

bmfs:
The second image doesn't work in native or neko for another reason. This result of the url is a 302 HTTP Code meaning the resource. For the second image to work in native at least we need to add a allowlocation flag to curl options.

```
curl -i "http://avatars.miniclip.com/players/en/resize.php?uid=1&w=200&h=200"
HTTP/1.1 302 Found
Content-Type: text/html; charset=utf-8
Date: Mon, 12 Oct 2015 19:11:56 GMT
Location: https://mc-avatars.miniclipcdn.com/1-200.png?v=04cdc35d0f02c763729905a115840d81
Server: nginx/1.6.2
Content-Length: 102
Connection: keep-alive

<a href="https://mc-avatars.miniclipcdn.com/1-200.png?v=04cdc35d0f02c763729905a115840d81">Found</a>.

```

jgranick:
I just added FOLLOWLOCATION to our cURL code in OpenFL. Now it properly redirects, but I get an SSL error when it tries to load the image. When I try to manually load the image using the given URL, I get the same result, unless I do so an HTTP address, then it works.

Perhaps the server does not allow a connection unless we include SSL certificates?

Issue #889, opened 2015-10-20
The path is always closed [HTML5]


ngrebenshikov:
Here is the sample of a curve: https://github.com/openfl/openfl-samples/blob/master/features/display/DrawingShapes/Source/Main.hx#L110-L115

It should draw the only curve but it closes the path on HTML5 target. It's OK on the mac target.

Screenshots are below.

HTML5:
![html5](https://cloud.githubusercontent.com/assets/275250/10597593/04da2824-7712-11e5-9da9-20a0bcce30d9.png)

MAC:
![mac](https://cloud.githubusercontent.com/assets/275250/10597595/0b9b2532-7712-11e5-99ba-93388adec5a8.png)

jgranick:
This was fixed some time ago, thank you :grin:

Issue #888, opened 2015-10-20
The filling of complex shapes is wrong [html5, mac-legacy]


ngrebenshikov:
Here is the sample of the rings: https://github.com/ngrebenshikov/openfl-samples/blob/master/features/display/DrawingShapes/Source/Main.hx#L118-L150 

The results:
- HTML5 - works wrong;
- MAC - works OK;
- MAC (legacy) - drawRect doesn't work, drawPath works OK.

The screenshots are below.

HTML5:
![html5](https://cloud.githubusercontent.com/assets/275250/10597439/c4b61c9a-7710-11e5-9168-18d3820e3fc2.png)

MAC:
![mac](https://cloud.githubusercontent.com/assets/275250/10597441/cce51f56-7710-11e5-9c52-97cefb2d3314.png)

MAC with legacy:
![mac-legacy](https://cloud.githubusercontent.com/assets/275250/10597444/d181ed5a-7710-11e5-85b5-a2967328f60c.png)

ngrebenshikov:
It seems it's because that the default winding rule for Canvas2d is nonzero. But it should be evenodd.

ngrebenshikov:
Yep. It works if I change context.fill() to context.fill(CanvasWindingRule.EVENODD).

ngrebenshikov:
I've tested on Crome, FireFox, Safari, IE. It's OK. I'll make a pull request.

PR #890, opened 2015-10-20
Issues #888 and #889 The filling and stroking issues related to the HTML5 target

ngrebenshikov:


ibilon:

jgranick:
Thank you again for your help! I hesitated to include this change because Flash Player really uses both even/odd and non-zero winding in the SWF specification, and I figured that using even/odd in this case would probably fix certain examples, but break others. I'm embarrassed to realize that we are using even/odd in our Cairo renderer. In your sample, the latest OpenFL works properly except for the rings (the other shapes are fine), only on HTML5, since Cairo used a different winding order. Oops.

This should now be resolved 😊

Issue #893, opened 2015-10-21
CanvasTextField render redraw & fix


KnightMearh:
Recently have been developing a javascript game to work inside Ludei / CocoonJS Launcher (free testing app here https://play.google.com/store/apps/details?id=com.ideateca.cocoonjslauncher, also available for iOS).

When running the game inside that environment, we noticed TextFields were not refreshing and would generate a stamp effect every time the field changed. This was also causing the game performance to slow down. I believe this framework converts Canvas to Canvas+ and this changes how rendering works a little. We have found a fix for it, and this is probably good for browsers anyway (they still perform as expected with text behavior).

Please insert this into CanvasTextField.hx:170...

<code>context.clearRect( 0, 0, graphics.__canvas.width, graphics.__canvas.height );</code>

This causes the text fields to clear properly before redrawing the new text whenever they change. I don't believe it to be more expensive or anything since I think you guys were assuming setting the width to itself would also invalidate it, but this will force it to work on additional systems.

Leander1P:
Submitted a pull request here to foster discussion: #902

jgranick:
Oops, linked the wrong issue

https://github.com/openfl/openfl/commit/d9ac19c407c952fdc3c28202fc98ec481fbfe32e

Thanks for all of your work and help, please let me know if this makes a difference, and if there are other cases we need something similar.

Have a great day :grin:

PR #902, opened 2015-10-23
Add explicit clearRect calls to CanvasTextField when dirty.

Leander1P:
This is primarily a fix for Ludei's Canvas+ impl, where we'd see updated text render over old text.

Note that it has potential tradeoffs.  From comments:

// On most canvas impls (but not all, see Ludei's Canvas+ as of approximately 2015-10),
// setting width or height above (even if same) will be enough to both reset the context
// and clear the canvas bitmap.

// But this behavior doesn't appear to be required by either of:
//
//   http://www.w3.org/TR/2dcontext/
//     or
//   http://www.w3.org/TR/html5/scripting-1.html#attr-canvas-width
//
// ...except for a link from "set bitmap dimensions" to a dead 2dcontext2 standard.

// A description of this behavior *does* appear in the the WHATWG living HTML standard here:
//
//   https://html.spec.whatwg.org/#concept-canvas-set-bitmap-dimensions

// There's some evidence that avoiding the width/height set and only doing clearRect
// where appropriate may be a performance win, see:
//
//    http://simonsarris.com/blog/346-how-you-clear-your-canvas-matters

// Playing it safe and adding the explicit clearRect, since it fixes Canvas+.

Please note this may be a performance regression for normal browsers (although see the above comment about a possible optimization opportunity related to this).

Note that other _internal/renderer/canvas/*.hx renderers may warrant similar changes; these have not been audited.

Reviewed in-house by Danielle Cerniglia and tested extensively in a current app.

Submitting a pull request here to foster discussion re the above points.

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasTextField.hx
+++ b/openfl/_internal/renderer/canvas/CanvasTextField.hx
@@ -167,6 +167,30 @@ class CanvasTextField {
 					graphics.__canvas.width = Math.ceil (bounds.width);
 					graphics.__canvas.height = Math.ceil (bounds.height);
 					
+					// On most canvas impls (but not all, see Ludei's Canvas+ as of approximately 2015-10),
+					// setting width or height above (even if same) will be enough to both reset the context
+					// and clear the canvas bitmap.
+					// 
+					// But this behavior doesn't appear to be required by either of:
+					//
+					//   http://www.w3.org/TR/2dcontext/
+					//     or
+					//   http://www.w3.org/TR/html5/scripting-1.html#attr-canvas-width
+					//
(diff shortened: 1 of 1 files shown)


Leander1P:
Linking back to issue #893; holler if there was a better way to associate the PR directly with that issue.

Leander1P:
Ugh, screwed up another rebase.  Investigating again, sorry.

Leander1P:
Ah, that appears to be because this targeted openfl:master rather than openfl:develop, and I just rebased over develop.  Will restore it to being over master, but what's the proper protocol here?  Close this PR and open a new one?

ibilon:
I don't think you can change to branch of a pull request, so yeah close and open another.

Leander1P:
Thanks, closing and will reopen another momentarily.

PR #911, opened 2015-10-29
Fix closing line joints on cairo paths :), Fixes #909

### LINKED ISSUE #909: Error with drawing lines in Cairo
This regression is still happening with line drawing:

limemiter3

There's two things wrong here:
1) Looks like the final joint is not rendering correctly on native targets (
This causes a bug in defender's quest where arrows that I draw for tutorials have a "bite" taken out of the arrowhead's tip)

2) The shapes are drawn in a different position than in Flash (though html5 and neko seem lined up to one another:)

linemiter3_lines

The red line is for flash, the blue is the same lines from flash offset to line up with neko/html5. Notice that although the neko/html5 shapes are lined up with one another, that the spacing is off from flash as well, not just the overall offset position -- if the spacing was the same, the blue line would be on the left edge of the left triangle rather than cutting through it.

larsiusprime:
Fixes #909

### DIFF
--- a/openfl/_internal/renderer/cairo/CairoGraphics.hx
+++ b/openfl/_internal/renderer/cairo/CairoGraphics.hx
@@ -555,6 +555,7 @@ class CairoGraphics {
 					hasPath = true;
 					cairo.lineTo (c.x - offsetX, c.y - offsetY);
 					
+					closeGap = true;
 					positionX = c.x;
 					positionY = c.y;
 				


larsiusprime:
@jgranick any reason not to pull this yet?

ibilon:
@larsiusprime if this is still needed could you push your commit to the develop branch?

PR #928, opened 2015-11-10
Update to bring XMLSocket in line with _legacy functionality

sumo961:


jgranick:
This pull request seems to be conflicted. Would you mind testing if the current OpenFL XMLSocket class seems to solve the problem? Thank you! :smile:

PR #944, opened 2015-11-24
Bmpdata frombytes fix

larsiusprime:
There's an error in hxcpp that makes flixel fail to compile without this fix to BitmapData.hx. The problem is these two lines in FlxAssets.hx:

```
@:keep @:bitmap("assets/images/ui/button.png")
class GraphicButton extends BitmapData {}
```

We found two fixes -- remove in inline in ByteArray.readUnsignedByte, or pull it out to a local variable and drill out of the abstract. We opted for the latter so we could keep the inline.

Big thanks to @mauvecow for helping on this one.

### DIFF
--- a/openfl/display/Bitmap.hx
+++ b/openfl/display/Bitmap.hx
@@ -127,7 +127,7 @@ class Bitmap extends DisplayObject {
 	
 	@:noCompletion private override function __hitTest (x:Float, y:Float, shapeFlag:Bool, stack:Array<DisplayObject>, interactiveOnly:Bool, hitObject:InteractiveObject):Bool {
 		
-		if (hitObject == null || !hitObject.visible || __isMask || bitmapData == null) return false;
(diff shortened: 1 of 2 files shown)


jgranick:
Do you still need this change, or does it work as-is currently?

larsiusprime:
No, looks like I don't need this change anymore with latest Lime + OpenFL.

However, there are other issues. FileReference seems to be broken, for instance. I've put that and other issues I found in this PR here:
https://github.com/openfl/openfl/pull/952

Also in Flixel we have an issue where we do stuff like this:
```
@:file("assets/images/ui/virtual-input.txt")
class VirtualInputData extends ByteArray {} 
```

Which is not possible since ByteArray is now an abstract rather than a class. My "get it working hack" is this:

```
@:file("assets/images/ui/virtual-input.txt")
class VirtualInputData extends #if lime_legacy ByteArray #else ByteArrayData #end {} 
```

But if there's a "proper" way to do this I'm all ears.

PR #952, opened 2015-12-05
Fixes FileReference to work with new ByteArray

larsiusprime:
This won't compile with ByteArray as an abstract, so I changed as littlle as I could to get it working again


larsiusprime:
Also, I bumped the version # in the haxelib.json. Even if you're not ready to do a release yet, with the changes to ByteArray we really need to bump the version internally so that we can put in proper compiler constant version checks so that flixel can be made compatible with both this upcoming release and older versions from before the change

Issue #913, opened 2015-10-30
BitmapData.perlinNoise does not work on html5 target (no output)


cemir:
Here is the code used :+1:

```
            var bmpFrac:BitmapData = new BitmapData(500,500, false, 0);
            bmpFrac.perlinNoise(8.7, 7.9, 2, 1516468464, false, false, 1, true);

            bmpFrac.setPixel(10,10,0xFF00FF00);

            var testPic : FlxSprite;
            testPic = new FlxSprite();
            testPic.loadGraphic(bmpFrac);
            add(testPic);
```

in html5 it gives a black picture with a colored pixel at 10,10
in cpp it gives a "cloudy" picture, as expected.

openfl version : 3.4.0
lime : 2.7.0

Regards

ICTylor:
Hi, Perlin noise doesn't work for me in any platform except Flash. I'm using this code
```
class Main extends Sprite 
{
	private var _bitmapData:BitmapData = new BitmapData(400, 400);
	private var _bitmap:Bitmap;

	public function new() 
	{
		super();
		_bitmapData.perlinNoise(400, 400, 6, 0, false, false);
		_bitmap = new Bitmap(_bitmapData);
		this.addChild(_bitmap);
	}
}
```
In my case it draws the right image in Flash, in any other platform it draws a white square.

Haxe version: 3.2.1
OpenFL version: 3.4.0
Lime: 2.7.0
Compiling on a Windows 8.1 x64 machine

larsiusprime:
I don't think we've implemented Perlin noise yet, I originally assumed that this was because the algorithm was patented, but it looks like it might be okay:

http://computergraphics.stackexchange.com/questions/19/does-the-original-perlin-noise-ever-have-any-advantage-over-simplex-noise

MSGhero:
Simplex noise isn't patented for 2D, and I don't think perlin noise is either. I'm not sure which one bmd uses, but simplex noise is faster (I have it coded already and can post it).

larsiusprime:
@MSGhero : cool, if you want, here is where you would add it:

https://github.com/openfl/openfl/blob/master/openfl/display/BitmapData.hx#L1568

larsiusprime:
More information if it helps:
http://stackoverflow.com/questions/8467685/how-is-perlin-noise-in-flash-implemented

MSGhero:
[...]

PR #953, opened 2015-12-06
Added noise calculation function

gamepopper:
I have noticed that the functions for noise in the BitmapData class still haven't been implemented yet. I was originally approached to by Joshua Granick to write implementations for both noise and perlin noise, and this was the function that I wrote.

It takes into account the channel options and high and low colour options as well.

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -987,9 +987,9 @@ class BitmapData implements IBitmapDrawable {
 		var blueChannel:Bool = ((channelOptions & ( 1 << 2 )) >> 2) == 1;
 		var alphaChannel:Bool = ((channelOptions & ( 1 << 3 )) >> 3) == 1;
 		
-		for (y in 0...noiseTestData.height)
+		for (y in 0...height)
 		{
-			for (x in 0...noiseTestData.width)
+			for (x in 0...width)
 			{
 				//Default channel colours if all channel options are false.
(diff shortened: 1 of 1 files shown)


MSGhero:
Does this have the same results as flash's implementation?

gamepopper:
The output should be virtually identical, here's a render test using a basic OpenFL application. The left is a bitmap of 100,100 using the BitmapData's current noise function, and on the right is a bitmap of the same size using my implementation.

Test Preview

MSGhero:
I mean as far as the exact noise function used, which is different in those two cases. In the discussion in #913, we figured we could either find and implement the exact flash function for noise and perlin, or we could just override flash's default behavior for those two functions only.

gamepopper:
Problem I had with implementing Perlin noise was that the parameters made it much more complex. At least with this noise implementation, it also works on cpp targets (although HTML5 appears to have some issues with the setPixel32 function).

CPP Output Comparison

MSGhero:
In that issue, someone found flash's implementation written out already, but there was some kind of issue. I don't know if it's implemented correctly in openfl legacy which is where it was tested.

Anyway, not for me to decide.

jgranick:
This looks like a clean step forward. This is a good discussion to continue to have, regarding reproducibility and performance, but I'm going to pull since its an increment forward from a missing implementation (for sure!)

PR #956, opened 2015-12-10
fixes #934, adds optional html credentials to URLRequest

### LINKED ISSUE #934: URLLoader should expose requests withCredentials for html5
This is required for cross domain applications, currently I'm editing URLLoader sources by hand

er453r:
fixes https://github.com/openfl/openfl/issues/934

### DIFF
--- a/openfl/net/URLRequest.hx
+++ b/openfl/net/URLRequest.hx
@@ -13,7 +13,10 @@ import openfl.utils.ByteArray;
 	public var requestHeaders:Array<URLRequestHeader>;
 	public var url:String;
 	public var userAgent:String;
-	
+
+	#if (js && html5)
+	public var withCredentials:Bool = false;
+	#end
 	
 	public function new (inURL:String = null) {
 		
--- a/openfl/net/URLLoader.hx
+++ b/openfl/net/URLLoader.hx
@@ -89,7 +89,7 @@ class URLLoader extends EventDispatcher {
 	public function load (request:URLRequest):Void {
 		
 		#if (js && html5)
-		requestUrl (request.url, request.method, request.data, request.formatRequestHeaders ());
+		requestUrl (request.url, request.method, request.data, request.formatRequestHeaders (), request.withCredentials);
 		#else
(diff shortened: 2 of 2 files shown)


openfl/net/URLRequest.hx
```diff
@@ -13,7 +13,10 @@ import openfl.utils.ByteArray;
 	public var requestHeaders:Array<URLRequestHeader>;
 	public var url:String;
 	public var userAgent:String;
-	
+
+	#if (js && html5)
```
Gama11: Aren't `js` and `html5` the same, making this redundant?

er453r: Thats the way most of the openfl sources do it, so I did it the same.

And I think int is not the same, since `js` (node) in not `html5` (browser)

larsiusprime: Correct, ```js``` is not the same as ```html5```

Properly understood, ```js``` is similar to ```cpp``` whereas ```html5``` is similar to ```windows```. 

IE, ```html5``` implies `js`, but `js` does not imply `html5`. 

In practice you can have JS+HTML5 (ie, a browser app), or you can have JS+node (a non-browser-app).

jgranick: `#if js` could be Node.js. In the past, our documentation was generated using C++ and -Dhtml5, since a true JS build did not unify. However, now our documentation does use real Flash and JS builds

ibilon:

jgranick:
Does using `crossOrigin = "Anonymous";` help with the cross-origin permissions issue you are having?

Should we enable `withCredentials` by default?

I want to consider a way to keep the API in `lime.net.HTTPRequest` simple, but I want to find a way to support the use-case you need without resorting to manual patches. Thank you for your help :)

jgranick:
Since we added official Adobe AIR support, it opened up a good way for us to support `urlRequest.manageCookies` (doing `withCredentials` under the hood) as a real property on most targets, and a dummy property on Flash Player, so now `withCredentials` is officially available from the OpenFL API. Thank you again for your help!

PR #958, opened 2015-12-10
add "fullscreenHotkey" to project.xml window config

larsiusprime:
default value is true, if true, alt+enter for windows/linux (and ctrl+meta+F for mac) will fullscreen the app. If false, it will not when such keystrokes are pressed.

the config value is "fullscreenHotkey", and the value in xml is "fullscreen-hotkey"

I'm not particularly attached to the names/style, just the functionality.


larsiusprime:
requires:
https://github.com/openfl/lime/pull/648

larsiusprime:
This should now be closed I assume.

PR #959, opened 2015-12-11
Allow unresizable HTML5 apps

larsiusprime:
If the user sets the window config "resizable" to false in an html5 app, then it forces the window to always be a certain fixed size.

I just add some macro logic to the template. If resizable is true, width/height are "100%" (the current behavior). If resizable is false, width/height are the exact values specified by the user in project.xml


larsiusprime:
Significance: this will be extremely helpful in maturing HaxeFlixel for proper HTML5 support.

Issue #961, opened 2015-12-12
Flash Target: Context3DTriangleFace


peteshand:
[Fault] exception, information=ReferenceError: Error #1065: Variable openfl.display3D::Context3DTriangleFace is not defined.

peteshand:
Looking at other classes in the same package the following also have the same issue:
Context3DTriangleFace
Context3DTextureFilter
Context3DProfile
Context3DMipFilter
Context3DBufferUsage
Context3DWrapMode

deleting the externs at "3,5,2\extern\openfl\display3D\..." resolves the issue, however I'm sure they are there for a reason, any suggestions on how we can resolve this without deleting the externs?

peteshand:
Anyone?

Steps to reproduce.
Create an OpenFL 3.5.2 project in flashdevelop, target flash and place:
Context3DTriangleFace.BACK;
within the constructor.

package;

import openfl.display.Sprite;
import openfl.display3D.Context3DTriangleFace;

class Main extends Sprite 
{
	public function new() 
	{
		super();	
		Context3DTriangleFace.BACK;
	}
}

Publish and you'll get the following error:
[Fault] exception, information=ReferenceError: Error #1065: Variable openfl.display3D::Context3DTriangleFace is not defined.

Obviously this is just one of many classes that are effected. It seems it's a result of changes to abstracts, enums and externs. 

Also why are OpenFL externs included when you already include the OpenFL library?

Finally is there any continuous integration being done on OpenFL?

PR #963, opened 2015-12-13
fix abstract use for flash target

peteshand:
This pull request relates to issue: https://github.com/openfl/openfl/issues/961


jgranick:
Thanks!

Issue #973, opened 2015-12-17
IE/Edge isPointInStroke is unsupported


er453r:
As the title says, but here is a fix I'm using and i I think it is good enough for openfl:

    #if js
    untyped __js__('// openfl IE/Edge isPointInStroke shiv - not super exact for wide strokes
    if(typeof CanvasRenderingContext2D.prototype.isPointInStroke == "undefined"){
        console.log("Pathed CanvasRenderingContext2D with isPointInStroke approximation");

	    CanvasRenderingContext2D.prototype.isPointInStroke = function(path, x, y) {
	        return this.isPointInPath(x, y);
	    }
     }');
    #end

CharlesAnifowose:
Where are you inserting this?

CharlesAnifowose:
fix is working for me until Lime goes live :)

PR #977, opened 2015-12-23
CanvasGraphics null check

KnightMearh:
CanvasGraphics.hx
Hoping this commit can spawn a discussion around this area of code. But it seems there may be some platforms (CocoonJS being one) that do not support or intend to support isPointInPath and isPointInStroke. So for this change, I just added a safety check to ensure the functions exist before trying to use them. If they don't exist, we assume the hit was good (something better to do here?). This change will prevent some uncommon edge cases from crashing regarding platform specifics.

The other thing to consider is optimizing this hitTest function. Seems there may be some early checks we can do to return early, and I believe the code redraws the entire canvas each time that might be able to be optimized.

Sound.hx
Regarding the second file (accidentally lumped into this pull, sorry!), I wanted to add m4a fallback support. If this warrants a discussion we can have that on a separate thread.

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasGraphics.hx
+++ b/openfl/_internal/renderer/canvas/CanvasGraphics.hx
@@ -261,6 +261,12 @@ class CanvasGraphics {
 			
 			var data = new DrawCommandReader (graphics.__commands);
 			
+			// CocoonJS (and maybe some other platforms) does not support isPointInPath or isPointInStroke.
+			// So if we don't have isPointInPath/Stroke, we just assume this was a hit if we made it through Graphics.hitTest()
+			// TODO: safe to call this once and cache result?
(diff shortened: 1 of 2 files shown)


er453r:
I proposed a fix here: https://github.com/openfl/openfl/issues/973

james4k:
@KnightMearh Can this be closed?

KnightMearh:
Thanks for asking. I actually ran into the same issue just today (as I described above) in the latest OpenFL, so it is still crashing.

james4k:
Might try adding a similar polyfill in for isPointInPath: https://github.com/openfl/lime/blob/develop/lime/_backend/html5/HTML5Application.hx#L139

If it works, send a new PR. Also, if you want the 'm4a' change to go in, I would suggest a separate PR for it.

ibilon:

PR #989, opened 2015-12-30
console: various renderer updates

james4k:


james4k:
Rebased and added some fixes wrt. latest internal graphics API changes.

PR #992, opened 2015-12-30
Drawtiles color offsets

larsiusprime:
Adds the ability to specify color offsets (not just color tint) in drawtiles API. I added the extra parameters to the end so that it's just purely extra optional information to add, otherwise the new code should have no effect on existing drawtiles commands.


larsiusprime:
Quick overview of how this works:

Instead of passing in a colorTransform to state by that is either null or not, each State now always has its own colorTransform that is only ever modified to match what's passed in. It is also never null, the skipColorTransform property now covers that functionality.

SpriteBatch also has its own colorTransform instance that is only ever modified, not replaced. This way it's easier to keep track of where colorTransform objects are winding up. They are never passed around, only used to copy information from one another.

The actual rendering looks at the state and sees if skipColorTransform is true, if so, it passes a null to the currentState pointer, otherwise it passes the actual colorTransform.

larsiusprime:
Rebased in a new PR

PR #999, opened 2016-01-06
Drawtiles color offsets 

larsiusprime:
Rebased version of this: https://github.com/openfl/openfl/pull/992

--------------

Adds the ability to specify color offsets (not just color tint) in drawtiles API. I added the extra parameters to the end so that it's just purely extra optional information to add, otherwise the new code should have no effect on existing drawtiles commands.

--------------

Quick overview of how this works:

    Add 4 color offset floats to the end of the drawtiles data.
    Add the TileSheet.TILE_TRANS_COLOR flag.

Each State now always has its own colorTransform that is only ever modified to match what's passed in, rather than passing either null or a pointer to the world transform. This colorTransform is never null, and a new skipColorTransform property covers that "null" case.

This is important because the State object needs to retain the color offset information until rendering, and passing in a global colorTransform object or setting it to null destroys that connection.

SpriteBatch also has its own colorTransform instance that is only ever modified, not replaced. This way it's easier to keep track of where colorTransform objects are winding up. They are never passed around, only used to copy information from one another.

The actual rendering looks at the state and sees if skipColorTransform is true, if so, it passes a null to the currentState pointer, otherwise it passes the actual colorTransform.

### DIFF
--- a/openfl/_internal/renderer/opengl/utils/SpriteBatch.hx
+++ b/openfl/_internal/renderer/opengl/utils/SpriteBatch.hx
@@ -541,7 +541,8 @@ class SpriteBatch {
 				currentState.texture = nextState.texture;
 				currentState.textureSmooth = nextState.textureSmooth;
 				currentState.blendMode = nextState.blendMode;
-				currentState.colorTransform = nextState.colorTransform;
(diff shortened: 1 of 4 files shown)


jgranick:
merged :smile:

PR #1029, opened 2016-01-31
contentType bug fix

hpgood:
request.contentType bug fix

### DIFF
--- a/openfl/display/Loader.hx
+++ b/openfl/display/Loader.hx
@@ -62,7 +62,7 @@ class Loader extends DisplayObjectContainer {
 		
 		contentLoaderInfo.url = request.url;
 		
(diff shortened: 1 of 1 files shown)


ibilon:
The second comparison should probably be changed to `== ""` instead of `!=`.

hpgood:
you are right. 

https://github.com/openfl/openfl/pull/1030

MSGhero:
You can make changes, and the pull request will update with those changes automatically. You don't have to make a new pull :)

hpgood:
sorry, i change code in the fellow url now.it is right ?

https://github.com/hpgood/openfl/blob/patch-3/openfl/display/Loader.hx

PR #1043, opened 2016-02-17
StageVideo classes added (currently only support Flash/AIR)

peteshand:
While Flash/AIR are currently the only supported platforms it's still handy to have these classes you the user can easily us StageVideo on supported targets and fall back on unsupported targets. Additionally it will probably be quite easy to add fallback support internally


peteshand:
Fix Clipboard typedef recursive loop

peteshand:
Hmm looks like the travis cpp build fail, however looking at the details it says:
Tests: 765  Passed: 765  Failed: 0 Errors: 0 Ignored: 2 Time: 0.44898
I am reading this wrong or did it pass everything?

Gama11:
The unit tests were succesful, but another command wasn't: https://travis-ci.org/openfl/openfl/jobs/116254840#L1514

```
Aborted (core dumped)
The command "if [[ $TARGET == 'cpp' ]]; then haxelib run lime test linux --window-hardware=false ; fi" exited with 1.
```

peteshand:
Thanks for the pointer

PR #1044, opened 2016-02-17
autosize bug fix

hpgood:
when __textEngine.height is change,the __textEngine.bottomScrollV need to update  the right value.
otherwise ,the text will display wrong.

### DIFF
--- a/openfl/text/TextField.hx
+++ b/openfl/text/TextField.hx
@@ -869,7 +869,7 @@ class TextField extends InteractiveObject {
 						}
 						
 						__textEngine.height = __textEngine.textHeight + 4;
-					
+						__textEngine.update ();
 					default:
 						
 					


jgranick:
I am concerned that running `textEngine.update` twice like this might be expensive, what value are you getting and what should it be?

Thanks!

hpgood:
example:
 `		var mytextfield:TextField = new TextField();
		var tf:TextFormat = new TextFormat();
		tf.font = "/system/fonts/DroidSansFallback.ttf";
		var tf:TextFormat = new TextFormat();
			tf.font = XUtils.FONT;
			tf.size = 30;
			tf.color = 0x0000ff;
			mytextfield.defaultTextFormat = tf;
			mytextfield.embedFonts = false;
			Lib.current.stage.addChild(mytextfield);
			
			mytextfield.width = 100;
			mytextfield.height = 300;
			mytextfield.x = 10;
			mytextfield.y = 10;
			mytextfield.multiline = true;
			mytextfield.autoSize = TextFieldAutoSize.LEFT;
			
			mytextfield.text = "";
			mytextfield.height;// it will call  __updateLayout ()
			mytextfield.text = "111\n222\n333\n000";

`
openfl test android -x -v -Dnext 

 it can not display the text.
bottomScrollV is wrong

crazyjul:
Just running getLineMeasurements is enough in this case. I fixed it like this : 
https://github.com/FishingCactus/openfl/commit/050877f81501ae276d2a93a1ee189fb76d049ada

jgranick:
I believe this is fixed in current releases, but if it is not, please let me know!

jgranick:
I just tested your code in the latest release, it appears to be working :smile:

PR #1045, opened 2016-02-18
readFunction bug,read max bytes every time.

hpgood:
when the data is big than max bytes,the position do not move in the old code,it will make mistake.


jgranick:
I think this is resolved by moving to `haxe.io.Bytes` in the new implementation for LIme, but if there is still issue, could you share some test code?

Here's the relevant method in Lime:

https://github.com/openfl/lime/blob/develop/lime/_backend/native/NativeHTTPRequest.hx#L334-L338

Thank you :smile:

PR #1052, opened 2016-02-26
Feature fix depth

crazyjul:


ibilon:
What issue does this fix?

crazyjul:
Ok, it seems I pushed on the branch. So there's several fixes at the same time.

I'll push the first one alone

PR #1062, opened 2016-03-14
TexField htmlText improvements

kasparsj:
Adds support for `<p>, <b>, <u>, <i>, <em>` tags.
Adds support for nested tags, e.g. `<font>` inside `<p>`, etc.
Fixes multiline text.


loudoweb:
Hi,

great work! But 2 problems : 

1. When adding `<b>` to a text, it seems that a space is added before
2. When using `<font>` to a text that is close to the end of the textfield, it does NOT return to the next line but it appears cut.

1st problem is not very important, but second one make <font> difficult to use without bug.

loudoweb:
Maybe problem **2** is related to https://github.com/openfl/openfl/issues/1093 ?

kasparsj:
@loudoweb can you post your test texts?

loudoweb:
I've just mailed you with a test project.

Here are my code : 

```
var bugWithRN:String = 'Lorem it blablablabla amet, it amet o dolor soe <font face="Helvetica Neue LT Std 75 Bold">blablablabla blablabl blablabl</font> (the following text appears quite badly as you can see).\r\nIn the previous text in parenthesis you should read: "the following text appears quite badly as you can see"';
		
		
		
var format:TextFormat = new TextFormat(Assets.getFont("fonts/helveticaneueltstd_cn_webfont.ttf").fontName, 17, 0x000000);
format.align = TextFormatAlign.CENTER;

var text:TextField = new TextField();
text.defaultTextFormat = format;
text.width = 205;
text.height = 295;
text.multiline = true;
text.wordWrap = true;
text.htmlText = bugWithRN;
text.x = 0;
addChild(text);
```

It seems that removing or adding just a character in some places can make the textfield works again.
 
It also seems that the bug is occuring because of the linebreak ( \r\n ) after the font tag.

You can see my result in this [screenshot](http://icecream.me/231e17f91941e7a3999f4042bedbd97a).

loudoweb:
@kasparsj false alarm. It seems that it's not related to your pull request but to https://github.com/openfl/openfl/issues/1123

loudoweb:
So there is still :

>1. When adding <b> to a text, it seems that a space is added before

loudoweb:
So there are some additionnal details on issue 1, here are screenshots with related text.

Looks like there are 2 spaces or one big space before the bolded word.
`'amet <b>lorem</b> ipsum'`
screenshot

Looks like the first letter of the bolded word is deleted. It happens if it is the last word.
`'amet lorem <b>ipsum</b>'`
screenshot

Note that if I set bold to the first word, it looks ok.

jgranick:
Merged! :sunglasses:

loudoweb:
I just wanna let you know that the bug I described here (the 17th May) has been fixed in this PR https://github.com/openfl/openfl/pull/1282

kasparsj:
@loudoweb thanks!

PR #1066, opened 2016-03-18
[Complete] Implement setTextFormat in TextField.

tienery:


tienery:
I just want to leave a comment NOT to merge yet, because there is currently an issue when using the `setTextFormat` function on the same line multiple times, where the format ranges are not being correctly rendered in the text engine.

I'm currently looking into it and hope to patch soon.

tienery:
Nevermind, spoke too soon. The Regex I was using to match patterns for Markdown was not perfect, it was matching asterisks as well within the group and needed to exclude the character from the captured group.

That being said, there is an unusual bug with word-wrapping, which I believe should be part of a different pull request (I'm not sure how to do this if I were to personally investigate) and/or separate issue.

MSGhero:
Can you test something in this branch? The current implementation fails, but I don't want to raise an issue if this fixes it...

Calling `setTextFormat()` _before_ setting `.text` doesn't do anything on Flash. `sTF` after setting `.text` is how it's supposed to go.

However on next, you can call `sTF` before or after, and it'll apply the format in either case.

tienery:
`Can you test something in this branch? The current implementation fails, but I don't want to raise an issue if this fixes it...`

That's because the current implementation (in `next`) doesn't really do anything. In Flash, `TextField` is a typedef to `flash.text.TextField`.

My test uses the following code, where `setTextFormat` applies after adding `text` not before:

    package;

    import openfl.display.Sprite;
    import openfl.Lib;
    import openfl.text.TextField;

    class Main extends Sprite 
    {

        public function new() 
        {
            super();
            
            var txt = new TextField();
            txt.defaultTextFormat = new TextFormat("Times New Roman", 12, 0x0000FF);
            txt.width = 500;
            txt.selectable = false;
            txt.text = "CAPITAL LETTERS are blue!";
            txt.setTextFormat(new TextFormat("Times New Roman", 12, 0xFF0000), 16);
            
            addChild(txt);
        }

    }

"However on next, you can call sTF before or after, and it'll apply the format in either case."

I'm assuming that's using my implementation, not the current? As far as I am concerned, I don't think it should matter in what order functions are called. In fact, I think it's better practice that functions don't rely on each other in a specific order, as otherwise it would cause unnecessary complications.

MSGhero:
In Flash I mean. In Flash, the order matters.

If you set the format before setting the text, it'll set the format of the existing text (but there is none). Then when you set the new text, that text uses `defaultTextFormat`, which is the original font and not yours.

If you set the format after setting the text, it'll set the entirety of the existing text (index 0 to -1). If you then set the text to something else, that new text uses `defaultTextFormat` again.

If you `appendText`, that new text uses `defaultTextFormat`.

Flash is entirely unnecessary complications 😛

tienery:
Sounds mostly how I've done it ;)

Let me break it down for you:

Firstly, I check if `beginIndex` is lower than 0, or `endIndex` is greater than the length of the current text. Set it appropriately to prevent errors.

Then I do some calculations to determine what format ranges already exist.

Let's take this scenario, where you have two formats that already exist in the text field:

The text is "Hello, how are you?"

The "Hello" is formatted blue, and the "how are you?" is formatted red.

Let's say you want to format the "how" word with green, 0x00FF00. So use the following code: `setTextFormat(new TextFormat(myFontName, 12, 0x00FF00), 8, 10);`

I create a new copy of the `TextFormatRanges`. I loop through them, checking if the format we want to apply is going to overlap any of the existing ones. The word "Hello" is not overlapped, so we add that to our new copy. The "how are you?" is being overlapped with the one that we want to add, so I shift the existing format's `beginIndex` up to the 11th character, 1 up from the character we set as the `endIndex` for our new format.

I then add the new format first, then the existing one.

If the word were "are" in "how are you?" the method to change this would be that "how" and "you?" are split up into two other formats using the existing format that uses red as the colour, and the word "are" is highlighted green.

The `defaultTextFormat` in my implementation is untouched, and is never changed, so adding new text **should** use this format, regardless of this implementation.

I have done my best to make the `setTextFormat` implementation true to the description found on the Flash API: http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/text/TextField.html#setTextFormat()

MSGhero:
👍

jgranick:
I have been terrible about keeping up with pull requests. We have a new implementation of `setTextFormat` now, but I'm not sure if yours is actually better?

https://github.com/openfl/openfl/blob/develop/openfl/text/TextField.hx#L555-L744

I'm willing to swap things out if your implementation is better, but I don't have enough test code to be sure either way. Would you still be interested in helping with this? 🏆

tienery:
I'll check the two as soon as I can, see which one works better :)

Most of my focus is on using Kha, since it's a lot simpler to use, in my opinion, but perhaps I can give this a go once again.

tienery:
Just a small update.

I have not specifically done any tests and compared results, but it might help to have a reference to someone who has used my implementation: http://community.openfl.org/t/implementing-settextformat/7513/5

tienery:
Might as well close this since this has been implemented now. I have no intention to make any changes to this.

PR #1078, opened 2016-03-31
Enable m4a support for HTML5

Nallebeorn:
Adds m4a to the SoundJS alternateExtensions array, enabling AAC audio in HTML5. Gives us another option than mp3 for Internet Explorer and Safari, which don't support ogg.


ibilon:
Thanks!

PR #1091, opened 2016-04-05
fixes #1090: Tilesheet.drawTiles __bound position when setting tile center and using no flag (on HTML5)

### LINKED ISSUE #1090: Tilesheet.drawTiles __bound position is incorrect when setting tile center and using no flag (on HTML5)
Before fix:
bugged

After fix:
fixed

The issue can be fixed by removing multiplication by tile size on lines https://github.com/openfl/openfl/blob/develop/openfl/display/Graphics.hx#L662 & https://github.com/openfl/openfl/blob/develop/openfl/display/Graphics.hx#L663.

damrem:


ibilon:
`drawTiles` was removed, thanks anyway.

Issue #1074, opened 2016-03-29
Webgl target : black screen since last release (openfl 3.6.1 lime 2.9.1)


TBaudon:
Hello,

I use ``<set name="webgl" if="html5" />`` in my project to export to webgl, but since last update , this does not work anymore, the game is black and no error show anywhere. No loading as well, it's like nothing is running at all.

Maybe it comes from the shortening of js output ? The problem happens when I use openfl 3.6.1, if I use openfl 3.6.0 ans lime 2.9.1 the game works)

I use haxe 3.2.1 btw.

Thanks.

TBaudon:
I think it comes from this commit, skipped it and it works now : 

https://github.com/openfl/openfl/commit/351fddb754a9b214b1785a6ccb491ebde9271a50

TBaudon:
The problem seems to come from openfl/display/Stage.hx line 510.

using <set name="webgl"> produce a webgl output but doesn't seem to define the "webgl" compiler variable. 

Added <haxedef name="webgl"> as well and now the game renders again.

raserg:
I looked on line 510, add "<haxedef name="webgl" if="html5"/>" and it worked.

larsiusprime:
Pull request?

TBaudon:
Sorry, the initial solution I found was to add the haxedef in your project.xml so I didn't thought of a pr ...

PR #1092, opened 2016-04-12
define "webgl" compilation variable to fix  webgl issue #1074

TBaudon:
issue #1074 

Here, I think this fix the problem.
Thanks


ibilon:
Thanks.

PR #1094, opened 2016-04-13
[Complete] Correct word wrap, closes #1093, #1123

### LINKED ISSUE #1093: Word wrap correction
I can't make a pull request from another branch because I've messed up my checkouts on my local repo, but the fix is very simple.

On line 869 of `TextEngine.hx`, the code:

```
offsetX = widthValue + spaceWidth;
```

should be:

```
offsetX += widthValue + spaceWidth;
```

Results before change:

img

Results after change:

img

MSGhero:
See #1093

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -866,7 +866,7 @@ class TextEngine {
 						layoutGroup.height = heightValue;
 						layoutGroups.push (layoutGroup);
 						
-						offsetX = widthValue + spaceWidth;
+						offsetX += widthValue + spaceWidth;
 						marginRight = spaceWidth;
 						
 						wrap = false;


loudoweb:
That's strange because it was like that before this commit : https://github.com/MSGhero/openfl/commit/b6b5d8c562331c0832d8153f77b42ce40f2c9f3a

tienery:
I think it depends on how many formats are being used, so perhaps there should be a if block to detect the number of formats in what range, and on what lines, to determine whether or not `offsetX` should equal or be incremented.

In the instance of using `setTextFormat` (which has not yet been implemented) and setting `multiline` to true with the current state of OpenFL (without the proposed correction I made), you will achieve the results as found in issue #1093 

Perhaps if only one format is used, you shouldn't increment the offset, and now that I think about it, that makes a lot more sense.

MSGhero:
Yeah that wasn't me 👍

tienery:
Oh, sorry, it was linked to your Github fork originally (before the link was corrected) but didn't notice it was a change by Josh.

james4k:
If you can, it would be very helpful to gather test cases for the problems these two conflicting changes are intended to fix.

tienery:
Test Case 1 (my fork, using unchanged word wrap commit, using multiple formats): #1093 
Test Case 2 (my fork, using unchanged word wrap commit, using one format):

Tested code:

	public function new() 
	{
            super();
		
	    var tf = new TextField();
            tf.defaultTextFormat = new TextFormat(Assets.getFont("font/OpenSans-Regular.ttf").fontName, 10);
            tf.wordWrap = true;
            tf.multiline = true;
            tf.embedFonts = true;
            tf.selectable = false;
            tf.width = 100;
            tf.text = "This is some text. This is some more text. Even more text.";
        
            addChild(tf);
	}

Result:

img

Test Case 3 (my fork, using changed word wrap commit, using multiple formats):

By applying the change to: `offsetX += widthValue + spaceWidth;` with the following code:

    var tf = new TextField();
    tf.defaultTextFormat = new TextFormat(Assets.getFont("font/OpenSans-Regular.ttf").fontName, 10);
    tf.wordWrap = true;
    tf.multiline = true;
    tf.embedFonts = true;
    tf.selectable = false;
    tf.width = 100;
    tf.text = "This is some text. This is some more text. Even more text.";
    tf.setTextFormat(new TextFormat(Assets.getFont("font/OpenSans-Bold.ttf").fontName, 10, 0), 26);
        
    addChild(tf);

Result:

 img

Test Case 4 (my fork, using changed work wrap commit, single format):

Same as above, but with a small change to code:

    var tf = new TextField();
    tf.defaultTextFormat = new TextFormat(Assets.getFont("font/OpenSans-Regular.ttf").fontName, 10);
    tf.wordWrap = true;
    tf.multiline = true;
    tf.embedFonts = true;
    tf.selectable = false;
    tf.width = 100;
    tf.text = "This is some text. This is some more text. Even more text.";
    tf.setTextFormat(new TextFormat(Assets.getFont("font/OpenSans-Bold.ttf").fontName, 10, 0), 26);
        
    addChild(tf);

Result:

img

[My Fork](https://github.com/tienery/openfl)
The only change in my fork is the implementation of `setTextFormat`, no other changes have been made.

Gathering from these test cases, with each possible scenario including both the implementation of `setTextFormat` and without its use, when `wordWrap` is set to `true` in either case, the proposed change is favourable over the current commit.

Of course, it's up to you to decide if you wish to go ahead and make the change.

tienery:
With regards to your comment @james4k on the recently referenced pull request, I believe if you set `multiline` to `false` in the TextField, when you move the caret beyond the width of the layout, the `offsetX` would be set to match the relative location of the caret, effectively shifting the text itself left and right.

If my presumption is correct, the final code should be:

    if (!multiline)
        offsetX = widthValue + spaceWidth;
    else
        offsetX += widthValue + spaceWidth;

With `wordWrap` on and `multiline` false, you create the illusion that you can navigate a text box left and right with the caret, which is what you want. I think the proposed change previous to this comment would cause the text to expand instead of remaining with the bounds of the `TextField` itself.

MSGhero:
I can update the pull once someone confirms the `multiline` thing.

MSGhero:
@tienery Your final code is incorrect. `multiline` only affects input textfields, with `true` meaning you can hit enter and it'll add a `\n`.

We might be looking at 2 cases though. I'm looking at regular textfields, where `multiline` doesn't do anything, I don't think. The code should be
```haxe
if (!wordWrap)
    offsetX = widthValue + spaceWidth;
else
    offsetX += widthValue + spaceWidth;
```

But for input textfields, where multiline matters, the fix might be different. Would need to test with those and see.

MSGhero:
`multiline` is not functional in next (master), so I'm gonna say what I have is complete since it fixes `wordWrap`.

Once `multiline` gets implemented in master, there can be a new issue if `offsetX` needs to be changed to account for that.

tienery:
"Your final code is incorrect. multiline only affects input textfields, with true meaning you can hit enter and it'll add a \n"

How does it only affect input text? As far as I am concerned it works on both input and dynamic text:

img

That's with wordwrap off. With it on:

img

As far as I can tell, no problems there.

MSGhero:
Multiline doesn't mean your textfield is allowed to have multiple lines. It's always allowed to have multiple lines.

Multiline means hitting enter adds a newline. The only way you can hit enter is if it's input text.

tienery:
Well, yeah... Just the way you originally worded it didn't sound right.

tienery:
Just so you know, the default behaviour of `multiline` true and `wordWrap` false is supposed to allow for moving the caret left and right and create scrolling text:

https://twitter.com/tienery/status/732481922912247808

I can't upload video demonstrations here, so my demo is on Twitter.

That being said, this is a completely different issue considering nothing in the `getLayoutGroups` function relates to this behaviour. It's all in TextField from line 1825.

MSGhero:
Looking back at this, I kinda want to tackle input text and `multiline`... Maybe this weekend if I can get cpp compiles working

PR #1095, opened 2016-04-14
Partially fix setRenderToTexture()

player-03:
The image is flipped vertically and the depth+stencil buffers don't work, but at least you can work around the former. The important thing is that rendering to texture works at all.

Another note: `setRenderToBackBuffer()` will leave the depth and stencil buffers disabled (suggesting that the answer to line 183 is "no"). You'll have to call `GL.enable(GL.DEPTH_TEST)` and/or `GL.enable(GL.STENCIL_TEST)` to re-enable them.


vroad:
I removed those ```GL.disable``` calls in my fork since I didn't think they were needed. I encountered issues with scissor when I used render target texture, because scissor had been disabled by ```setRenderToBackBuffer()```.

vroad:
This PR is outdated, and my new PR will fixes issues with render target textures.

https://github.com/openfl/openfl/pull/1237

`GL.BGRA_EXT` is not a valid format in OpenGL ES 2.0 either, but almost all devices support that format via extension.

https://www.khronos.org/registry/gles/extensions/EXT/EXT_texture_format_BGRA8888.txt

vroad:
I meant BGRA is not part of OpenGL ES "standard". It's support is just not mandatory.

Issue #1100, opened 2016-04-18
EventDispatcher ignores nested events of same type


TomByrne:
There is a problem with how EventDispatcher dispatches events of the same type.

If an event is fired with the same type as an event that is currently being dispatched, it gets ignored.
Unfortunately this breaks Event.ADDED in many circumstances, and as a result breaks several MVC libraries which are watching for views being added to the stage.

This is the problematic process flow:
- DisplayObject gets added to stage
- Event.ADDED is dispatched and gets caught by some system
- Somewhere within the handler, a child DisplayObject is created and added to the first view
- An ADDED event gets dispatched on this child, which then bubbles to the parent DisplayObject
- The parent DisplayObject goes to dispatch the event, but as it is already dispatching a Event.ADDED the new event it is ignored.

The line which does the ignoring is 199 (in EventDispatcher.hx @ commit f68e4da).

Fixing this is not so trivial, as the __newEventMap/__dispatching collections are only designed to track a single dispatching event.

ibilon:
I intend to test these pull requests this week,
could you post a code sample showing this issue?

Leander1P:
I probably cannot until Tuesday, apologies, but will try to then.  Would very much like to back it up with asserts (any suggestions about how to do asserts?) and tests.  Note #1105 is a different approach than #1101, I saw some other possible bugs and went with a deeper refactor. If you'd like I can base 1105 on 1101.

We had at least one instance where an addEventListener in a callback was enough to get a different unrelated type listener to fire when it wouldn't with the old code.  Will try to reduce that to a testcase.

ibilon:
No worries, there's a lot of bugs for me to see until then ;)

Leander1P:
Okay, I'm having a bit of a time coming up with reliable testcases here -- hopefully @TomByrne may be able to help with a formal version of what he mentioned in the original post?

In the interim, I'm going to comment below with the points that I think might cause problems and attempt to derive some testcases from those, and revisit our most recent game codebase to see if I can extract a reliable exact pattern for a testcase from the issue we saw.

Roughly, in the original code, things that should be able to convert to tests:

1. **Dispatching goes false too soon**: The reset of dispatching at the tail of `__dispatchEvent`, namely the `__dispatching.set (event.type, false);` line, is unconditional.  Clearly we want to keep dispatching true if we're nested, so we should only unset that if we're the "outermost" dispatcher.
[...]

PR #1101, opened 2016-04-18
Fix for EventDispatcher nested event issue

TomByrne:
https://github.com/openfl/openfl/issues/1100

EventDispatcher now tracks how many of each event type it is dispatching.
When the last event (of a type) finishes execution, the __newEventMap info gets committed back into the __eventMap.

### DIFF
--- a/openfl/events/EventDispatcher.hx
+++ b/openfl/events/EventDispatcher.hx
@@ -10,7 +10,7 @@ import openfl.events.IEventDispatcher;
 class EventDispatcher implements IEventDispatcher {
 	
 	
-	private var __dispatching:Map<String, Bool>;
+	private var __dispatchingCount:Map<String, Int>;
 	private var __targetDispatcher:IEventDispatcher;
 	private var __eventMap:Map<String, Array<Listener>>;
 	private var __newEventMap:Map<String, Array<Listener>>;
@@ -31,7 +31,7 @@ class EventDispatcher implements IEventDispatcher {
 		
 		if (__eventMap == null) {
 			
-			__dispatching = new Map ();
+			__dispatchingCount = new Map ();
 			__eventMap = new Map ();
 			__newEventMap = new Map ();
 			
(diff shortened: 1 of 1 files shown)


Leander1P:
We saw this locally as well, just found this patch.

As a note, I don't think (if I'm reading things correctly here) you actually need to track counts.

A few changes to make this possible:

1.) Revert all changes above `__dispatchEvent`.

2.) Change: 

    var wasDispatchingCount:Int = __dispatchingCount.exists (type) ? __dispatchingCount.get(type) : 0;

to:

    var wasDispatching:Bool = ( __dispatching.get (type) == true ); // should default to false for null?

3.) Your new `if (wasDispatchingCount > 0)` block is the same but becomes `if (wasDispatching)`.

4.) `__dispatchingCount.set (type, wasDispatchingCount + 1);` becomes a straight unconditional `__dispatching.set (type, true);` -- this line can also stay in that `else` block immediately above if desired to avoid redundant writes of "true" when it's already true.

5.) The `if(wasDispatchingCount == 0)` block becomes `if (!wasDispatching)` instead.  The `__dispatchingCount.remove` in here becomes a `__dispatching.set (type, false);` -- this is the key piece that allows us not to remember count, we have a local bool there. The `else` for this block can disappear.

Will test and then see about pushing up to the issue.

Leander1P:
If I'm reading things right, the two `__newEventMap = null;` lines (both before and after this change) are possibly unsafe?  Seems silly to assume `__newEventMap` is empty if `__eventMap` is.  I'm having trouble coming up with an example of how this breaks, but it strikes me as very fragile.

Similarly, pretty sure the `__dispatching = new Map ();` (or similar `__dispatchingCount = new Map ();`) lines could stomp an existing dispatching map based on the `removeEventListener` and `__dispatchEvent` methods only `null`ing the other two maps (this clause only checks `__eventMap == null` before newing).  This also seems unsafe.

Trying to reason through this a bit more, and I have a refactor I'm working on to bundle the lifetime of the three together (e.g. Map<String,ArraysAndDispatchingBool>).  I'd rather not lose the opportunistic freeing of the allocs if there are truly no more listeners around, but don't want to do that unless it's provably safe.

jgranick:
Re-entrant event dispatching is fixed 🎉

PR #1102, opened 2016-04-21
fix cast error on html5 target

crazybaz:
cast crashed every time on writing data

### DIFF
--- a/openfl/net/Socket.hx
+++ b/openfl/net/Socket.hx
@@ -182,7 +182,7 @@ class Socket extends EventDispatcher implements IDataInput implements IDataOutpu
 			try {
 				
 				#if (js && html5)
-				__socket.send (cast (__output, ArrayBuffer));
+				__socket.send (__output);
 				#else
 				__socket.output.write (__output);
 				#end


troyc:
I tried this, but it always tries to send the data as text, and if I do something like this:
```
s.writeInt(3);
s.flush();
```
it sends no data at all.

jgranick:
I believe this is meant to trigger a `ByteArray` abstract method for getting an `ArrayBuffer` from the object. Would `var buffer:ArrayBuffer = __output; __socket.send (buffer);` work better?

troyc:
That works better, but it also sends extra null bytes. For instance, this:
```
s.writeByte(1);
s.flush();
```
sends three bytes, 01 00 00, instead of just 01.
This:
```
s.writeByte(0x41);
s.writeByte(0x42);
s.flush();
```
sends 41 42 00.
This:
```
s.writeInt(3);
s.flush();
```
sends 00 00 00 03 00 00 00 instead of 00 00 00 03.

jgranick:
Hmm, perhaps we need to trim the array buffer if the length does not match. I think we have some optimizations internally that ByteArray sometimes allocates extra so every write does not cause a new allocation...

something like?

```haxe
var buffer:ArrayBuffer = __output;
if (buffer.length > __output.length) buffer = buffer.slice (0, __output.length);
__socket.send (buffer);
```

troyc:
Thanks, that works great. It would be ideal if there were a solution that didn't involve copying the buffer (perhaps flushing should just take the opportunity to trim the buffer instead?).

jgranick:
Thank you for your help! I think the change I just made should help -- I agree that avoiding the copy would be nice, although we avoid doing an allocation over and over if you `write` a lot, so my guess is that we might balance it out in the end

PR #1105, opened 2016-04-22
EventDispatcher: refactor against nested event bugs.

Leander1P:
Also add comments, dedupe some logic, avoid unnecessary copies, avoid
redundant lookups, and name and grouping changes to be slightly easier to
reason about.

Stepped through all the branches in getList, addEventListener,
removeEventListener, and __dispatchEvent in the Chrome Canary debugger.

This is intended to be posted in #1100 ; see comments in #1101 .

Specifically:

* Merge all the __eventMap, __newEventMap, and __dispatching maps together
  into one map of a new support struct-style class "EventTypeInfo" --
  saves lookups by key and makes the lifetime of the three members
  relative to one another more obvious.

* There should be no risk of losing listeners by discarding __newEventMap
  (one key/value pair or entirely) too early in removeEventListener,
  before merging it back into __eventMap.

* Toss out listener lists and the entire map when appropriate.

* Versus pull #1101 above, don't *always* copy the list from eventMap to
  newEventMap during __dispatchEvent; if we haven't seen any
  modifications, there's no reason to do this.  eventMap will never change
  out from under us while we're dispatching -- if something comes along
  and asks to mutate the list (Add or Remove mode), *then* we'll copy and
  return the newEventMap (now newListeners) list.

Other notes:

* We now return false from __dispatchEvent on all types of empty list, not
  just null __eventMap.  Before this would have returned true on some and
  false on others, depending on whether the list was nulled after it was
  emptied or not.  Seems like the increased consistency here would be a
  good thing.

* We are taking an extra "new" / indirection in lookups -- we now have
  three levels of null where we'd have two potentially before: __typeInfos
  itself, __typeInfo.get(type), and __typeInfo.get(type).listeners or
  .newListeners.  This only adds a smallish block to getList versus the
  old two-level way, and because we're saving so much on lookups the extra
  indirection should not be costly.  This makes lifetime easier to
  reason about.

Summarizing the getList modes:

* Add will never return null.  Others may (for any of the three levels).
  It may return an empty list (new or existing).

* ReadCopy will never directly return a list that may be mutated; that is,
  it may return "listeners" by ref, which is never mutated while
  dispatching, or it may return a copy of newListeners.

I'd like to add a number of asserts (per the invariants mentioned above
and in GetListMode enum docs), but not sure the best way to do this.

Also, I'm not sure the explicit sets in EventTypeInfo's new method are necessary.

Reviewed locally (first pass) by Danielle Cerniglia and Josh Elliott.
Posting before extensive review for comment.  Any errors are of course my
own.

Please feel free to point out any style or submission problems; I'm still new to contributing to openfl!

### DIFF
--- a/openfl/display/DisplayObject.hx
+++ b/openfl/display/DisplayObject.hx
@@ -225,7 +225,7 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if !disa
 	
 	private function __broadcast (event:Event, notifyChilden:Bool):Bool {
 		
-		if (__eventMap != null && hasEventListener (event.type)) {
+		if (hasEventListener (event.type)) {
 			
 			var result = super.__dispatchEvent (event);
 			
--- a/openfl/events/EventDispatcher.hx
+++ b/openfl/events/EventDispatcher.hx
@@ -1,6 +1,7 @@
 package openfl.events;
 
 
+//import haxe.PosInfos; // for assert
 import openfl.events.EventPhase;
 import openfl.events.IEventDispatcher;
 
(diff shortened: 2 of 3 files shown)


jgranick:
I wanted to try and design `EventDispatcher` to be as low overhead as possible -- to avoid performing a `clone` of the event listener array on each dispatch, but it is important to properly handle the behavior of adding a new listener while we are still dispatching.

I'm open to a different implementation, but I'm having trouble seeing if these two points are covered. Do you know what would happen if you tried adding listeners while you're in the dispatch loop?

Thank you!

Leander1P:
(Rebased onto develop HEAD and added some tests from #1100 here.  Holler if these tests should be broken off into a different PR.  Also considering changing this to a patch/PR series, since it might be possible to satisfy both of these tests with far less churn, although I'd still love to see the refactor go up, I think it's far less prone to lifetime and other issues -- items 2 and 3 in https://github.com/openfl/openfl/issues/1100#issuecomment-214902204 for example, which I haven't yet figured out how to break things with, but seem risky.)

jgranick:
Could you review the behavior after this?

https://github.com/openfl/openfl/commit/3836c81e68225ef9d6228398466c37ba920af346

I merged your unit tests, and everything is passing. If you are able to find any more discrepancies (such as the `return false` behavior) let me know

jgranick:
We are passing all tests, currently. The capture/no capture callback is resolved, as well as re-entrant issues and a behavior I found in Flash Player where it allows removal of listeners in the current dispatch loop if it has not been visited yet.

I'm going to close this issue, but let me know if we find any other discrepancies, thanks 😁

PR #1108, opened 2016-04-25
[wip] Conformant child management

crazyjul:
This PR allows for non-continuous depth in child management.

This is required to play swf timelime correctly. According to the flash documentation, a child could be add at any depth. The timeline use this to let space for child that are going to be added in future frame.

This PR will come with code change in swf as well, but I wait for the dependency to be merged

### DIFF
--- a/openfl/display/DisplayObjectContainer.hx
+++ b/openfl/display/DisplayObjectContainer.hx
@@ -76,17 +76,16 @@ class DisplayObjectContainer extends InteractiveObject {
 	
 	
 	public function addChildAt (child:DisplayObject, index:Int):DisplayObject {
-		
-		if (index > __children.length || index < 0) {
-			
+
+		if (index < 0) {
 			throw "Invalid index position " + index;
 			
 		}
 		
 		if (child.parent == this) {
-			
-			__children.remove (child);
-			
+
+			var index: Int = __children.indexOf (child);
+			__children[index] = null;
 		} else {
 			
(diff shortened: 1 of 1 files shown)


ibilon:
Flash seems to disallow it too, `RangeError: Error #2006: The supplied index is out of bounds.`,
although the documentation doesn't mention such limitation.

Could you provide a sample that is working in flash?

crazyjul:
I didn't try that one in flash directly, but from the a MovieClip timeline. From the documentation, the exception is explained, But from test code, some MovieClip ends up with empty depth slot.

Maybe timelime player has specific right, I'll rework to re-add the exception.

Yet I need to add child at any depth. Does a private API fit?

ibilon:
If that's a special access used by the timeline, here used by the swf library, I guess a private API would be ok.
Thoughts @jgranick?

jgranick:
Would you mind trying `getChildAt` with SWF content? Internally, I assume there are distinct layers, but frame-by-frame, I would guess that `getChildAt` has no gaps. Thanks!

crazyjul:
Sorry, I forgot to update. In flash, the behavior is as you describe. So there's no gap. I'm currently looking for a better solution. I'll update this PR with it asap.

jgranick:
I think that I got another approach implemented to fix this issue in SWF MovieClips, but let me know if you have issues with the current releases :smile:

PR #1116, opened 2016-04-30
Fix setGLSLVertexBufferAt for Context3DVertexBufferFormat.BYTES_4

vroad:
I don't think numBytes is needed here (it's always 4 in all cases), I removed that as well.


ibilon:
Thanks!

PR #1127, opened 2016-05-12
Added line feeds to TextFields, closes #1120

### LINKED ISSUE #1120: TextField does not "\r" as it handles "\n" for line breaks
We need to improve the code to handle both "\r" and "\n" as line breaks

MSGhero:
#1120 

Needs review, but the string `"test\ntest\rtest\n\rtest\r\ntest"` comes out correctly now.

HTML5:
image

I'm having trouble exporting to cpp targets if someone can double check that.


jgranick:
Merged! :smile:

PR #1129, opened 2016-05-13
Added controllable caching - needs lime update.

ShaheedLegion:
Created a mechanism whereby apps which load http resources can control the cache string - by adding a parameter to their project meta tag.


ibilon:
You need to create the pull request against the develop branch.

PR #1130, opened 2016-05-13
Added controllable caching

ShaheedLegion:
Added controllable caching - depends on an update in lime (pull request waiting).

Fixes issue mentioned here: http://community.openfl.org/t/how-to-disable-cache-breaking-string/7432/4


jgranick:
Thanks again for your help :smile:

PR #1164, opened 2016-06-20
Fixed issue for native where TextFields have a thin border on right edge

xavierstampslafont:


openfl/_internal/renderer/cairo/CairoTextField.hx
```diff
@@ -33,6 +33,11 @@ class CairoTextField {
 		
 		var textEngine = textField.__textEngine;
 		var bounds = textEngine.bounds;
```
cactusanctuary: I was a little unsure if I should actually `clone()` the `textEngine.bounds`, but it renders fine. So I think it's OK.

jgranick:
Do you have any test code I can use to see the before and after?

Thank you :grin:

xavierstampslafont:
Hi @jgranick sorry for the long, long silence, my computer broke down a few months back and I've been utterly unable to do any work in my spare time. I don't have anything at the moment to show the issue and how it looks after this fix, unfortunately :(

jgranick:
I think this should be resolved, but please let me know again if there still seems to be an issue. Thank you so much!

PR #1165, opened 2016-06-20
Allow encode(rect, "png")-style calls

IBwWG:
The FlxScreenGrab of HaxeFlixel depends on this.  It doesn't let you pass in a quality, so quality 1 is assumed, but at least it works instead of just returning null, on HaxeFlixel's current default of using OpenFL legacy.  (That's why I left the compatibility comment in there--not sure how it was planned to add that in, without adding an optional parameter, which seems to work well in general except on flash, where it would be a breaking change.  So: partial compatibility without the breaking-change cost.)

### DIFF
--- a/openfl/_legacy/display/BitmapData.hx
+++ b/openfl/_legacy/display/BitmapData.hx
@@ -194,6 +194,10 @@ class BitmapData implements IBitmapDrawable {
 				
 				return byteArray = lime_bitmap_data_encode (__handle, "jpg", cast (compressorOrQuality, JPEGEncoderOptions).quality / 100);
 				
+			} else if (Std.is (compressorOrQuality, String)) {
+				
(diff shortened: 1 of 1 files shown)


ibilon:
Wrong branch, pull request should be done on `develop`,
also legacy was dropped, no new release will have it,
so we can't accept this pull request.
Thanks anyway for contributing.

IBwWG:
No new release will have it?  So HaxeFlixel will have to either use Next or support only OpenFL up to 3.6.1?  Anyway thanks for explaining.  At least this patch is here for anyone who needs it.

ibilon:
> No new release will have it? So HaxeFlixel will have to either use Next or support only OpenFL up to 3.6.1?

Yes, but from what I heard their next support is pretty good.

PR #1168, opened 2016-06-23
html5 colorTransform application

jerobarraco:
rationale:
i want to clone and tint a display object with nested sprites.
the sprites have a negative x because i want to rotate against one specific point.
the copy.draw( ) ends up cropping the image because it doesn't care about the matrix 
the bitmap.transform.colorTransform  doesnt work on html5

Original code can get into perf issues because :
1) one more function call
2) one more object (plus GC)
3) one more render (copy.draw it's the same code as bitmap.draw)
4) copy.draw doesn't care about matrx, that means it could be rendering the whole image, instead of a part (maybe?). 

change brought to you by killabunnies

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -379,16 +379,19 @@ class BitmapData implements IBitmapDrawable {
 		
 		if (!__isValid) return;
 		
-		#if (js && html5)
+		#if (js && html5) 
+		//ill apply the change on this #if, ill leave it up to you to implement on the other #else
 		
-		if (colorTransform != null) {
+		/*if (colorTransform != null) {
 			
 			var copy = new BitmapData (Reflect.getProperty (source, "width"), Reflect.getProperty (source, "height"), true, 0);
-			copy.draw (source);
+			copy.draw (source); //this calls this function, 1) it creates a new function frame, 2) it calls the same code as here 3) it will render the image 1 more time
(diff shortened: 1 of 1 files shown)


jgranick:
But what if you `bitmapData.draw` an object with transparency? What if you draw only to a portion of the target `bitmapData.rect` size?

I wish I knew of a faster way to perform colorTransform on HTML5 than doing pixel-by-pixel manipulation

jerobarraco:
Well, i'm not a openfl core developer so i think that is something they are more entitled to answer you.
The code does almost **exactly** the same thing as the original, with one less function call and respecting the source rect and behaves in a coherent way in all the use cases. 
If you inspect the changes you see that i just removed a call to the exact same function, so if the change doesn't support something, the original doesn't either.
If the code has to clone the source, it has to clone it completely, it shouldn't ignore part of the content of the display object (which is the part that lies with negative coords).
I think the code respects the general design of the library (by using rects to define the portion to be used).
We've been using this exactly to implement a color transform and works perfectly with nested display objects and bitmaps with a negative top-left coordinate and transparent ones.

As for "what if .. portion of" well, the proposed change would acknowledge the proper matrix that is being passed as argument, unlike the original code, so i think that my proposal actually improves on this

jgranick:
This is blocked by https://github.com/openfl/openfl/issues/1403, with `cacheAsBitmap` implemented, we can use it to better implemented color transform for this method :smile:

jerobarraco:
Well, i've already tried the solution i posted, and works pretty fine, i hope you could use that in the time being.

PR #1172, opened 2016-06-29
Added ProgressEvent support for HTML5 target

TomByrne:
This change adds ProgressEvent functionality to the Loader/LoaderInfo classes in HTML5.
The underlying functionality is slightly different, so to opt-out of the new functionality (if for example, it's not behaving correctly in some browser), remove the ProgressEvent listener, which doesn't work in older versions of openfl anyway.

It depends on this Lime commit:
https://github.com/ImaginationSydney/lime/commit/a6bf5edf959d14672782aede9dbc0629030fac66

Which has a pull request to Lime:
https://github.com/openfl/lime/pull/769

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -580,10 +580,10 @@ class BitmapData implements IBitmapDrawable {
 	#end
 	
 	
-	public static function fromFile (path:String, onload:BitmapData -> Void = null, onerror:Void -> Void = null):BitmapData {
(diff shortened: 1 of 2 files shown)


jgranick:
We added asynchronous load calls under the hood, and need to wire them up to `openfl.display.Loader`

I'm going to close this PR, but we'll track the enhancement here:

https://github.com/openfl/openfl/issues/1435

Thank you :smile:

PR #1174, opened 2016-07-02
Fix for segmentation error in textfield on linux

alexeipopov:
This can fix segmentation error on Linux.


ibilon:
Thanks, but it'd be better to fix `System.fontsDirectory` instead of hacking it here.

Could you open an issue with a sample code we could use to reproduce the segfault?

jgranick:
@alexeipopov I think this was added to Lime, let us know if using the dev version of OpenFL and Lime fixes it for you, thanks! :smile:

alexeipopov:
@jgranick yep, seems it fixed. Thanks

PR #1183, opened 2016-07-07
Improved preloader creation/extension

NemoStein:
Pushing PR #704 again, against develop branch this time...

---

The current preloader requires the user to extend NMEPreloader and override its methods, but the NMEPreloader have it's own, hard-coded, behavior.

This pull request changes NMEPreloader to extend OpenFLPreloader, which is a empty class (almost an interface that extends Sprite).
The only requirement to create a preloader now is 1) extend OpenFLPreloader and dispatch an Event.COMPLETE when you are done with the preloader.

### DIFF
--- a/openfl/display/Preloader.hx
+++ b/openfl/display/Preloader.hx
@@ -168,15 +168,10 @@ class Preloader extends LimePreloader {
 }
 
 
-#if tools
-typedef OpenFLPreloader = NMEPreloader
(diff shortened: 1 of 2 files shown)


ibilon:
I wonder if we still need the NMEPreloader template actually,
background color/height/width should be available through lime.app.Application.config

Issue #1185, opened 2016-07-08
Japanese doesn't word wrap correctly in TextFields


larsiusprime:
Repro case:
https://gist.github.com/larsiusprime/5df42543a4b09e07274cce2ad9943b83

What it should look like (flash isn't loading the font but the wrapping is correct):
![2016-07-08 13_24_40-adobe flash player 17](https://cloud.githubusercontent.com/assets/739064/16697479/686df3f4-450f-11e6-801a-8755637736bf.png)

What it does look like (neko/html5 have same results):
![2016-07-08 13_25_06-badjptext](https://cloud.githubusercontent.com/assets/739064/16697481/6b86881c-450f-11e6-843c-3c4598bc8ebd.png)

As you can see, it seems to be an issue with Japanese specifically, as cyrillic (Russian in this case) text wraps correctly. 

I tried with Korean text just now:
![2016-07-08 13_27_48-badjptext](https://cloud.githubusercontent.com/assets/739064/16697562/ce801d7a-450f-11e6-8260-43a0747a95ad.png)

And that seems to be working correctly. I don't have any Chinese strings to test with.

The Korean test string is:
`'시체\n시체들은 종종 죽은 지 몇 일 후에 구덩이 속에서 다시 일어납니다. 이런 괴물들은 ""다시 돌아온 자""라는 의미의 ""망령""이라 알려져 있습니다.'`

And the font is "nanum gothic", bold, available here:
http://www.ffonts.net/NanumGothicBold.font

Everything else necessary to repro with for Japanese is included in the gist.

larsiusprime:
So from some conversations deduced it's because Japanese doesn't use spaces.

Update, just did a test to see what happens if I remove all the spaces from english text:

Flash:
![2016-07-08 13_49_41-adobe flash player 17](https://cloud.githubusercontent.com/assets/739064/16698211/d6156eac-4512-11e6-9751-22912adc6e33.png)

OpenFL (neko/html5):
![2016-07-08 13_48_06-badjptext](https://cloud.githubusercontent.com/assets/739064/16698198/c7275400-4512-11e6-8737-951dbf85ef37.png)

If you remove the english spaces, flash's behavior (but not openfl's) is to break on whatever character, even in the middle of a line.

So it seems like this approach would be "no worse" than flash, and would be appropriate regardless of locale.

larsiusprime:
Presumably the issue is somewhere in this giant function:
[...]

PR #1186, opened 2016-07-09
[Incomplete] Break long words

MSGhero:
Reference #1185 

@larsiusprime, if you don't need multiple formats, I think it's good/only has minor issues

HTML5/Flash results so far:
image

Test string 1-4: `これがうまくいけば、日本語を使用することができることを強調するためにいくつかのテストテキストです。` left, center, right, justify
Test string 5: `これがうまくいけば、日\n本語を使用するこ\nとができ\nることを強調するためにいくつ\nかのテストテキストです。` left

They are different, but HTML5's text is ever so slightly thinner than Flash's for whatever reason, which is causing a few characters to stay on the line when in Flash they would be bumped down. I tried a bunch of English test cases in every text align, but I wouldn't call it thorough.

**Issues:**
1. `RIGHT` align does not give the text a 2px buffer. Hard to see with Japanese, but it was obvious in English (existing issue)
2. I haven't tried multiple formats
3. Selecting text barely works (existing issue)
4. I can't run cpp (still), so I couldn't test there
5. I haven't tested this fully when `wordWrap` is false

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -777,6 +777,8 @@ class TextEngine {
 					
 				}
 				
+				if (i == 0) i = 1;
+				
 				layoutGroup = new TextLayoutGroup (formatRange.format, textIndex, textIndex + i);
 				layoutGroup.advances = getAdvances(text, textIndex, textIndex + i);
 				layoutGroup.offsetX = offsetX;
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -754,6 +754,55 @@ class TextEngine {
 			
 		}
 		
+		inline function breakLongWords (endIndex:Int):Void {
+			
+			var tempWidth = getTextWidth(text.substring(textIndex, endIndex));
+			
+			while (tempWidth > width) {
+				
+				var i = 1;
+				
+				while (textIndex + i < endIndex + 1) {
+					
+					tempWidth = getTextWidth(text.substr(textIndex, i));
+					
+					if (offsetX + tempWidth > width - 2) {
+						
+						i--;
+						break;
+						
+					}
+					
+					i++;
+					
+				}
+				
+				layoutGroup = new TextLayoutGroup (formatRange.format, textIndex, textIndex + i);
+				layoutGroup.advances = getAdvances(text, textIndex, textIndex + i);
+				layoutGroup.offsetX = offsetX;
+				layoutGroup.ascent = ascent;
+				layoutGroup.descent = descent;
+				layoutGroup.leading = leading;
+				layoutGroup.lineIndex = lineIndex;
+				layoutGroup.offsetY = offsetY;
+				layoutGroup.width = getAdvancesWidth(layoutGroup.advances);
+				layoutGroup.height = heightValue;
+				layoutGroups.push (layoutGroup);
+				
+				lineIndex++;
+				textIndex += i;
+				
+				offsetX = 2;
+				offsetY += heightValue;
+				
+				advances = getAdvances(text, textIndex, endIndex);
+				widthValue = getAdvancesWidth(advances);
+				
+				tempWidth = widthValue;
+			}
+			
+		}
+		
 		nextFormatRange ();
 		
 		lineFormat = formatRange.format;
@@ -762,6 +811,10 @@ class TextEngine {
 		while (textIndex < text.length) {
 			
 			if ((breakIndex > -1) && (spaceIndex == -1 || breakIndex < spaceIndex) && (formatRange.end >= breakIndex)) {
+				
+				if (wordWrap && previousSpaceIndex <= textIndex) {
+					breakLongWords(breakIndex);
+				}
 				
 				layoutGroup = new TextLayoutGroup (formatRange.format, textIndex, breakIndex);
 				layoutGroup.advances = getAdvances (text, textIndex, breakIndex);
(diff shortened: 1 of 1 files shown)


larsiusprime:
So those issues seem to be of general concern before merging, but are probably not an issue for me. To my knowledge multiple formats was already broken (I use a hack to enable it in DQ, and it only supports multiple colors, not multiple true formats). I don't care about pixel-perfect fidelity to flash, either, and I don't think I use right-aligning too extensively, and I have no selectable text.

So this is almost certainly enough to monkey patch straight into DQ, and an excellent start for a general solution for OpenFL in general.

Well done!

larsiusprime:
Hmmm, I'm getting what feels like infinite loops on certain screens of DQ, I wonder if we need to look more closely at some of those while() functions.

larsiusprime:
Wonder if this has something to do with it -- Here's HTML5, looks right:

2016-07-11 09_37_20-badjptext

And here's neko:

2016-07-11 09_38_12-badjptext

So it looks like it's working correctly on html5, but not yet on native, and this missing text might cause some infinite loops on complex edge cases?

larsiusprime:
Ya know, I bet it's the fact that we're using a naive char.indexOf() test for finding the line breaks. That'll work when transpiled to Javascript as it's natively UTF-16, but it won't work in C++/Neko as it'll assume the encoding is ASCII or whatever rather than UTF-8.

larsiusprime:
Okay, figured it out, two things are necessary!

First, you have to fix rendering to be UTF-8 aware.

https://github.com/openfl/openfl/blob/develop/openfl/_internal/renderer/cairo/CairoTextField.hx#L321
```haxe
cairo.showText (text.substring (selectionStart, selectionEnd));
```
should be:
```haxe
cairo.showText (Unifill.uSubstring(text, group.startIndex, group.endIndex));			
```
That gets you this: the text renders, but the logic isn't quite right.

2016-07-11 11_23_13-badjptext

@MSGhero's basic logic is right. You just need to make it UTF-8 aware on all platforms. I did this locally with unifill, you just have to replace very `charAt()`, `substr()`, `substring()`, `length`, etc, any function that will slice & dice a string or return internal information about character positioning, with UTF8-aware versions.

Doing that gets you this:

2016-07-11 11_26_42-badjptext

Which appears to be the correct result!

NOTE: for this particular test I had removed the original "\n" line break in the japanese text, it's all just one big string.

larsiusprime:
Final result:

2016-07-11 11_29_35-badjptext

larsiusprime:
Heads up -- I'm getting some infinite loops in battles if I try to use this in production (ie, Defender's Quest). I bet it's from a change to one of the many while loops, let's see if I can pinpoint it.

MSGhero:
I'll do the infinite loop when I get home, but bigger question:

@Jgranick Lars is using unifill library to deal with all of text engine's length, substring, charAt, indexOf, etc calls which mess up with multibyte Japanese text. Haxe 3.4 iirc is the big utf string/bytes redo and is slated for late November. Should we wait till then or add the dependency now?

larsiusprime:
FWIW, my hideous monkey-patch can be found at:

https://github.com/larsiusprime/openfl/tree/allfixes

This is based on a slightly older version of OpenFL, so it will have some noise in the diff that is just because it's a bit behind in various areas.

larsiusprime:
I personally vote for including Unifill as a dependency so long as we can make sure to properly version guard it so that we don't get hit with a surprise regression or whatever if they update their library unexpectedly. I'm very impressed with it and so far, and it has done much better than trying to write my own unicode-aware functions.

For me the question is simply whether to link it as a haxelib or do something else that's less "fragile" or whatever.

tienery:
I have noticed that there is that line that I suggested changing in issue #1093 so multiple Japanese text formats should also work as well with these changes in place.

MSGhero:
@tienery Yeah that still hasn't been merged :) My concern was that I did something incorrectly here (when creating a new layout group) that causes your/my fix to incorrectly handle multiple formats

MSGhero:
Fixed the infinite loop, but not how flash does it. For "aa" and `tf.width = 5`, flash displays nothing while this will show the left bit of the a's on separate lines. But for the Japanese test text, flash was showing what I assume are the left edges of characters. `¯\_(ツ)_/¯ `

tienery:
Wasn't test cases for TextField's supposed to be implemented at some point? I remember one conversation we had (can't find it) related to the [Improvements to TextField](https://github.com/openfl/openfl/milestone/4) milestone where someone said (possibly @ibilon if I remember correctly) that we needed test cases before continuing to merging current issues. And apparently until those test cases have been completed for current implementations will we then move forward onto the milestone.

Has that been done yet? Because it's been several months now.

MSGhero:
It hasn't, not sure if it's in development. It's just really hard because there are so many weird little things that have gone wrong in textfields, lots of issues we haven't discovered yet, plus the output doesn't perfectly match flash's.

I don't think it's possible to get a comprehensive test suite, just a collection of issues that have broken in the past perhaps.

tienery:
What I might do is re-implement the entire TextField and get an idea on which part of the code is doing what, split it up, add comments identifying where things are and what they're doing, and then copy-paste into their appropriate places. At least that way when we return to editing the source, we know what's causing what.

MSGhero:
Found a few more edge cases, will be testing those.

I'm also gonna comment this section for the next poor soul who needs to make changes here (probably will be me). I don't know what advances are, but the rest makes sense.

MSGhero:
Moved to #1661

PR #1201, opened 2016-07-21
ByteArray fixes for HTML5/JS - Fixes issue #5474

dschreck:


jgranick:
Let's keep our eyes on this one, please let me know if you have a test case that fails with the latest version of ByteArray, and we can work together on a solution. Thank you! :smile:

PR #1205, opened 2016-07-24
2 small bug fixes for Sound class in html5

kasparsj:
- remove from __registeredSounds when closing
- dispatch COMPLETE event after a delay, so that first a listener can be set

### DIFF
--- a/openfl/media/Sound.hx
+++ b/openfl/media/Sound.hx
@@ -1,6 +1,7 @@
 package openfl.media;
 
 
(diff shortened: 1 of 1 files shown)


ibilon:
I think the listener should be added before calling a function that can send rather than hacking it with a timer delay.

kasparsj:
The listener cannot be added before "load" is called, since "load" is called from the constructor.

PR #1224, opened 2016-08-01
Contributing Zynga's implementation of Stage3D under MIT.

dschreck:
Didn't get much time this weekend after this contribution was approved so I know it's pretty rough. Feel free to fork from my repo back into your main line if you need to make major changes. 

Feel free to sync with me so I can point out some of the functionality that I have yet to implement in Haxe/OpenFL.

### DIFF
--- a/externs/core/openfl/openfl/display/ShaderJob.hx
+++ b/externs/core/openfl/openfl/display/ShaderJob.hx
@@ -1,0 +1,82 @@
+package openfl.display;#if (display || !flash)
+
+import openfl.events.EventDispatcher;
+import openfl.display.Shader;
+
+@:final extern class ShaderJob extends EventDispatcher {
+    @:isVar public var width(get, set):Int;
+    @:isVar public var height(get, set):Int;
+    @:isVar public var progress(get, never):Float;
+    @:isVar public var shader(get, set):Shader;
+    @:isVar public var target(get, set):Dynamic;
+
+    public function new(shader:Shader = null, target:Dynamic = null, width:Int = 0, height:Int = 0){
+        super();
+    }
+
+    public function cancel():Void
+    {
+        //Stub
+    }
+
+    public function start(waitForCompletion:Bool = false):Void
+    {
+        //Stub
+    }
+
+
+    public function set_target(t:Dynamic):Dynamic
+    {
+        target = t;
+        return target;
+    }
+
+    public function get_target():Dynamic
+    {
+        return target;
+    }
+
+    public function set_shader(s:Shader):Shader
+    {
+        shader = s;
+        return shader;
+    }
+
+
+    public function get_shader():Shader
+    {
+        return shader;
+    }
+
+    public function get_progress():Float
+    {
+        return 0.0;
+    }
+
+    public function get_height():Int
+    {
+        return height;
+    }
+
+    public function set_height(h:Int):Int
+    {
+        height = h;
+        return height;
+    }
+
+    public function set_width(w:Int):Int
+    {
+        width = w;
+        return width;
+    }
+
+    public function get_width():Int
+    {
+        return width;
+    }
+
+
+}
+#else
+typedef ShaderJob = flash.display.ShaderJob;
+#end
--- a/externs/core/openfl/openfl/display3D/textures/RectangleTexture.hx
+++ b/externs/core/openfl/openfl/display3D/textures/RectangleTexture.hx
@@ -1,4 +1,5 @@
-package openfl.display3D.textures; #if (display || !flash)
+package openfl.display3D.textures;
+#if (display || !flash)
 
 
 import openfl.display.BitmapData;
@@ -6,13 +7,28 @@ import openfl.utils.ByteArray;
 
 
 @:final extern class RectangleTexture extends TextureBase {
-	
-	
-	public function new ():Void;
-	public function uploadFromBitmapData (source:BitmapData):Void;
-	public function uploadFromByteArray (data:ByteArray, byteArrayOffset:UInt):Void;
-	
-	
+
+    public var width(get, null):Int;
+    public var height(get, null):Int;
+    private var mWidth(default, null):Int;
+    private var mHeight(default, null):Int;
+    private var mFormat(default, null):String;
+    private var mOptimizeForRenderToTexture(default, null):Bool;
+
+    public function new ():Void;
+    public function uploadFromBitmapData (source:BitmapData):Void;
+    public function uploadFromByteArray (data:ByteArray, byteArrayOffset:UInt):Void;
+
+    public function get_width():Int
+    {
+        return mWidth;
+    }
+
+    public function get_height():Int
+    {
+        return mHeight;
+    }
+
 }
 
 
--- a/externs/core/openfl/openfl/display3D/textures/SamplerState.hx
+++ b/externs/core/openfl/openfl/display3D/textures/SamplerState.hx
@@ -1,0 +1,22 @@
+package openfl.display3D.textures;
+#if (display || !flash)
+@:final extern class SamplerState {
+
+    public var MinFilter(default, null):Int;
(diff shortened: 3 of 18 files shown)


tienery:
What's "Zynga's implementation of Stage3D"? I'd rather just the title be "Implementing Stage3D" at least that way you're not being perceived as directly endorsing a company, and especially Zynga of all of them (no offence).

That being said, I've researched up on this and I see they have their own language and binding for Flash called PlayScript, apparently a so-called successor to ActionScript 3. Lol... Maybe as the superior community, we should market Haxe more and make sure Zynga doesn't get any ground on this? Just saying.

larsiusprime:
@tienery : dschreck works for Zynga

jgranick:
Merged! :shipit:

dschreck:
@tienery - Correct, we released PlayScript under Apache 2.0 back in 2013, but then withdrew it. This is the only version of this code under the MIT license, which we've discussed in detail with @jgranick

vroad:
@jgranick Excuse me, but could you revert these stuff? I can't run Starling programs with this because of many bugs. Why did you push stuff that break many existing functionalities? That is called regression, and you should avoid that when it's possible. Making software software stable is much more important than madding new functions to software. It shouldn't be in develop branch at least.

I think it would be much better to merge these stuffs one by one. That way you can guess which change broke existing functionalities. It's a bit hard to debug this because you included many stuff in a few number of commits...

This PR contains some interesting stuffs that are not currently supported by current implementation, I'd like to take a look at them anyway.

vroad:
It would be much easier to start from partially broken state than from completely broken state...

MSGhero:
Maybe there should be a superdev branch, but you can go back yourself via haxelib iirc.

jgranick:
We are very thankful for the contributions, and because of the origins of our older Stage3D bindings (the only written documentation of wighawag's original code is listed as GPL) something had to be done from a licensing perspective. There are a number of great things about these bindings, I think that we will just need to help integrate it a bit more. I will be testing Starling soon.

We have a "develop" branch to guard in case someone needs a stable master branch to use. Anyway, let's all be positive, and yes, let's make things solid and stable! We just need a free open-source basis to do it upon. Thanks everybody :smile:

vroad:
wighawag said on his github repo it's not GPL.
https://github.com/wighawag/openfl-stage3d/issues/23

vroad:
So there is no reason to erase all previous work done by me and other developers. That's just waste of time. Did you even try to contact him?

I'm threatened by these commits because many users will report bugs with that. Do I need to fix new bugs/reintroduced bugs by myself?
This is the issue you could avoid by reviewing code carefully before merging.

You might be able to guess that "develop" branch has some issues, but this case is kind of extreme. I've been using development version of softwares such as Haxe, and even if developers say it's in development it's fairy stable.

larsiusprime:
Couple a things -- 

I agree with vroad it might make sense to start a temporary feature branch to work out the regressions. "develop" branches do seem like they're fair game to be unstable, but I've been bitten by this myself a few times and I think it's worth it to take time to nail down regressions without slowing anyone else down. The code isn't going anywhere so there's no harm in giving it some time.

As for the license however, even if wighawag said it's not GPL, if it was originally marked as GPL, and then others contributed to it, their contributions are probably legally presumed to be under GPL as well. In order to get the whole thing back under MIT you would need to track down every single contributor since the original license statement and get them all to agree that each individual contribution is fair game for MIT. Even then many companies would be afraid to use it in their work. I'm not saying it can't be overcome. If the original license mis-statement was long ago this can be a big issue, if it was fairly recent it might not be too hard to fix.

vroad:
I guess it's mostly written by wighawag. That project has a few number of contributors, but do we get sued in practice? I guess it's highly unlikely. Nobody will sue us on not releasing a few lines of code with GPL. And they've never complained about it.

It may make sense to get permission from major contributors. It would be better to get permissions from contributors who made small changes, but it's highly unlikely to get sued on that. They will gain almost nothing from sueing us.

I don't think you need to get permissions from anyone contributed to away3d-core-openfl as it says it's MIT.

larsiusprime:
> Nobody will sue us on not releasing a few lines of code with GPL.

Whether or not that's true for you and me, that might not be the opinion of a company seeking to use OpenFL and could serve as a barrier to wider adoption.

I thought this was the repo in question:
https://github.com/wighawag/openfl-stage3d/blob/master/haxelib.json#L4

vroad:
Do other projects simply gives up relicenslng if they cannot contact somebody?
Yes, It's hard to prove something "does not" have copyright issues... devil's proof?

Not sure how can I "rewrite" code when a few lines have some legal issues.
It looks hard to write something very unique for that.
You may end up writing the same kind of code in this case. Do others claim that you stole the code?

larsiusprime:
It's a common problem with relicensing, but it can be worked around from what I've heard. If the total lines in question that are unaccounted for are small, I think you can re-write them so long as you follow a few guidelines. Again, I'm not a lawyer.

To be clear, the main issue here is not "oh no will this contributor sue me" -- though that's a possibility, as you say it's very unlikely. The real issue is that a company's legal department does due dilligence on OpenFL, notices a discrepancy like this, and refuses to touch it because of the technical liability. So it's either deal with it now or deal with it later.

From what I can tell, the Dolphin Emulator team had to re-write a small % of the code from people they could no longer contact, and that was enough to satisfy the legality of their license, for more details I guess it would be best to ask one of them.

vroad:
I don't want OpenFL userbase to shrink, but if we release OpenFL with unreliable code, that will prevent users from using OpenFL... That might be worse than having (unlikely) legal issues. 

As stage3d layer being a thin wrapper around OpenGL ES, it's very hard to write implementation in a very unique way.
Maybe AGALMiniAssembler is not. You are required to do some complicated work to convert AGAL to GLSL.

larsiusprime:
Well it looks like we have two different workable implementations here and any legal issues can likely be surmounted with the proper expertise. As both of those issues seem to be well over my head I trust it to you guys to sort it all out and I'll bow out of here now :D

jgranick:
Don't worry, we're not doing a new release yet. We can discuss more on Slack. Thanks :smile:

tienery:
Here is some information (since I know my copyright laws well, even though I'm not a lawyer) about GPL licensed code:

 * You can sublicense GPL code under a similar license, i.e. You can choose to use version 2 of the GPL instead of version 3, and vice versa.
 * Normally, under the GPL, you need to attribute back to the author regardless of how you use the code.

However, all of that being said you do need to use a similar license to the GPL and MIT is certainly not one of them. There is no harm in asking permission from contributors of the original source code.

Also, you very rarely hear a case about a lawsuit between two open source developers, simply because of the money involved, unless the open source developer on one end is run by a private company who has millions of dollars to expend. Even that being said, most companies are reasonable enough not to go after the little guy, simply because it's not worth it. I wouldn't worry about a lawsuit, personally. I don't think it would ever go that far, even if the original developers had money.

That being said, you have "merged" the commit and if any of that source code is identified to be GPL code, what you have effectively done is sublicensed the code under MIT. But MIT isn't compatible with GPL, so even if it isn't a release, you're still technically violating the GPL. However, if the original developers are reasonable enough, you may get away with it. I can't make any promises on that front.

Sunjammer:
As someone working for a company currently striving very hard to sort out licensing issues so we can go OS, yeah, don't take licensing stipulation lightly. When it becomes business it is the very serious kind.

jgranick:
GPL is especially bad, because of its a "copy left" nature, using GPL source could mean that someone could approach a business and ask for all of their source for any products created using the library.

In C++, you can create a shared library and avoid some of these problems. That isolates it safely from user code, but we cannot do that with Haxe source right now. It's a non-starter.

For Linux system libraries, I understand the purpose of the license, but not for this.

@vroad thanks for your pull request today, let's continue to sync up on Slack on how we can make this the strongest representation of Stage3D in OSS :smile:

PR #1227, opened 2016-08-04
add dumpBits (remove bitmapData from RAM while keeping _texture)

loudoweb:
adding `dumpBits`to openfl next allow same behavior as legacy (where a dumpBits method already exists). Some game engine like HaxeFlixel uses dumpBits when compiling with legacy so it will be great to have it in next too.

I've test the method and it works (sometimes I need to call `System.gc()` right after to remove RAM immediately)

Of course we have to call dumpBits after the texture is generated once so I've added `canBeDump()`method.

### DIFF
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -376,6 +376,18 @@ class BitmapData implements IBitmapDrawable {
 		
 	}
 	
+	public function dumpBits ():Void {
+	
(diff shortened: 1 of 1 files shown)


loudoweb:
I'm closing because it has been implemented recently https://github.com/openfl/openfl/commit/d72f0ce5ef962ede97737890262ffac55e10203b

PR #1231, opened 2016-08-06
fix: agal bits should be in little endian ordering

gNikro:
agal bits should be in little endian because of AGALMiniAssembler sources

`agalcode.endian = Endian.LITTLE_ENDIAN;
agalcode.writeByte( 0xa0 );// tag version
agalcode.writeUnsignedInt( version );// AGAL version, big endian, bit pattern will be 0x01000000`


jgranick:
Thank you :smile:

PR #1241, opened 2016-08-17
Fix for stroking too early when changing lineStyle

vizanto:
Before this patch, the speech bubble in [this test program](https://github.com/vizanto/openfl-linestyle-bug) gets a diagonal line drawn trough it, while the Flash version is obviously correct.

Here's a picture of what's wrong:
HTML5

This is how Flash renders the same code:
Flash


vizanto:
Travis fails on docs :(

ibilon:
> Travis fails on docs :(

most likely because you targeted the master branch instead of develop, changed that for you,
I'll see if I can relaunch the travis test.

jgranick:
Merged, thank you :smile:

PR #1246, opened 2016-08-18
Apply ColorTransform directly when rendering text and shapes

vizanto:
This patch refactors and de-duplicates code in the Cairo and Canvas rendering classes and implements the application of color transforms to TextFields and (Shape) Graphics.

Luckily worldColorTransform was already implemented so it turned out to be fairly straightforward :-)

It introduces 2 new macros to ColorTransform that help calculate the RGBA values without temporary allocations. If and when Haxe supports multiple inline return values, these could be rewritten. Hopefully you'll find these readable enough... :)

Drawing Bitmaps with ColorTransform is untested and not implemented by this patch.

### DIFF
--- a/openfl/_internal/renderer/cairo/CairoGraphics.hx
+++ b/openfl/_internal/renderer/cairo/CairoGraphics.hx
@@ -19,6 +19,7 @@ import openfl.display.GradientType;
 import openfl.display.Graphics;
 import openfl.display.InterpolationMethod;
 import openfl.display.SpreadMethod;
+import openfl.geom.ColorTransform;
 import openfl.geom.Matrix;
 import openfl.geom.Point;
 import openfl.geom.Rectangle;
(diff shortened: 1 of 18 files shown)


vizanto:
Added some ColorTransforms to [my little test program here](https://github.com/vizanto/openfl-linestyle-bug/tree/colortransform)

Before this patch, the ColorTransforms change nothing on targets other than Flash.

vizanto:
I just discovered that changing the `colorTransform` after rendering, doesn't invalidate the Graphics and the display is not updated. Working on that now...

vizanto:
Discovered a copy-pasta error in `CanvasGraphics.render`.  Rebased the proper commit.

mtari:
I have tested your changes and discovered one issue. Drawing a line isn't working, but drawing other shapes seems to work fine.

Example code:
var line = new Sprite ();
line.graphics.lineStyle (10, 0x24AFC4);
line.graphics.moveTo (10, 0);
line.graphics.lineTo (755, 0);
addChild (line);

vizanto:
Reopened as https://github.com/openfl/openfl/pull/1251 because I keep pushing to this branch :/ 
Sorry.

PR #1251, opened 2016-08-24
Apply ColorTransform directly when rendering text and shapes

vizanto:
This patch refactors and de-duplicates code in the Cairo and Canvas rendering classes and implements the application of color transforms to TextFields and (Shape) Graphics.

Luckily worldColorTransform was already implemented so it turned out to be fairly straightforward :-)

It introduces 2 new macros to ColorTransform that help calculate the RGBA values without temporary allocations. If and when Haxe supports multiple inline return values, these could be rewritten. Hopefully you'll find these readable enough... :)

Added some ColorTransforms to [my little test program here](https://github.com/vizanto/openfl-linestyle-bug/tree/colortransform)

Before this patch, the ColorTransforms change nothing on targets other than Flash:

HTML5:
HTML5

Flash:
Flash

Drawing Bitmaps with ColorTransform is untested and not implemented by this patch.

### DIFF
--- a/openfl/_internal/renderer/cairo/CairoGraphics.hx
+++ b/openfl/_internal/renderer/cairo/CairoGraphics.hx
@@ -19,6 +19,7 @@ import openfl.display.GradientType;
 import openfl.display.Graphics;
 import openfl.display.InterpolationMethod;
 import openfl.display.SpreadMethod;
+import openfl.geom.ColorTransform;
 import openfl.geom.Matrix;
 import openfl.geom.Point;
 import openfl.geom.Rectangle;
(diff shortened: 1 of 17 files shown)


ibilon:
I can confirm that this fix the issue,
but no idea about the code itself, cc @jgranick

vizanto:
The patch also works for the "Tint" Color Effect in Flash / SWF files.

vizanto:
@jgranick maybe 4.4.0? 😀

vizanto:
Maybe... 4.8.0? 😃

vizanto:
How about having a threshold in Graphics (count of draw commands?) for this to kick in VS pixel/shader based color transforms?

If you're drawing a square and transforming the color, it's probably faster to do in software?

PR #1252, opened 2016-08-25
Change "get, null" to "get, never."

player-03:
https://haxe.org/manual/class-field-property-rules.html

Any variable marked "`(get, null)`" is what's called a "physical field." This means that Haxe allocates space for it. For example, let's take the `Point` class:

```haxe
class Point {
    public var length (get, null):Float;
    public var x:Float;
    public var y:Float;
    //...
}
```

When you create an instance of `Point`, Haxe allocates space for three floats: `x`, `y`, and `length`. But, when we look at the rest of the class, we find that `length` is never used. Not even in the getter:

```haxe
private function get_length ():Float {	
    return Math.sqrt (x * x + y * y);
}
```

There's no need to allocate that space, and in an application with enough `Point` objects, it could actually make a difference. By changing "`(get, null)`" to "`(get, never)`," you can tell Haxe that you'll never write to this field, allowing Haxe to optimize it out.


ibilon:
Good catchn thanks!

Hasufel:
+1!

MSGhero:
Should setting length scale the x and y though? That's what it's used for in nape, but I'm on mobile and not in a position to check if the same is true in as3.

Issue #284, opened 2014-07-10
MouseEvent.MOUSE_OVER and MouseEvent.MOUSE_OUT dosen't work on html5 target


Kickerua:
Actually MouseEvent.ROLL_OVER and MouseEvent.ROLL_OUT dosen't work too.

Here is modified HandlingMouseEvents example:
https://dl.dropboxusercontent.com/u/17378315/haxe/mouseOverExample.zip

vizanto:
Um this is closed, but the patch doesn't implement `ROLL_OVER` or `ROLL_OUT`...

PR #1259, opened 2016-09-06
Implement ROLL_OVER and ROLL_OUT MouseEvents

vizanto:
According to the ActionScript 3.0 documentation, the MOUSE_OVER and MOUSE_OUT events bubble, while the ROLL_OVER and ROLL_OUT events do not.

Testing with Flash Player shows ROLL_OVER is dispatched before MOUSE_OVER and ROLL_OUT after MOUSE_OUT.

Related to #284


vizanto:
I don't think travis failed because of my patch

jgranick:
Thank you!

Please give the latest a try, I believe it implements the proper bubbling behavior for MOUSE_OVER/MOUSE_OUT, while being more specific about ROLL_OVER/ROLL_OUT with a preserved stack (similar to how we did mouse over/out before)

Issue #1225, opened 2016-08-01
ColorTransform inconsistent behaviour


Misiur:
```
package;

import openfl.display.Sprite;

class Main extends Sprite {
    public function new () {
        super ();

        var a = new Sprite();
        a.graphics.beginFill(0x0000FF);
        a.graphics.drawRect(0, 0, 200, 100);
        a.graphics.endFill();
        a.x = 100;
        a.y = 100;
        a.rotation = 45;

        addChild(a);

        var bitmapData = new openfl.display.BitmapData(500, 500, false, 0x00000000);

        var matrix = a.transform.concatenatedMatrix;
        bitmapData.draw(a, matrix/*, new openfl.geom.ColorTransform(1,0,0,1,255,0,0,50)*/);
        var bitmap = new openfl.display.Bitmap(bitmapData);
        bitmap.x = 500;
        addChild(bitmap);
    }   
}
```

![](https://i.imgur.com/IK3hGr4.png)

`bitmapData.draw(a, matrix, new openfl.geom.ColorTransform(1,0,0,1,255,0,0,50));`

![](https://i.imgur.com/pRhYnRf.png)

https://github.com/openfl/openfl/issues/1214 might affect the transform. Flash target doesn't share those bugs, but both Cairo and Canvas do.

Misiur:
Currently digging through HTML5 implementation code.

https://gist.github.com/Misiur/6194cdc6d0b66c460d3ec954142529c8

> Boot.hx:62 ImageDataUtil.hx:55: 0
> Boot.hx:62 ImageDataUtil.hx:55: 565.685424949238
> Boot.hx:62 ImageDataUtil.hx:55: 1131.370849898476
> Boot.hx:62 ImageDataUtil.hx:55: 1697.0562748477141

Seems `ImageDataView#stride` is not an integer after all! After applying `Math#ceil`, all pixels are colored, that's a good start.

Misiur:
Ok, so the root cause is getting source width and height by reflection - this bypasses type-checks and makes a mess.

```diff
diff --git a/openfl/display/BitmapData.hx b/openfl/display/BitmapData.hx
index 55a2209..b923dc4 100644
--- a/openfl/display/BitmapData.hx
+++ b/openfl/display/BitmapData.hx
@@ -384,8 +384,9 @@ class BitmapData implements IBitmapDrawable {
 		#if (js && html5)
 		
 		if (colorTransform != null) {
-			
[...]

PR #1266, opened 2016-09-13
Make sure reflected properties are integers. Partial fix for #1225

Misiur:
In HTML5 target reflected `source#width` and `source#height` can be floats, which causes type mismatch for width and height temporary BitmapData, causing problems with ColorTransform.

More info in https://github.com/openfl/openfl/issues/1225


ibilon:
Thanks.

PR #1281, opened 2016-09-24
Convolution filter implementation for gl renderer

Beeblerox:
i've tried to make as less changes to other classes as possible, but convolution shader needs to know the size of texture, so i've added `object:DisplayObject` parameter to filter's `__initShader()` method. Not sure if it's correct, but hope to get some feedback about it.


jgranick:
I wonder if we can sample from texture2d without knowing the size (using a `.x - 1`, etc?) but I implemented it with a different method of getting the size. I also included some additional changes to improve the behavior. Thanks for your help :smile:

PR #1282, opened 2016-09-25
[HTM5] IE11 text measuring and formatting fix #1273

### LINKED ISSUE #1273: overlap characters bug on HTML5 IE 11/edge (bounty 150 EUR)
Hello,

it seems that `textfields`on **IE 11/edge** don't work properly. Sometimes words overlap on each other.

Example : 

should be : "**best in his or her work,**"

but I've got : 

bug on IE11

It is not related with https://github.com/openfl/openfl/pull/1186.
Chrome works perfectly, it is only on IE 11/edge.
This bug seems random because adding some characters on a textfield can fix it or make the bug appears.

mayakwd:
The root of the problem is text measuring mechanism that depends on browser.
IE 11 return rounded values for Context.measureText, because of that summ of pixels for every character in the text line and width of text line by itself will be different.

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -643,16 +643,8 @@ class TextEngine {
 			
 			var advances = [];
 			
-			#if (js && html5)
-			
-			for (i in startIndex...endIndex) {
-				
-				advances.push (__context.measureText (text.charAt (i)).width);
-				
-			}
-			
-			#else
-			
+			#if !(js && html5)
+
 			if (__textLayout == null) {
 				
 				__textLayout = new TextLayout ();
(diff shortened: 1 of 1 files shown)


jgranick:
I am not certain what the best solution is here, I am concerned that using a single measurement per text group will result in poor word wrap support. One possible solution (performance willing) is to measure (at least on IE) each increasing letter ("h" then "he" then "hel"...) rather than each letter individually, and use those values to determine the correct advance.

Another solution might be to use a DIV to do something similar, I wonder how performance and accuracy compare between a DOM element and canvas measureText 😕

loudoweb:
@MSGhero What conflicts are you talking about?

MSGhero:
If this gets merged, my own PR would have merge conflicts

loudoweb:
This one ? https://github.com/openfl/openfl/pull/1186 ? Oh I see. It's important that everything merge correctly. I also need your PR :)

loudoweb:
The PR fixes some of my bugs but not each of them. I've still some disappearing space and overlap characters. I think there are some problem with rangeIndex and spaceIndex which generate wrong offsetX for TextLayoutGroup (that's just a guess I don't have any proof).

Also I've found that if I round the values of  `Context.measureText`. These bugs now appear in Chrome, Opera,etc.
I think we need to round all values when using `Context.measureText`. I know that it will currently break some text in html5 project that doesn't target IE11.
But we will be sure to have similar rendering across browser.
Then we can fix the remaining overlap characters bugs.

mayakwd:
Lucas, I'm still working on it. 
We must not round values, because it's producing side effect, sum of widths of each character in word "Hello" will be definitely not the same as width of word by itself. 

Imagine that you have a long sentence where each character has a decimal part in the width. For example: each character has width 6.12px, every 10 characters mistake of measuring will be grow by 1.2 px, and so on.

So final measuring of whole sentence and rounded width of characters will be different for a lot of pixels - that's the reason why in IE 11 measureText gives not the same values for characters and sentences.

mayakwd:
Well, I've made some research and found that my PR broke all text input functionality, and it's not that simple, I have another solution that will not broke any previous functionality, I'll test it today and try to implement in new PR.

MSGhero:
If it makes you feel better, text input is already super broken. Input is such a different beast that I want to tackle that once the current set of textfield PRs get completed and merged.

But they probably will never get merged because they each break input even more 😅

mayakwd:
@loudoweb , I'm totally found the solution and going to change this PR, tomorrow I'll commit it.

loudoweb:
Awesome! Thanks!

mayakwd:
It's better to close this one and check #1308

PR #1288, opened 2016-09-28
Fix TextField auto-sizing remaining too small and cropping text.

vizanto:
Before this patch, sometimes one would need to actually set the width to textWidth to get a proper sized textfield after setting `text`:

```
label.text  = str;
label.width = label.textWidth;
```

By moving width and height calculation responsibility to TextEngine.getLineMeasurements() this workaround is no longer required.

---

However I'm not completely sure this works in all cases as intended :)
Are there any TextField behaviour tests?

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -604,6 +604,27 @@ class TextEngine {
 		}
 		
 		maxScrollV = numLines - bottomScrollV + 1;
+
+
+		if (autoSize != NONE) {
+			
+			switch (autoSize) {
+				
+				case LEFT, RIGHT, CENTER:
+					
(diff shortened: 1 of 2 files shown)


MSGhero:
That is the intended behavior unless you set `tf.autoSize`. Does that property work?

vizanto:
Looks like I wasn't clear in the code sample:  
Before this patch, with `autoSize` set to anything other than `NONE` it won't actually _always_ change size.

jgranick:
Thanks! I made a few minor changes, let me know, guys, if you see differences in how autoSize behaves (for example, when and whether it allows increased size -- I thought it normally only reduced size?)

vizanto:
Looks good, thanks for the improvements!

PR #1293, opened 2016-10-05
addEventListener fix for varying useCapture:

Leander1P:
Per docs from externs/core/openfl/openfl/events/EventDispatcher.hx:

```
 * <p>Keep in mind that after the listener is registered, subsequent calls to
 * <code>addEventListener()</code> with a different <code>type</code> or
 * <code>useCapture</code> value result in the creation of a separate
 * listener registration.
```

Using `list[i].match` here and comparing both `listener` and `useCapture`, rather than just `compareMethods` on the `listener`, will ensure we do add two listener entries if called once with useCapture = true and once with useCapture = false.  This allows you to listen for the event both during capture and target/bubbling phases with the same method.

See also http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/EventDispatcher.html#addEventListener()

Reviewed-by: Chris Howe

### DIFF
--- a/openfl/events/EventDispatcher.hx
+++ b/openfl/events/EventDispatcher.hx
@@ -68,7 +68,7 @@ class EventDispatcher implements IEventDispatcher {
 			
 			for (i in 0...list.length) {
 				
-				if (Reflect.compareMethods (list[i].callback, listener)) return;
+				if (list[i].match (listener, useCapture)) return;
 				
 			}
 			


jgranick:
Have you tested this in Flash Player?

When I was testing using a recent version of Flash Player, it seemed they changed this behavior to allow only one type per callback, so to listen to both phases you would need separate callback functions and cannot reuse the same

Leander1P:
Will try to test shortly.

If that's the case, `removeEventListener` probably needs changes to ignore `useCapture`, since having them behave asymmetrically may not be great?  Something else to test in recent Flash Player, I guess.

Annoying that they're no longer following http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/events/EventDispatcher.html#addEventListener() if so -- am I missing a more recent doc reference?

Leander1P:
Trying to set up a test and noticing in existing EventDispatcherTest.hx that this clause:

```
		// Capture is true
		
		var correctPhase = true;
		var dispatcher = new EventDispatcher ();
		
		var listener = function (event:Event) {
			
			correctPhase = (event.eventPhase == EventPhase.CAPTURING_PHASE);
			
		}
		
		dispatcher.addEventListener ("event", listener, true);
		dispatcher.dispatchEvent (new Event ("event"));
		
		Assert.isTrue (correctPhase);
```

...is never actually reaching `listener`; since `correctPhase = true` to start, this doesn't fail the test.  AFAICT the `CAPTURING_PHASE` is only really triggered by various things in `Stage` and/or `DisplayObject.__fireEvent`.  It makes sense, since there's only really a capture if you're navigating a DO hierarchy, but how is best to test this with a detached hierarchy?  Or does this need to change to add to stage and then somehow go through a `__fireEvent` path?  Should a random detached-from-stage DO tree not have a `CAPTURING_PHASE` if you `dispatchEvent` through it?

(Should I file a separate issue for this?)

Contriving an attached-to-stage case for testing AS3 behavior now, but I'm curious...

Leander1P:
@jgranick: I ran the flash tests.  The behavior seems to be as I assumed and the documentation describes: registering the same event type, same listener, with different `useCapture` will cause both `CAPTURING_PHASE` and non-`CAPTURING_PHASE` callbacks to the same listener.

I changed `FocusEventTest` (the old `(using anonymous functions to prevent repeat listener optimization)` block) to:

```
		// Now register our listeners...
		for (s in [ root, old1, old2, new1, new2 ]) {
			s.addEventListener(FocusEvent.FOCUS_IN, checkEvent );
			s.addEventListener(FocusEvent.FOCUS_IN, checkEvent, true );
			s.addEventListener(FocusEvent.FOCUS_OUT, checkEvent );
			s.addEventListener(FocusEvent.FOCUS_OUT, checkEvent, true );
```

...and that checks out; the test passes in the JS munit pane with my fix and fails without it.

It also passes in the AS3 munit pane in both cases (with my changes and without) with current flash player (NPSWF32_23_0_0_162.dll for Firefox was what I tested with).

(May be worth noting there's a failure in TextFieldTest in AS3, but that seems to be preexisting.)

I'll roll up the test changes as another PR.

jgranick:
I wish I remembered how I tested, but I found on current Flash Player versions (contrary to my expectations) you could register two methods, one for capture, one not, but could not share the same method on both. There is also the chance this was a bug/regression in a Flash Player release that was later fixed.

Even if it's a project I have to run to test the behavior, it would be helpful to know we have this

jgranick:
Thanks, I had missed your other PR that tested this. Merged, thank you :smile:

PR #1295, opened 2016-10-05
EventDispatcher: cleanup after outermost dispatch for a type.

Leander1P:
This avoids resetting dispatching for a particular type to `false` in cases where we nest.

In some cases, such as dispatch `FooEvent` -> dispatch `FooEvent` again from listener -> return from inner dispatch (`__dispatching (type)` was set `false` here) -> remove or add a listener for `FooEvent` -> return from outer dispatch, this could have potentially caused mutation of the list we were iterating over in `__dispatchEvent`.

The `if (listener == list[index])` paranoia can probably be removed now, but I'd like to get some more paranoia and more tests in place first.

-----

The bulk of this change is simply:

1. caching the results of the `( __dispatching.get (type) == true )` test as `wasDispatchingThisType` (so named to avoid confusion with "was dispatching *any* type", which may be different), and
2. only *conditionally* executing the tail of the function if `!wasDispatchingThisType`: cleaning up `__newEventMap` and related, and resetting dispatching to `false` for the type.

This is part of breaking #1105 up into a more manageable patch series, and addresses the most pressing/obvious bug I saw.

See also https://github.com/openfl/openfl/issues/1100#issuecomment-236187526 ; this also fixes the "Dispatching goes false too soon" case in https://github.com/openfl/openfl/issues/1100#issuecomment-214902204 .

### DIFF
--- a/openfl/events/EventDispatcher.hx
+++ b/openfl/events/EventDispatcher.hx
@@ -192,8 +192,9 @@ class EventDispatcher implements IEventDispatcher {
 		
 		var type = event.type;
 		var list;
+		var wasDispatchingThisType = ( __dispatching.get (type) == true );
 		
-		if (__dispatching.get (type) == true) {
+		if (wasDispatchingThisType) {
 			
 			list = __newEventMap.get (type);
 			if (list == null) list = __eventMap.get (type);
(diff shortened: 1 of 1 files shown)


Leander1P:
This was made irrelevant by jgranick's 3836c81e68225ef9d6228398466c37ba920af346 .  The outermost dispatch's activity is now tracked by `reset` and `hasNext` in `private class DispatchIterator`.

jgranick:
After our conversation yesterday, I decided to make an array of iterators, to be sure we remove an event listener if it has not been visited yet, no matter how many levels of recursion there are in the event dispatcher.

The event dispatcher test has only one level of recursion with this behavior, not two, but I believe it should be fixed now

https://github.com/openfl/openfl/commit/125b42d7353145a02f9dab68952b935505c65d31

Instead of nagging me that there's an edge case we "probably" will not hit, but might, I think this fix will make it so that we can handle (almost) infinite levels of recursion and still behave.

Leander1P:
> 125b42d

Thanks, will give this a look.

Since the "deletions effect lists we're iterating over" is not the most intuitive thing and isn't documented in either the Adobe AS3 docs or elsewhere that I can see, is it worth a change to docs somewhere to explicitly mention that removals affect dispatch-in-progress but other things don't?

We're considering keeping a patch locally to have the currently-iterating-over lists *never* change, since it's a simpler (and perhaps more expected) behavior model for our programmers to remember, and we're not as concerned with compatibility with Flash behaviors that seem unintuitive/buggy/asymmetrical.

I assume you have a mandate to follow Flash as closely as possible here, but I'd be curious to hear any thoughts you have on this kind of policy -- what quirky behaviors are worth reproducing (although it may well be that this was completely desired by the AS3/Flash lib authors and not a "bug" per say, for reasons that escape me at the moment -- maybe remove is more "important/immediate" than add conceptually).

This gets back to my long and unorganized rambling about "ideal" behavior at https://github.com/openfl/openfl/pull/1296#issuecomment-251981743 , and concerns about other corner cases in the whole dispatch model.

jgranick:
I try to follow the Flash behaviors in general, but avoid things that seem like implementation quirks, not a feature of the API, per se. There are some caching behaviors in `TextField` for example, where values don't change until you re-assign the `text` property. I'm not sure if that's worth duplicating.

The `event` object in AS3, as far as I recall, is cloned when it is dispatched to each listener. For garbage collection and performance, this is not a behavior I have chosen to replicate (though recently I considered whether we should "reset" the values of the event object before each dispatch, but I'm not sure if we need to worry about that?)

I believe the `removeEventListener` behaving immediately makes sense... you don't want to receive that event, and if you do (because it's currently in the dispatch list) may seem wrong. Adding in another dispatch may well be necessary to prevent an infinite loop. Removing right away may also be necessary for a similar reason.

The behavior of `EventDispatcher` is an example of something that is tightly related to assumptions various projects may make about expected behavior. If we make this different, we may break many projects "for no good reason", which is not the idea :smile:

So we're not an emulator, but we do want to follow what Flash projects expect

PR #1296, opened 2016-10-05
EventDispatcherTest: add tests for nested behaviors.

Leander1P:
The first test succeeds on current `develop` branch; the second fails without #1295 and succeeds when it is present, showing abberant behavior.  Here's the failure:

    FAIL: massive.munit.AssertionException: Value [a(abc)bac] was not equal to expected value [a(abc)bc] at openfl.events.EventDispatcherTest#testDispatch2 (312)

...which indicates we get an additional call to the "a" callback we were not expecting.  (Adds or removes should not affect an in-flight dispatch's list.)

-----

This originally appeared in #1105 as https://github.com/openfl/openfl/pull/1105/commits/636e1d20c91d65a06101e5ccea32f3b4dc2ca893 .

### DIFF
--- a/tests/unit/test/openfl/events/EventDispatcherTest.hx
+++ b/tests/unit/test/openfl/events/EventDispatcherTest.hx
@@ -244,4 +244,73 @@ class EventDispatcherTest {
 	}
 	
 	
+	var test01aCallCount = 0;
+	var o:EventDispatcher = null;
+	var test02aCallCount = 0;
+	var test02Sequence:String = "";
+
+	private function test01a (e:Event):Void {
+		++test01aCallCount;
+		if ( test01aCallCount == 1 ) { // avoid infinite recursion, but we still should get a second call
+			o.dispatchEvent( new Event("Test01Event") );
+		}
+	}
+
+	@Test
+	public function testDispatch1()
+	{
+		test01aCallCount = 0;
+		o = new EventDispatcher();
+
+		o.addEventListener( "Test01Event", test01a );
+		o.dispatchEvent( new Event( "Test01Event" ) );
+
+		Assert.areEqual(2, test01aCallCount);
+	}
+
+	public function test02a (e:Event):Void {
+		test02Sequence += "a";
+		++test02aCallCount;
+		if ( test02aCallCount == 1 ) {
+			test02Sequence += "(";
+			o.dispatchEvent( new Event( "Test02Event" ) );
+			test02Sequence += ")";
+
+			// dispatching should still be true here, so this shouldn't modify the list we're traversing over
+			// ...but it does...
+			o.removeEventListener( "Test02Event", test02a );
+			o.removeEventListener( "Test02Event", test02b );
+			o.addEventListener( "Test02Event", test02a, false, 4 );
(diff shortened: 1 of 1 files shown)


Leander1P:
This should likely not use members, and use local lambdas etc, to avoid unecessary overhead / state between tests.  Will clean up and amend to look more like `EventDispatcherTest.dispatchEvent`

Leander1P:
I've exposed a strange behavior in AS3 here, see the 4a2852f commit:

```
Failure: testDispatchingRemainsTrue (0.001s)
massive.munit.AssertionException: Value [a(abc)c] was not equal to expected value [a(abc)bc] at openfl.events.EventDispatcherTest#testDispatchingRemainsTrue (310)
```

...AS3/Flash is mutating the outermost dispatch list *after* we've started iterating in this case.

I'm not sure what to do about this.   Neither pre-#1295 or post-#1295 matches AS3's behavior with the latest iteration of this test.

Leander1P:
So -- having slept on this a bit, it seems like it would be good if we had a list of AS3 behaviors.

My intuition is that *ideally*, the EventDispatcher ought to behave _as if_ each dispatch got a separate `.copy()` of the listeners array before starting to loop over them and dispatch.  (We don't want it to _actually_ be a `.copy()` for performance reasons.)

To put it another way, again *ideally*, once an event starts to dispatch, the list of listeners that will receive the event is set in stone, and cannot change (excepting a possible handwave for event cancellation, to some degree) -- specifically, `addEventListener` and `removeEventListener` will not alter who receives "already in flight" dispatches.

These ideals strike me as the lowest cognitive load / "least surprise" for users of the API.

4a2852f demonstrates that AS3 is not doing holding to the above ideals -- I've triggered a case where the `b` listener is not triggered despite being in the list at the start of the dispatch.  Presumably something to do with removal and reinsertion at a different priority, but we can probably narrow this down.

We could get a better idea of how Flash Player is implementing this by doing something akin to "fuzzing" the interface, and that might give us a list of its behaviors and some guess at its implementation.

We could then try to mirror the implementation to faithfully reproduce its behavior -- but I think most people would find behaviors like this unintuitive (versus my "ideally" above).  Perhaps if we had such a list, we could figure out what was *worth* reproducing and what was just more likely a "bug" in the Flash Player AS3 impl.

...I'm not sure whether this is worth the effort versus just accumulating bugs over time, but it does strike me as a huge amount of uncertainty to be living with.  From personal experience, we've "backed into" odd corner cases in the dispatch system, and working around them hasn't always been trivial; I'd far rather see an implementation that is easy to reason about, unsurprising, and has easy-to-validate preconditions and postconditions around its API calls.  This was kind the aim of #1105.

It's also unsettling that some of these can cause events that normally would be matched to suddenly not be matched, or some single-fire events to be received repeatedly, etc.

Would appreciate any other thoughts on how to proceed here.  It may simply be that there's not enough time/interest available to scope out the full behavior of AS3 in this regard and come up with a reasonable policy for how OpenFL should behave vis-a-vis "reentrant" behavior -- API use while events are already in flight.  (Could also possibly scope out other things like the question raised in #1299 about detached DisplayObject hierarchy dispatch behavior.)

PR #1301, opened 2016-10-08
Add MIT header to Matrix.hx

justinmclean:
Hi,

The Apache Flex project uses a modified version of Matrix.hx and it would be a help the project if a MIT header was added to the original file. This is not needed by the terms of the MIT license (as you already have a LICENSE file with the MIT text in it) but does help if that file becomes separated from the LICENSE file (as is the case in the Apache Flex project). I can also understand that you may not want to add this header and if that is the case please just reject this PR.

You may also want to consider adding MIT headers to all files but that's totally up to your project.

Thanks,
Justin

### DIFF
--- a/openfl/geom/Matrix.hx
+++ b/openfl/geom/Matrix.hx
@@ -1,3 +1,27 @@
+/* 
+ * MIT License
+ *
+ * Copyright (c) 2013-2016 Joshua Granick and other OpenFL contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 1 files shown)


jgranick:
Hi Justin,

I am sorry that I took so long to respond, we've had a backlog of issues that have needed spring cleaning. Doing much better now :wink:

We choose not to put license information in each file, because it makes it difficult to update the license, makes it a bit harder to jump into a file, and can make file searches fail :smile:

For your copy, it sounds like adding the license on the top would be helpful. Do you have a link to where it's hosted?

Thanks :smile:

justinmclean:
Hi,

Thanks for that. Here's a GitHub mirror of the Apache repo it's in:
https://github.com/apache/flex-asjs/blob/241d20201c6bf298879a2123c40328b5817b6c2b/frameworks/projects/Core/src/main/flex/org/apache/flex/geom/Matrix.as

We'll go ahead and ad date header like you suggested. If possible could you supply the copyright name and year?

Thank,
Justin

justinmclean:
It's "Copyright (c) 2013-2017 Joshua Granick and other OpenFL contributors" right?

PR #1302, opened 2016-10-08
data.image0 null check

Beeblerox:
without this check i get null pointer exception with my custom display object, which using custom shader


ibilon:
Oops saw your message before this pr and made the change manually :)

Thanks anyway.

Beeblerox:
thanks for adding this check :)

PR #1306, opened 2016-10-11
Don’t remove and add when child is already in place

vizanto:
Because.


MSGhero:
Can there be events that get dispatched on removing or adding? If so, you would now be passing over those, which may or may not happen in flash.

jgranick:
@MSGhero AFAIK, no events are dispatched when you changed the display order of a DisplayObject on the same parent :smile:

PR #1308, opened 2016-10-12
[HTML5] TextEngine fixes

mayakwd:
Instead of previous PR (https://github.com/openfl/openfl/pull/1282) that broke a lot of things is better to use this one.

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -925,7 +925,6 @@ class TextEngine {
 				
 				while (true) {
 					
-					if (textIndex == formatRange.end) break;
 					if (spaceIndex == -1) spaceIndex = formatRange.end;
 					
 					advances = getAdvances (text, textIndex, spaceIndex);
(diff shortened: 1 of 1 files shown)


mayakwd:
@loudoweb, try this one please.

loudoweb:
Hello. Thanks for the PR. I'm currently trying but there is something wrong somewhere (maybe this PR, maybe last Lime or Openfl, I don't know) because my project sample crashes.

mayakwd:
I've tested on the latest openfl and lime.

loudoweb:
Does last openfl/lime need last Haxe? I'm on Haxe 3.2.0.

jgranick:
I test OpenFL and Lime against Haxe 3.2.1, personally

loudoweb:
I've made further test and the reason my sample crashes is not because of this PR, it's because : 
`TextextField.htmlText` is buggy in last openfl when using html tags.

There are recent commits on TextEngine and TextField, I'm pretty sure the culprit is one them (be careful, flash works but not other targets).

loudoweb:
If I remove this commit https://github.com/openfl/openfl/commit/d2c98111bd6c400b8757b288834eac869c8144cc everything is fine! So this the culprit.

So I've test your PR and indeed you fixed the issue https://github.com/openfl/openfl/issues/1273 without input caret regression. Good job!

loudoweb:
Thank you!

jgranick:
Thanks for your help, I believe both of these have been resolved :smile:

PR #1315, opened 2016-10-18
Fix Sound.length for HTML5 howler.js based Audio

vizanto:
In HTML5, `__buffer.data` is never actually set and before this patch calling `length` would result in a null-pointer exception.

### DIFF
--- a/openfl/media/Sound.hx
+++ b/openfl/media/Sound.hx
@@ -169,10 +169,17 @@ class Sound extends EventDispatcher {
 	private function get_length ():Int {
 		
 		if (__buffer != null) {
+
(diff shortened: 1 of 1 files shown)


jgranick:
Would it work to use __buffer.length or __buffer.length + __buffer.offset on all targets?

vizanto:
This is all I see in lime AudioBuffer:

```haxe
	public var bitsPerSample:Int;
	public var channels:Int;
	public var data:UInt8Array;
	public var sampleRate:Int;
	public var src (get, set):Dynamic;
```

jgranick:
Oh, right, audios ounce has the length, thanks

PR #1323, opened 2016-10-22
Smoothing fix #1222

### LINKED ISSUE #1222: Nearest-neighbor and graphics.beginBitmapFill()
I am a bit new to OpenFL so it might be that I am doing something wrong. Using the latest version (4.0.3) of OpenFL, I am trying to draw an upscaled pixel-art sprite. The code I am using:
```
graphics.clear();
graphics.beginBitmapFill(bitmapData, null, false);
graphics.drawRect(0, 0, bitmapData.width, bitmapData.height);
graphics.endFill();
```
The third argument of the `beginBitmapFill` method turns smoothing on or off. Yet the sprite still looks blurry and not the way it should look if it was using nearest-neighbor interpolation.

I was not able to reproduce this issue on OpenFL version 3.6.1 and Lime 2.9.1.

mayakwd:


jgranick:
Thanks for the help :smile:

I walked through and may have gotten a complete fix, I'm a little unsure, though, of the smoothing values in canvas. Unlike Cairo (which does smoothing per image pattern), smoothing is global to the context, so we need to change the smoothing value when we are finished with a bitmap fill or stroke.

Let me know if you see anything else

PR #1329, opened 2016-10-29
Replace NMEPreloader template for backwards compatibility.

bendmorris:


ibilon:
@jgranick how does https://github.com/openfl/openfl/commit/c3237c9b5eb69ec9ab3e7d8ac211df75e43f724e mix with this?
This file can't be removed just like this, people are extending it.

jgranick:
I have made a workaround for now

https://github.com/openfl/openfl/commit/ee5f10dbcb8c3d2f59e5051b8089c0f2e16307ff

I am still concerned, though, that we will have the same problem in the future, if we ever remove this code. I'm a little concerned about the "magic" nature of this, but I think it has been long enough that we should not have this file included by default

ibilon:
A typedef with a `@:deprecated` metadata left for a version would have been a lot simpler.

It hasn't been long enough, you just removed its usage and added something else, instead of getting a deprecation message people would get a build error if they use this class, that's not good.

PR #1330, opened 2016-10-30
Reduce memory allocations in TextEngine.

bendmorris:
A few changes to overwrite existing arrays and avoid splicing unless necessary. According to hxScout this significantly reduces memory churn when a fixed number of TextFields have frequently changing contents. Still room for improvement elsewhere, but submitting this part now since it seems pretty uncontroversial.


jgranick:
After you posted this PR, I changed them to use `Vector` instead of `Array`, so we could set the length without removing the object. I believe this solves this problem of churn on these objects, but please let me know if you continue see any issues here.

Thanks a lot :smile:

PR #1333, opened 2016-11-01
Fix: cursor not visible on the last empty line

vizanto:
As discussed on Slack:

Before this patch when typing ENTER in a multiline TextField in HTML5 or Native target, the cursor disappeared. As soon as any non-space character was entered after, the cursor would appear again.

This patch fixes it by ensuring a TextLayoutGroup is created for the last empty line.

`//end of painful debug session`


vizanto:
[Small TextField test project can be found, here.](https://github.com/vizanto/openfl-textfield-test)

PR #1355, opened 2016-11-24
Clarify NMEPreloader deprecation warning

bendmorris:
http://community.openfl.org/t/preloader-now-that-nmepreloader-is-depreciated/8437/2

Right now there's a deprecation warning but it's not clear to users what they should do about it.

### DIFF
--- a/templates/haxe/ApplicationMain.hx
+++ b/templates/haxe/ApplicationMain.hx
@@ -347,7 +347,7 @@ class ApplicationMain {
 							
 						} else if (searchTypes.pack.length == 0 && searchTypes.name == "NMEPreloader") {
 							
-							Sys.println ("Warning: Use of NMEPreloader has been deprecated");
+							Sys.println ("Warning: Use of NMEPreloader has been deprecated; use openfl.display.Preloader.DefaultPreloader for a drop-in replacement");
 							break;
 							
 						}


jgranick:
Thank you for your help

The new preloader system is much simpler, I also removed the deprecation warning. I believe it will throw an error that it cannot find `NMEPreloader` if you still extend it, but it should be simple to include this file in your project.

The new system expects the preloader extend `Sprite`, and will dispatch `Event.COMPLETE` as well as `ProgressEvent.PROGRESS` events on it. Optionally, the preloader cancel `preventDefault()` on `Event.COMPLETE`, and dispatch `Event.UNLOAD` when it is finished

jgranick:
Do you think we should do something to make `NMEPreloader` work without additional changes?

I'm in favor of causing there to be less reasons to break a previous project, however, I'm also concerned that if we have a workaround, "NMEPreloader" will live on forever :smile:

Thanks

bendmorris:
Does this mean you've removed the workaround that kept NMEPreloader working?

I was encouraged by your post on adopting SemVer today. Since this is a long-standing public API that many users have extended, removing it would qualify for a major version bump. So please do keep a workaround in place (with a clear deprecation warning) at least until the next major release.

jgranick:
I agree, since we are not on OpenFL 5, we should keep a workaround in place. I've opened an issue

https://github.com/openfl/openfl/issues/1436

With SemVer, we should be able to use `#if openfl < 5` for the workaround. I'll try and get something working :smile:

PR #1366, opened 2016-12-02
Fixed of `BitmapData#encode`

SlavaRa:
Now method `BitmapData#encode` works according to offical documentation http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/BitmapData.html#encode() properly


jgranick:
Thank you :smile:

PR #1374, opened 2016-12-07
Feature/filters cache as bitmap

Greg209:
This introduces software based filters for Blur, Glow and DropShadow which adds the rendered filter as a texture and rendered as any other texture. This has the advantage of generating the filter output once (on change) reducing need for multipass shaders. This was started prior to the multipass rendering added recently. 

In addtion, an implementation of cacheAsBitmap is included to flatten the display list to a single bitmap for filters to be applied. This cached bitmap also supports masking.


openfl/text/TextField.hx
```diff
@@ -2082,13 +2106,9 @@ class TextField extends InteractiveObject {
 		switch (key) {
 			
 			case RETURN, NUMPAD_ENTER:
-				
-				if (__textEngine.multiline) {
-					
-					replaceSelectedText ("\n");
-					dispatchEvent (new Event (Event.CHANGE, true));
-					
-				}
```
sbimikesmullin: @Greg209: nice PR, but why remove the `if (__textEngine.multiline) {` block here?

Greg209: Looking back at it, I think it was a result of a merge that removed the if condition. It seems that the if condition was added a few days prior to this PR in 2dd1c15c0b8abb98905ff5be2d8ed09f1acda875. I don't see any reason why I would have removed it intentionally.

jgranick:
Should be merged now, thank you so much :grin:

PR #1375, opened 2016-12-08
Use TextField SWF offset when rendering

vizanto:
This fixes my problem mentioned in comments here: 6dc0ae1abb50c177c6ba21899bcabd77d4b15a2a

Is this a reasonable solution?


vizanto:
Soundcard broken in travis

PR #1446, opened 2017-01-24
Much much much crispier 2D Vector Graphics

vizanto:
This PR does a few things:

- gets rid of a lot of "roundPixels" logic...
- AND gets rid of rounding errors altogether, by doing less floating point math.
- properly scales DOM canvas and webgl canvas for high DPI screens
  - (but WebGL only when `<window allow-high-dpi="true">`)

Before, with `allow-high-dpi="false"`: 
before

Before, with `allow-high-dpi="true"` - yes, it's completely broken:
before-wacky-allowhighdpi

After patch:
after

Just look and weep at those pretty rounded edges ;)

### DIFF
--- a/openfl/_internal/renderer/canvas/CanvasGraphics.hx
+++ b/openfl/_internal/renderer/canvas/CanvasGraphics.hx
@@ -239,6 +239,10 @@ class CanvasGraphics {
 		
 	}
 	
+	#if (js && html5)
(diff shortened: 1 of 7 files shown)


vizanto:
See the first commit message for details.

jgranick:
Merged! :grin:

PR #1459, opened 2017-01-27
Add a possibility to loop through dictionary by values

NenadBojkovski:
We are currently porting as3 to Haxe/OpenFL and are missing the "for each" feature to loop by values. Of course we are able to work it through but it would be much elegant if it is supported by the Dictionary API.
In the current openfl.utils.Dictionary implementation there is no such possibility. SInce it is a very simple change and will only make Dictionary API reacher I think it would be cool to merge it.
None of the existing Dictionary API is changed.

### DIFF
--- a/openfl/utils/Dictionary.hx
+++ b/openfl/utils/Dictionary.hx
@@ -53,6 +53,11 @@ abstract Dictionary<K, V> (IMap<K, V>) {
 		
 	}
 	
+	public inline function values ():Iterator<V> {
+		
+		return this.iterator ();
+		
+	}
 	
 	@:to static inline function toStringMap<K:String, V> (t:IMap<K, V>, weakKeys:Bool):StringMap<V> {
 		


jgranick:
I just implemented this for Flash, and renamed it to `each()`, so you can `for (key in dict)` or `for (value in dict.each ())` :smile:

NenadBojkovski:
Awesome! This was very fast :). Thank you so much. When could we expect it? In the next release ?

jgranick:
Yep!

If we wanted, we could make `Dictionary` behave like the `Map` class, where you need `.keys()` to get the keys, however, I wondered when I wrote it if we should have `Dictionary` behave similar to an ordinary `for (key in dict)` for sake of porting code. Interested in feedback :smile:

NenadBojkovski:
I've been also thinking to implement FloatMap in Dictionary in addition to other maps. In as3 we use some dictionaries with Numbers(Floats) as keys. What do you think would it be ok to have such map ?

NenadBojkovski:
I expected it to work as haxe maps and was surprised when I realised it actually is inverted. Namely, we use as3hx to convert as3 code to hx, we've been converting Dictionaries in the same way as Haxe Map would be converted only to realise that there is no keys method exposed and the loop is inverted. Additionally, when "for each"  is converted we needed to inject additional line of code where the dictionary value is being read. 
Now, having "each" method the provides Dictionary values will make converted code much more elegant. However being inverted, the dictionary loop could be contra intuitive for the devs that are use to haxe maps. Might be better to have Dictionary in sync with haxe map, but that change in Dictionary api might break application which are already build on an existing Dictionary api.

jgranick:
That came up recently with another developer, the issue there you might have multiple values that feel like the same value, but don't match. For example, `3.333333` vs `3.33333333`

Perhaps it could be done with a certain level of precision, and rounded to an int, or we could risk having lots and lots of keys :smile:

NenadBojkovski:
Agree! It could be implemented that way. I'll check that and keep you updated.

NenadBojkovski:
Hey @jgranick,  I forgot to ask you how could I join OpenFL slack channel? Do you need to send me an invitation ?

jgranick:
Here :smile:

https://openfl-slack-invite.herokuapp.com

NenadBojkovski:
Thanks a lot!

PR #1478, opened 2017-02-06
fix for line height calculation

kasparsj:
fix for TextEngine, to correctly set line height, when there are multiple font sizes on one line


MSGhero:
Oh wow, I didn't notice this before. This has been merged in #1669 and #1674, but I guess I should have looked through existing PRs first.

PR #1493, opened 2017-02-14
Fix/no caret

NuclearCookie:


NuclearCookie:
Wrong base... sorry Ignore this

PR #1496, opened 2017-02-14
allow to pass "assetsPrefix" in flash target via parameters

### LINKED ISSUE #1494: Cannot pass assets base path, when loading swf into another
I'm loading my openfl project compiled swf into a container swf, but when doing so it tries to load assets relative to my container swf (which is wrong). Is there a way to pass assets basePath in parameters?

kasparsj:
fixes #1494 
depends on https://github.com/openfl/lime/pull/924


jgranick:
Thank you again for your help in this.

We have a `rootPath` value in our HTML5 embed that allows setting the base asset path, but this is not done at compile-time, but is optionally changed in the HTML template.

Similarly, Flash Player supports a `root` value in its embed that affects relative paths. I have updated both Lime and OpenFL's default Flash web template to include a `root` of ".", which makes it work relative to the SWF URL, not to the page URL. I think this is a much better default, and may resolve the issue you were having?

Otherwise, it should be possible to override the HTML template for Flash, and define a different `root` value. I think that should resolve any asset loading problems :smile:

Issue #1508, opened 2017-02-20
TextField setTextFormat and letterSpacing has something wrong.


hpgood:
var txt:TextField = new TextField();
		var tf:TextFormat = new TextFormat();
		tf.font = "KaiTi_GB2312"; //embed unicode font name
		tf.size = 80;
		tf.color = 0x0000ff;
		tf.letterSpacing = 0;
		
		var tf2:TextFormat = new TextFormat();
		tf2.font = "KaiTi_GB2312";//embed unicode font name
		tf2.size = 80;
		tf2.color = 0xff0000;
		tf2.letterSpacing = 80;
		
		var tf3:TextFormat = new TextFormat();
		tf3.font = "KaiTi_GB2312";//embed unicode font name
		tf3.size = 80;
		tf3.color = 0x00ff00;
		tf3.letterSpacing = 80;

		
		txt.defaultTextFormat = tf;
		
		txt.embedFonts = true;
		txt.text = "a○b123";
		txt.x = 0;
		txt.y = 100;
		txt.width = 500;
		txt.height = 500;
		

		this.addChild(txt);
		
		
		txt.setTextFormat(tf2, 0, 1);
		#if android
		txt.setTextFormat(tf3, 1, 4);// 4 5
		#end
		#if flash
		txt.setTextFormat(tf3, 1, 2);
		#end

 it is ok fine in flash platform,but it is wrong in android platform.

hpgood:
![20170220](https://cloud.githubusercontent.com/assets/3827559/23113798/948d30e4-f775-11e6-8803-f9dfea9babd3.png)  it render in flash  platform.

but in android ,it is wrong.

openfl 4.7.3
lime    3.7.2
hxcpp 3.4.49
openfl test android -x -v -Dnext -debug

hpgood:
i add a pull request to fix this:
https://github.com/openfl/openfl/pull/1509

jgranick:
Closing in favor of https://github.com/openfl/openfl/issues/1700

PR #1509, opened 2017-02-20
Implementation of letter spacing function

hpgood:
letterSpacing function.
and multiple format of text bug fix.

this pull request + https://github.com/openfl/openfl/commit/36ba8a66ecb3d03fbc9a240b30a67a480925f719
can fix the bug:
https://github.com/openfl/openfl/issues/1508

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -681,6 +681,7 @@ class TextEngine {
 		var leading = 0;
 		var ascent = 0.0;
 		var descent = 0.0;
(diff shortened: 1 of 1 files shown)


loudoweb:
Cool. I need to try that!
It would be also great to add to it to the [html parser](https://github.com/openfl/openfl/blob/develop/openfl/_internal/text/HtmlParser.hx) and to swf lib ([here I guess](https://github.com/openfl/openfl/blob/develop/tools/format/swf/tags/TagDefineFont2.hx)).

PR #524, opened 2015-03-01
Event capture added for Event.ADDED_TO_STAGE JS target

peteshand:
Currently OpenFL doesn't support event capture for the html5 target. The changes in this pull request resolves this issue for Event.ADDED_TO_STAGE. This is a critical fix for the Robotleg Framework https://github.com/robotlegs/openfl-robotlegs-framework. Without this change the framework doesn't work while publishing to html5.

PR #1510, opened 2017-02-20
Feature/add event capture phase on added to removed from stage events

NenadBojkovski:
Having capture phase of ADDED_TO_STAGE and REMOVE_FROM_STAGE events is really important to have robotlegs framework working on top of openfl. There was a fix by peteshand similar to this one few years ago but it wasn't merged. Here is the link to it https://github.com/openfl/openfl/pull/524/files.

This fix IMHO is a bit cleaner and more performant, so I hope it is gonna see the light of the day and be merged :).

Any comments or improvements are really welcome and highly appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,41 @@
+4.8.1 (02/15/2017)
+------------------
+
+* Implemented sound.loadPCMFromByteArray
+* Improved behavior of sound.loadCompressedDataFromByteArray
+* Fixed version check when running `openfl upgrade`
+* Fixed loading cross-domain images on HTML5 without using CORS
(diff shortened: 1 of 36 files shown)


jgranick:
Thank you, we should have a capture phase implemented for these events now. It is giving me the same behavior in both Flash and Neko now, let me know how it works with RobotLegs :smile:

NenadBojkovski:
We already use it on our openfl dev branch and it works good so far. We had a case where we needed a capture phase of a custom event dispatched from a sprite, but we've fixed it by removing that specific exotic case from our code. But it might make sense to implement a capture phase for all view events. But that will be a bigger effort.

PR #1516, opened 2017-03-05
depending on server-side websocket impl, can be string or bytearray

sbimikesmullin:
needed this for nodejs sockjs server lib compatibility
see also: https://github.com/sockjs/sockjs-client/issues/252

considering using an alternative server-side lib that sends in binary frames exclusively, but submitting this upstream since technically strings are also valid datatypes in the browser onMessage callback api.


sbimikesmullin:
not sure why the build passes for html5 but fails for cpp and neko, since i didn't change anything cpp or neko related. assuming that `master` generally isn't building already, since i forked master.

PR #1518, opened 2017-03-07
Fix x and y pos being added to width and height.

vizanto:
It’s relative to parent after all. Or am I missing something here?

This bug is easy to test by creating a Sprite with some graphics:
- with `y = 0`, height will be equal to Graphics
- with `y = 1`,  height will be 1 + Graphics height...


vizanto:
Issue with doc build

jgranick:
Do you have example code of a use-case this fixes?

vizanto:
I was placing objects below each other in a loop and the y-position kept growing exponentially.

jgranick:
Merged, I thought this was going to cause a regression in the `textField.getBounds` behavior, but I was wrong. Makes sense :smile:

PR #1521, opened 2017-03-10
Fix shader initialization

justin-espedal:
Initialize shaderData in constructor because the subclasses don't actually have the __data field.

Before this fix, openfl.display.Shader has `__data = new openfl.display.ShaderData (null);` while subclasses are unmodified.

After this fix, all subclasses instead initialize their own custom ShaderData object in their constructor.


justin-espedal:
Given the recent ShaderMacro refactor, I'm closing this.

PR #1522, opened 2017-03-10
Let multiple parent blend mode updates be picked up.

justin-espedal:
Differentiate between an object's own BlendMode and its inherited BlendMode, so if a parent's blendmode is updated multiple times, each update is passed to the children instead of only the first one.

Prioritize the inherited BlendMode over the object's own BlendMode for closer resemblance to flash.

### DIFF
--- a/openfl/display/DisplayObject.hx
+++ b/openfl/display/DisplayObject.hx
@@ -83,6 +83,7 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if openf
 	
 	private var __alpha:Float;
 	private var __blendMode:BlendMode;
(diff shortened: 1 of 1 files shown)


jgranick:
Thank you

I have updated things to something to similar to your approach, but I have fixed the issue where an object's `blendMode` property returns a value different than what a user set.

Here's the commit:

https://github.com/openfl/openfl/commit/0e321560a6b3095fc873ee0b1b0b4725756e5ad2

This doesn't fully resolve multiple layers of blending, but should (I think) help make things work better for you :smile:

justin-espedal:
I don't expect it to be perfect, just wanted something better than what was there. Thanks!

Issue #1464, opened 2017-01-30
Suggestion: ability to launch openfl apps from anywhere


IBwWG:
On openfl 4.7.3, lime 3.7.2, haxe 3.4.0rc2, and hxcpp 3.4.43 on Linux Mint 18.1 MATE x64, you can't launch an openfl app from somewhere other than its own directory:

```
/DisplayingABitmap $ Export/linux64/cpp/release/bin/DisplayingABitmap 
[DefaultAssetLibrary] WARNING: Could not load asset manifest (bytes was null)
[lime.utils.Assets] There is no IMAGE asset with an ID of "assets/openfl.png"
```

If OpenFL/Lime expect the local assets directory to be there, shouldn't they look for it relative to the binary's path?

(Note: I suppose this belongs in Lime's domain, but I can't verify because lime-samples is currently broken.)

jgranick:
`Sys.executablePath` was unreliable, and caused things to break. Otherwise, we would definitely have this feature :smile:

However, the new Haxe 3.4 builds have `Sys.programPath`, which I hear is supposed to be better. If you want to test, add something such as `Sys.setCwd (Sys.programPath ())` to "lime/templates/haxe/ApplicationMain.hx" before it begins preloading.

Please let me know if you try this and it works! 😄

IBwWG:
That didn't seem to do anything.

OTOH, neither did a `trace(Sys.programPath());` line, either.  Nor even a `trace("hello");`.  Do traces not work from within the hx template?

IBwWG:
I seem to need help here.  Nothing I do to this file seems to have any effect, including deleting all its contents.

```
$ haxelib list
flixel-addons: [2.2.0]
flixel: [git]
hxcpp: [3.4.43]
lime-samples: 3.4.0 [git]
lime: 2.9.1 [3.7.2]
openfl-samples: [4.7.0]
openfl: 3.6.1 [4.7.3]
$ cat /usr/lib/haxe/lib/lime/3,7,2/templates/haxe/ApplicationMain.hx
$ rm -rf Export/
$ openfl test -debug linux
...
[no errors, working DisplayABitmap app window comes up]
```

OK, what obvious thing am I forgetting here?

IBwWG:
[...]

PR #1538, opened 2017-03-28
Implement #1464

IBwWG:


IBwWG:
OK maybe someone with write access can do me a favour and resolve the conflict?  Github insists on keeping the newline at the end of the file when I use the online editor.

jgranick:
Thank you!

I added this in a feature-branch for OpenFL, I have had trouble in the past with this type of a call on some platforms. I will try and test this more on additional platforms to make sure it is stable, then we'll merge when we're sure it's ready :smile:

PR #1551, opened 2017-04-17
ATF support

hklindworth:
Supports compressed textures on HTML5 and Neko targets. Can read block compressed textures in Adobe Texture Format (ATF). Cube Maps are not supported.

Compiles fine against lime master branch, but needs lime on develop branch for Neko targets and lime branch feature/fast-webgl-bindings to support HTML5.


jgranick:
Thank you :smile:

PR #1556, opened 2017-04-20
fix for missing graphicsdata

sbijoshj:
`readGraphicsData` returns with missing commands and data unless we set these vectors.
Here is one possible solution.


jgranick:
Thank you :smile:

This lead to realizing that a null Vector `toString` was failing. I fixed that, and it looks like Flash `GraphicsPath` creates `commands` and `data` once you use `curveTo` (or another method)

sbijoshj:
No problem :) thanks!

PR #1573, opened 2017-05-25
Fix integration with resource preloader.

oselivanov:
Fixes integration with resource preloader. 

Without matching resource loader code by setting bitmap.__image.crossOrigin in renderImage to "Anonymous" that image gonna be always fetched and not taken from cache.


oselivanov:
In our app we use preloader and then access preloaded images. In HTML5 build it leads to noticeable flickering because at some point in preloader crossOrigin = "Anonymous" was added, but OpenFL code doesn't match that yet.

jgranick:
Thanks guys :smile:

oselivanov:
Thank you!

PR #1589, opened 2017-06-01
Capture phase implemented for displayObject's events. 

NenadBojkovski:
Hi @jgranick!
I implemented a capture phase for all events which are fired from/through display objects, not only for ADDED_TO_STAGE and REMOVED_FROM_STAGE. Finally, the openfl events work exactly the same as flash events through all event phases. 
Please check the pull request and let me know if there are any issues. 
I could provide flash and openfl test example if you like to test it.

Best,
Nenad


jgranick:
Merged! Thank you for all the testing and work that went into this :+1:

PR #1593, opened 2017-06-06
Fix DOM render when cacheAsBitmap

kasparsj:
The __cacheBitmap does not have a stage reference, but DOMBitmap.render expects it.

### DIFF
--- a/openfl/display/DisplayObject.hx
+++ b/openfl/display/DisplayObject.hx
@@ -769,7 +769,15 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if openf
 		
 		if (__cacheBitmap != null && !__cacheBitmapRender) {
 			
+			var stageReferenceSet:Bool = false;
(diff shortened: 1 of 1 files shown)


jgranick:
Do you think it would work to set `__cacheBitmap.stage = stage` immediately before `DOMBitmap`? That might also fix the issue.

Thanks :smile:

kasparsj:
Yes, but keep the `if (__worldVisible) {`

jgranick:
There we go. `DOMBitmap.render` checks for both `stage != null` and `__worldVisible`, so I think my last commit should work. However, if we need to copy additional properties from the parent object to `__cacheBitmap` for it to work, let me know. We may be missing a cleanup step for DOM to work perfectly

kasparsj:
You'll still need the `if (__worldVisible) {`, because cacheBitmap's `__worldVisible` is always `true`, even if the DisplayObject's is `false`.

PR #1599, opened 2017-06-09
Adds frame script support (JS/HTML5 target only!)

sbimikesmullin:
targets not using this feature shouldn't notice the difference.

- assumes you inject your own frame scripts as strings into SWFLite `.dat` format
  as `SpriteSymbol.frameScriptDefs:Map<Int,String>`
  (our proprietary injection tool needs cleanup before it can be shared)
- evals string as javascript `Function` at runtime during frame animation
- provides context (this == current `MovieClip` instance)
- provides error handling w/ stack traces

ideas to make compatible with other targets:
- provide macros for them at runtime that eval in specific languages that make sense (e.g., for neko/cpp maybe lua script evaluation), or;
- provide hxscript implementation that provides fast runtime evaluation in all targets

:muscle: credit goes to each of: @asciiascetic, @sbijoshj, and @sbimikesmullin for this one.

### DIFF
--- a/openfl/_internal/swf/SWFLite.hx
+++ b/openfl/_internal/swf/SWFLite.hx
@@ -181,6 +181,39 @@ import openfl.Assets;
 	}
 	
 	
+	private static function __unserializeScripts (spriteSymb:SpriteSymbol) {
+
+		var keys = spriteSymb.frameScriptDefs.keys ();
+		if (!keys.hasNext()) return;
+
+		spriteSymb.frameScripts = new Map ();
+
+		for (k in keys) {
+
+			var s = spriteSymb.frameScriptDefs[k];
+
+			try {
+
+#if js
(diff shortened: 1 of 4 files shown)


sbimikesmullin:
i rebased this one several times to improve aspects of presentation. refresh to make sure you are looking at current.

jgranick:
Thank you! I have just made an initial commit, merging `SpriteSymbol` frame script support. It should work with both `frame.script` as a Haxe function, or `frame.scriptSource` as a `String`. 

In the future, we might consider different script types, but for now this should work with straight JavaScript. The order of the code is a bit different than the original pull request, but the core logic should be here.

We might want to consider a way to reuse the same scripts, and pass a scope when they are executed, but for now this is likely fine. I like the idea of HScript, but I don't know how we could presume what variables need to be bound to the script 😕

sbimikesmullin:
we pass in the MovieClip instance as the value of `this`. not sure what other variables there would be to consider

PR #1600, opened 2017-06-10
`Asset.getMovieClip()` on `.bundle/` should load globally

sbimikesmullin:
prior to this change, if we did not keep our own reference to a `.bundle` loaded at runtime, we were not able to access to again from any other scope. 

this adds it to the Lime/OpenFL AssetLibrary Map so we can find it by its url when using `Asset.getMovieClip()`

### DIFF
--- a/openfl/display/Loader.hx
+++ b/openfl/display/Loader.hx
@@ -278,8 +278,12 @@ class Loader extends DisplayObjectContainer {
 			if (Std.is (library, AssetLibrary)) {
 				
 				library.load ().onComplete (function (_) {
+					var _library = cast (library, AssetLibrary);
 					
-					contentLoaderInfo.content = cast (library, AssetLibrary).getMovieClip ("");
+					contentLoaderInfo.content = _library.getMovieClip ("");
+					
+					Assets.registerLibrary(Path.withoutDirectory(Path.withoutExtension(manifest.rootPath)), _library);
+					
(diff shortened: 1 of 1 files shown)


jgranick:
I added automatic registration of loaded `AssetLibrary` instances, but I'm using `contentLoaderInfo.url` for now, which should prevent collision (to my knowledge) between asset libraries. Hopefully it should :smile:

Now you should be able to `Assets.hasLibrary (loader.contentLoaderInfo.url);` or `Assets.getLibrary`, or `Assets.getMovieClip (loader.contentLoaderInfo.url + ":MySymbolName");`, etc

I tried to unregister the library on `loader.unload`, I'm not sure if this is consistent with how it would work on AS3?

You should also be able to `AssetLibrary.loadFromFile ("path/to/asset.bundle").onComplete (function (library) { ... });` which might be a little clearer

sbimikesmullin:
@jgranick we ended up being forced to change this to 
```
					Assets.registerLibrary(manifest.name, _library);
```

because when we upload the files to cdn we want to use a crypted/hashed name. we flatten the directory structure and we hash the filename to make it an a) unpredictable, b) non-linear sequence. but this made the library names get hashed, too.

we decided it was better to hash the filename and rely on the `name:` value inside `.bundle/library.json`

recommending we update your version as well

PR #1601, opened 2017-06-10
openfl_dynamic Reflect.fields() only

sbimikesmullin:
we wanted dynamic `Reflect.field()` additions for all `(MovieClip.__children:DisplayObject).__name`s, but not the `MovieClip` to be `implements Dynamic<DisplayObject>`, because not all our dynamic properties are `DisplayObjects`; we don't think Flash views them that way, either.

### DIFF
--- a/openfl/display/MovieClip.hx
+++ b/openfl/display/MovieClip.hx
@@ -406,7 +406,7 @@ class MovieClip extends Sprite #if openfl_dynamic implements Dynamic<DisplayObje
 		
 		displayObject.visible = frameObject.visible;
 		
-		#if openfl_dynamic
+		#if (openfl_dynamic || openfl_dynamic_fields_only)
 		Reflect.setField (this, displayObject.name, displayObject);
 		#end
 		


jgranick:
The issue here is that dynamic fields will not work properly with getters or setters unless they are cast to the appropriate object type. For example:

```haxe
clip.myChild.myChild.x = 100;
```

This will silently fail on JavaScript if the field type was `Dynamic`, but if it is a `DisplayObject` it will translate to `set_x (100);` and work properly.

I believe you should be able to cast to a different object type if it isn't a `DisplayObject`... I'm open to ideas :smile:

sbimikesmullin:
we noticed that setters don't work on Dynamic so we've just taken to syntax like this:
```
(cast movieClipInstance.childName1:MovieClip).visible = true;
```
or
```
(untyped movieClipInstance.childName1.childName2:MovieClip).visible = true;
```

we've actually tried like 4-5 different ways of dealing with this, including defining interfaces, extern classes, `using` keyword, and others... but this ended up being the simplest in practice so far.

sbimikesmullin:
@jgranick what do you think about `MovieClip` being `implements Dynamic<DisplayObjectContainer>`, instead of `DisplayObject`? Can't hurt, right? Seems much more helpful!

PR #1602, opened 2017-06-10
add linkageBaseClass name MovieClip constructor support

sbimikesmullin:
For example, we have a subclass of `MovieClip` which initializes custom Button objects with click sounds and common behaviors.

Without this change, attaching these AFTER `MovieClip` is instantiated is your only option--which is too late. You really need a way to inject some code as part of the `MovieClip` constructor, and subclasses or "Base Class" as Flash called it, are ideal. Especially since the artist can define which ones they mean during their creative process, and developers can just pull it out of the .swf.

This implies that the `.linkageBaseClass:String` names are being added to the correct spot in the `SWFSymbol` classes instantiated by the SWFLite `.dat` file format. Currently we use a proprietary script to inject these values into our `.bundle/swflite.dat` after `haxelib run openfl process` is done with them, but before runtime downloads them from CDN.

We may release that injector script once its cleaned up, or we may introduce something cleaner that can export `.FLA` to `.dat` in one pass that has everything in it.

This code should be merge-able as-is today, however. Shouldn't hurt anyone not using the feature.

### DIFF
--- a/openfl/_internal/symbols/SpriteSymbol.hx
+++ b/openfl/_internal/symbols/SpriteSymbol.hx
@@ -17,7 +17,7 @@ class SpriteSymbol extends SWFSymbol {
 	
 	
 	public var frames:Array<Frame>;
-	
+	public var linkageBaseClass:String;
 	
 	public function new () {
 		
@@ -37,7 +37,23 @@ class SpriteSymbol extends SWFSymbol {
 		MovieClip.__initSymbol = this;
 		#end
 		
-		if (className != null) {
+		
+		if (linkageBaseClass != null) {
+			
+			var symbolType = Type.resolveClass (linkageBaseClass);
+			
+			if (symbolType != null) {
+				
+				movieClip = Type.createInstance (symbolType, []);
+				
+			} else {
+				
+				trace("Error: Could not resolve .linkageBaseClass = \""+ linkageBaseClass +"\" for symbol \"" + className + "\"!");
+				
+			}
+			
+		}
+		else if (className != null) {
 			
 			var symbolType = Type.resolveClass (className);
 			


jgranick:
How do you feel about `linkageBaseClass` taking less priority than `className`?

...or what do you think about using `className` for custom base classes?

sbimikesmullin:
AFAICT `className` is the Symbol instance name. If so:
- These typically begin with a lowercase, in our practice. Violates Haxe rule of class name identifiers.
- They are not required to be unique by Animate CC.
- There is no analogy for this in the AS3/Flash world; the only way to link an AS3 class to a runtime symbol instance is via Base Class, Document Class, or explicitly via AS3 frame script code. This would be unique to OpenFL.

So my feeling is that `className` should not even be considered here, and we would remove its `else if` block.

sbijoshj:
I might be missing something but I believe using `className` first instead of `linkageBaseClass` would result in completely skipping and missing the custom base class code on that symbol.
But using `className` for custom base classes is actually probably how flash does it and I can see that working.

sbimikesmullin:
Ok, I was wrong above. 

First, linkage names can also point to AS3 classes in very specific circumstances, outlined below.

Animate CC does not allow one to simultaneously have a valid *Symbol linkage* name that maps to an AS3 class, when you also have a *Symbol base class* name that maps to a valid AS3 class. You can only have both if the *Symbol linkage* name is not referencing a valid AS3 class name.

THEREFORE, the logic should be if (linkage != null && baseClass != null) then instantiate from baseClass name, else if (linkage != null && baseClass == null) then instantiate from linkage name.

It should be important to differentiate between *Symbol instance* name and *Symbol linkage* name, though. Currently `className` does not permit us to do that. Ideally we'd have my four bulleted name properties in the previous comment as separate key values on the `SpriteSymbol` class, instead of trying to mix them. Then we couldn't accidentally try to instantiate a *Symbol instance* name with a matching class name. That is when accidents would occur.

sbimikesmullin:
Another discovery as I am mucking about in the AS3 ABC bytecode AVM2 instructions, Flash creates a set of "Scripts" bytecodes which are used to initialize the scene. Decompilers like JPEXS don't show these to you in the GUI, but there are 1-2 that are always there.

More importantly, every Sprite/MovieClip/SymbolInstance that has an entry in the SymbolClass [SWF tag] table is there because it has a corresponding AS3 class in the bytecode. Adobe Flash Pro / Animate CC creates these for you automatically on publish but does not require or even allow you to save all of these to disk with the .FLA. From an OpenFL standpoint, we may not even need to be aware of them, depending on the goal. Currently, because we aren't reading the ABC, we assume what initialization must need to be done and do it automatically, but Flash uses ABC to do this.

Sooo that means that I was wrong about Symbol instances not having AS3 classes at runtime. Technically, they do, but on our project we're choosing not to use them. But it means `className` could be used as-is (Symbol linkage name or else Symbol instance name; with no Symbol base class name), depending on how literally you want to implement ABC. Because the ABC clarifies the rest from there (from the SymbolClass table entry). As of current `develop`, I can confirm OpenFL ABC support is both commented out and incomplete.

But here's what is important to keep in mind about that. A SymbolInstance with no frame scripts (presumably there may be more conditions triggering automatic AS3 class creation than this, but this is the only one I care about atm) will not have an entry in the SymbolClass table. Therefore its `className` does not matter and should get ignored. But any Symbolnstance with a SymbolClass entry will 100% guaranteed have an AS3 class in the `abcFile:ABCData`  even if it has to be an ugly namespace like the following actually generated class:

```
package avatarSelect_fla
{
   import flash.display.MovieClip;
   
   public dynamic class downArrow_sml_btn_mouseCont_76 extends MovieClip
   {
```

Of course, in Haxe that class name would start with an uppercase "d". In Javascript or HScript, it would work as-is.

This is a spot where we may want to seize upon an opportunity to diverge from Flash (until/unless we get deep Animate CC plugin integration). Certainly, for the current project I'm working on, we have decided to diverge in the following strategy.

In our project we decided since we cannot compile an .FLA with complex reusable ActionScript3 Base Class references without exporting/loading them from the disk as `.as` files, it makes the most sense to store those `.as` files with the rest of our application sources and revision them. We could have included them into one .swf, exported, and shared them from other SWFs, but in practice we ended up including a copy of them into the ABC code of every .swf that referenced them.

Regardless, when we translated our existing project to Haxe + OpenFL, we used the `HaxeFoundation/as3hx` project, and it was able to translate those AS3 classes we had been using as ABC Base Classees. Since there will be a penalty paid between compiled Haxe and runtime eval() interpreted performance, the ideal is that ALL ABC could be converted to compilable Haxe; but this presumes that the resulting Haxe target can be loaded modularly (well, in our case we have a lot of frequently updated content, so we distribute it via CDN and would do so regardless of html5 or native binary, thats why this approach makes sense). Right now, all we have is the SWFLiteExporter's `.bundle/` format provided by `haxelib run openfl process` cli. So this means the runtime has to interpret the code in those files.

Therefore, we have decided that any reference to shared/reusable ABC code may as well be translated to .hx, while frame scripts can be translated to Javascript (or, perhaps in the future, another fast platform-agnostic interpreted runtime like HScript). Any frame scripts that have complex logic we can also move out into the reusable .hx class namespace. This limits runtime interpretation to the simplest code, and it also simplifies how much of the SWF ABC has to be disassembled correctly at compile time.

Particularly, all we care about then is the Symbol Base Class name strings, and the disassembled frame script (from ABC to JS).

But with respect to to the future and goals of OpenFL, this may not work for everyone. I mean it could. Actually, compared to the alternatives its probably the fastest option at runtime performance-wise. I think its a good compromise.

Particularly, I mean, this one of the reasons is why Flash has died, right? Apple famously rejected Adobe Flash on their mobile iOS platform for being too heavy on the CPU processing. That's because Flash did all sorts of runtime optimizations so artists didn't have to. Simultaneously Steve Jobs publicly endorsed HTML5 as the alternative. How is it better? It doesn't do much if ANY optimization at runtime; any optimizations happen much earlier in the production cycle, which lets mobile ARM chips rest easy. Instead optimization is done by the artist when the art is created (similarly to how html chop shops take .PSD artwork and "slice" it into a set of image atlases and CSS3 stylesheets). The artist becomes semi-fluent in code, and can basically describe what areas to cache, and how (9-slice, 3-slice, hover states, scaleX/Y/XY, or no-repeat), and this is stored with the art data in advance. Now, if we don't aggressively add that type of instruction (such as cacheAsBitmap, or the current trend of lets-make-everything-a-bitmap-sprite-atlas), then the HTML5 WebGL Canvas performance completely suffers. You end up with a near-pixel-perfect visual match, but a third to half the FPS you used to get with Flash Player. This can only get worse if we add an AVM2 runtime on top of all that.

So personally, I think it behooves us to implement a change to the tooling one way or another, so that its easier for an artist to make these indications and carry them through to production data formats. I have ideas about how Animate CC could change to facilitate that, but until then my proposal above lets us meet half-way.

In short, I'm not sure if we'll be forced to keep this on our own branch or not. Curious about your thoughts?

jgranick:
The behavior here in the class does not actually require an additional field (as you pointed out), but I think for the sake of mental clarity (regarding instance vs. linkage vs. base classes) I think a second field is justified here. I've just merged support using `baseClassName` as the name. The maintained idea in the code (currently) is that a `symbol.className` is unique among symbols, but multiple MovieClip symbols could share the same `baseClassName`, so I think that we have the same core behavior we need

We're maybe safe not preserving unique class names for instances that aren't given a name manually in the IDE. However, perhaps we should look into this more for the "generate" option, we could generate classes for all symbols, and that would allow for better code completion and child field access?

sbimikesmullin:
there are unique names generated by Animate in the SymbolClass tag list. everything that needs an as3 class is there.

everything else should just be referenced relatively like `.getChildAt()` and `.getChildByName()`. like css or xpaths. i don't think we should presume to generate unique names for those too.

PR #1603, opened 2017-06-10
these are passed needlessly; already assigned

sbimikesmullin:
we feel these lines are duplicate, or is it just us?

see lines 72-73 of same file


jgranick:
This looks repetitive in this case, but I believe `__fromSymbol` is also used when you use "generate" to create SWF-based classes, so we need it :smile:

PR #1604, opened 2017-06-10
support for shared font names in swflite

sbimikesmullin:
A little-known and poorly-documented feature in Adobe Flash/Animate CC.

Lets you declare fonts in a single .SWF and reuse them in subsequently loaded .SWF. 

Saves bandwidth.

Lets you centralize your creative font collection and update them in a single place.

Notice the way we match names; sneaky, but effective!

### DIFF
--- a/tools/format/swf/exporters/SWFLiteExporter.hx
+++ b/tools/format/swf/exporters/SWFLiteExporter.hx
@@ -706,6 +706,11 @@ class SWFLiteExporter {
 			
 		}
 		
+		if (tag.hasFontClass)
+		{
+			symbol.fontName = tag.fontClass;
+		}
+		
 		var bounds = tag.bounds.rect;
 		symbol.x = bounds.x;
 		symbol.y = bounds.y;
--- a/tools/format/swf/instance/DynamicText.hx
+++ b/tools/format/swf/instance/DynamicText.hx
@@ -102,6 +102,12 @@ class DynamicText extends TextField {
 			
 		}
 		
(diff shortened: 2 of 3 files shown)


jgranick:
Thank you again

I've just added what I believe will work the same in cases where an inexact match is the only option, but it will search an exact match first. This should help prevent "Arial" from matching "Arial Bold" if both exists, and "Arial Bold" shows up first in the list

tools/format/swf/exporters/SWFLiteExporter.hx
```diff
@@ -706,6 +706,11 @@ class SWFLiteExporter {
 			
 		}
 		
+		if (tag.hasFontClass)
+		{
+			symbol.fontName = tag.fontClass;
```
sbimikesmullin: @jgranick you forgot to include this piece. this is also important because when shared fonts feature is used the name is only present here. from the swf doc:

Field | Type | Comment
-- | -- | --
FontClass | If HasFontClass, STRING | Class name of font to be loaded from another SWF andused for this text

tools/format/swf/instance/DynamicText.hx
```diff
@@ -102,6 +102,12 @@ class DynamicText extends TextField {
 			
 		}
 		
+		if (tag.hasFontClass) {
+			
+			trace("TODO: Implement .hasFontClass for DynamicText");
```
sbimikesmullin: @jgranick this will probably become important for the same reason as above but we haven't run into this case yet so ok to delay

PR #1605, opened 2017-06-10
Fix DisplayObject.removeEventListener() to remove old listeners

sbimikesmullin:
EventDispatcher's removeEventListener was not properly removing old listeners, due to the match() method being used, which incorrectly determined equal scopes to not be equal. All of the logic has been duplicated here, with the exception being how listeners are compared. 

Without this correction, event listener callbacks continued to be added by user input triggers (e.g., hovering over a button) without ever being removed after evaluation. This caused the corresponding event to fire more and more times with every user interaction.

:muscle: all credit to @sidpitt89


sbimikesmullin:
🛑 ✋ 👮 we found a problem with this PR and are temporarily revoking it

PR #1607, opened 2017-06-10
refactor MovieClip to fix several bugs

sbimikesmullin:
Refactor `MovieClip.__enterFrame()` and friends to fix:
- `FrameObject` `CREATE` duplication
- `FrameObject` `REMOVE` failures, 
- bugs related to transposed zero-indexed/one-indexed frame number references
- bugs in frame script playback synchronization with animation

Added a **ton** of comments to improve maintainability.

Tried to code in a way that made it more obvious what was happening, with only mild performance optimizations in mind. Should be as-fast or faster than before.

The only major change is that all `MovieClip.__children:DisplayObject` instances are initialized from `MovieClip.new()` constructor now. This happens once instead of multiple times throughout animation loop/playback. We believe its a sensible approach inspired by other implementations, and a performance optimization (read comments in patch for detail).

The primary reason we needed this was to eliminate several bugs that were breaking playback for a very complex set of `.SWF` files we had imported using `SWFLiteExporter` and were playing back at runtime in the browser with OpenFL on the HTML5 target.

There are still issues to be resolved with the [Filters and Blend Modes](https://github.com/openfl/openfl/pull/1374/commits), but we'll get to those soon.

💪 credit @sbimikesmullin with minor contributions from **Elvir Tatarevic** and @sbijoshj.

💡 Interested in whatever feedback good/bad you have on this one @jgranick. Could go over it in detail on call if you want, and show before/after behavior.

🛑 ✋ 👮 **WARNING**: This PR depends on #1599 and #1601, as written. You should merge those first. They may cause this one to conflict but shouldn't because they have the exact same lines.

### DIFF
--- a/openfl/display/MovieClip.hx
+++ b/openfl/display/MovieClip.hx
@@ -1,7 +1,10 @@
 package openfl.display;
 
-
 import lime.utils.Log;
+import openfl.errors.ArgumentError;
+import openfl._internal.symbols.SWFSymbol;
+import openfl._internal.timeline.Frame;
+import haxe.CallStack.StackItem;
 import openfl._internal.swf.SWFLite;
 import openfl._internal.symbols.BitmapSymbol;
 import openfl._internal.symbols.ButtonSymbol;
@@ -20,7 +23,6 @@ import openfl.text.TextField;
 import hscript.Interp;
 import hscript.Parser;
 #end
-
 
 #if !openfl_debug
 @:fileXml('tags="haxe,release"')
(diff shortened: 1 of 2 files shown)


tools/format/swf/instance/MovieClip.hx
```diff
@@ -47,7 +47,7 @@ class MovieClip extends flash.display.MovieClip {
 	@:noCompletion private var activeObjects:Array<ChildObject>;
 	
 	#if flash
-	@:noCompletion private var __currentFrame:Int;
+	@:noCompletion private var __currentFrameOneIndexed:Int;
```
sbimikesmullin: 👷 🚧 **WARNING**: I do not feel I understand the tools version of the `MovieClip` class, and while these changes seem to satisfy the build, you should really check them to be sure the original intent is still functional. With or without them, I couldn't tell the difference. 

It would help to get an explanation from you @jgranick here.

openfl/display/MovieClip.hx
```diff
@@ -26,704 +28,996 @@ import openfl.text.TextField;
...
-					
-					if (displayObject != null) {
-						
-						__placeObject (displayObject, frameObject);
-						
+
+					if (null != duplicate) {
+
+						trace("Multiple CREATE tags for the same characterId and depth are being merged.\n"
+							+ "Asset \""+ untyped this.__swf.library.rootPath +"\", "
```
sbimikesmullin: this trace statement should be removed once done debugging

openfl/display/MovieClip.hx
```diff
@@ -26,704 +28,996 @@ import openfl.text.TextField;
...
-								removeChild (timelineObject.displayObject);
-								
+					else {
+
+						// instantiate child's DisplayObject first and only time
+						symbol = __swf.symbols.get (frameObject.symbol); // lookup Symbol class
+
+						if (null == symbol) {
+
+							trace("Unable to CREATE DisplayObject instance; got NULL symbol\n"
```
sbimikesmullin: once we figure out why null symbols exist (you may already know, i haven't had time to figure out yet), we should remove this unless/except debugging.

openfl/display/MovieClip.hx
```diff
@@ -26,704 +28,996 @@ import openfl.text.TextField;
...
+
+						}
+						else {
+
+							displayObject = symbol.__createObject (__swf); // instantiate Symbol
+
+							// TODO: find out why some displayObject instances are null
+							if (null == displayObject) {
+
+								trace("Unable to CREATE DisplayObject instance; got NULL __createObject()\n"
```
sbimikesmullin: same here

openfl/display/MovieClip.hx
```diff
@@ -26,704 +28,996 @@ import openfl.text.TextField;
...
+					// no CharacterId is required.
+
+					// rule: depth must already exist
+					displayObjectCache = null;
+					if (indexActiveFrameObjectEntryById.exists (frameObject.id)) {
+						frameObjectEntry = indexActiveFrameObjectEntryById.get (frameObject.id);
+						displayObjectCache = frameObjectEntry.displayObjectCache;
+					}
+					if (null == displayObjectCache || null == displayObjectCache.displayObject) {
+						trace("Tried to UPDATE a DisplayObject child that hasn't been CREATED yet. "
```
sbimikesmullin: this may benefit from an `#if debug` block

sbimikesmullin:
rebasing this one now

PR #1608, opened 2017-06-12
Disable leading on embedded text

sbimikesmullin:
We noticed a difference comparing our original .swf in Flash to our .dat in OpenFL, and it was caused by this line.

Perhaps because we use shared fonts instead of embedded fonts? Not sure. Seems the original author isn't sure, either. So we commented this line out.

All we know is, it doesn't appear that Flash does `format.leading += 4`.

If we had to guess, Flash shows a leading of 1 while OpenFL shows 5... in the end that messes up the text height.

:muscle: credit **Elvir Tatarevic**


jgranick:
Thanks :smile:

PR #1609, opened 2017-06-13
OpenFL should fail silently when fn missing from `ExternalInterface.call()`

sbimikesmullin:
Flash fails silently on all errors from `ExternalInterface.call()`, including `function not defined`.

http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/external/ExternalInterface.html#call()

However, now that we're in JS target, there is no concept of inside/outside of vm. Therefore, it makes sense to forward any exceptions thrown by the external functions. So we'll allow that, but we'll continue to ignore any errors about missing functions, since some [translated] code [likely] relies on that behavior still.

### DIFF
--- a/openfl/external/ExternalInterface.hx
+++ b/openfl/external/ExternalInterface.hx
@@ -39,30 +39,37 @@ import openfl.Lib;
 				functionName += '.bind(${thisArg})';
 			}
 		}
-		
+
+		var fn:Dynamic = js.Lib.eval (functionName);
+		if (Type.ValueType.TFunction == Type.typeof(fn)) {
+			// Flash does not throw an error or attempt to execute
+			// if the function does not exist.
+			return null;
+		}
+
 		if (p1 == null) {
 			
-			callResponse = js.Lib.eval (functionName) ();
+			callResponse = fn ();
 			
 		} else if (p2 == null) {
 			
-			callResponse = js.Lib.eval (functionName) (p1);
+			callResponse = fn (p1);
 			
 		} else if (p3 == null) {
 			
-			callResponse = js.Lib.eval (functionName) (p1, p2);
+			callResponse = fn (p1, p2);
 			
 		} else if (p4 == null) {
 			
-			callResponse = js.Lib.eval (functionName) (p1, p2, p3);
+			callResponse = fn (p1, p2, p3);
 			
 		} else if (p5 == null) {
 			
-			callResponse = js.Lib.eval (functionName) (p1, p2, p3, p4);
+			callResponse = fn (p1, p2, p3, p4);
 			
 		} else {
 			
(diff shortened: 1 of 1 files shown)


sbimikesmullin:
so i recommend we revert/rebase out my PR above and choose one of the following

option A:

```
		// Flash does not throw an error or attempt to execute
		// if the function does not exist.
		var fn:Dynamic;
		try {

			fn = js.Lib.eval (functionName);

		} catch (e:Dynamic) {

			return null;

		}
		if (Type.ValueType.TFunction == Type.typeof(fn)) {

			return null;

		}

		if (p1 == null) {
			
			callResponse = fn ();
			
		} else if (p2 == null) {
			
			callResponse = fn (p1);
			
		} else if (p3 == null) {
			
			callResponse = fn (p1, p2);
			
		} else if (p4 == null) {
			
			callResponse = fn (p1, p2, p3);
			
		} else if (p5 == null) {
			
			callResponse = fn (p1, p2, p3, p4);
			
		} else {
			
			callResponse = fn (p1, p2, p3, p4, p5);
			
		}
```

option B:

```
		// Flash does not throw an error or attempt to execute
		// if the function does not exist.
		if (untyped __js__("'function' != typeof {0}", functionName)) {
			return null;
		}

		if (p1 == null) {
			
			callResponse = js.Lib.eval (functionName) ();
			
		} else if (p2 == null) {
			
			callResponse = js.Lib.eval (functionName) (p1);
			
		} else if (p3 == null) {
			
			callResponse = js.Lib.eval (functionName) (p1, p2);
			
		} else if (p4 == null) {
			
			callResponse = js.Lib.eval (functionName) (p1, p2, p3);
			
		} else if (p5 == null) {
			
			callResponse = js.Lib.eval (functionName) (p1, p2, p3, p4);
			
		} else {
			
			callResponse = js.Lib.eval (functionName) (p1, p2, p3, p4, p5);
			
```

jcward:
@mikesmullin @jgranick 

try/catch is typically bad for performance.

`Type.typeof` vs js `typeof` is a meh. Actually the former is more accurate, as you _could_ (why would you, but could) call ExternalInterface.call on a MyClass or Enum constructor (both of which are technically JS functions) -- `Type.typeof` [guards against this](https://github.com/HaxeFoundation/haxe/blob/development/std/js/_std/Type.hx#L191-L194), whereas native `typeof` does not.

So I think maybe just leave it as-was. Sorry for causing chaos. :smile:

sbimikesmullin:
@jgranick but we do have to choose A or B because the version accepted fails to do what was intended; the `js.Lib.eval(functionName)` still throws an exception when `functionName` doesn't exist :picard_facepalm:

sbimikesmullin:
i'll submit 2nd PR for A

PR #1610, opened 2017-06-13
validate function exists for ExternalInterface.call()

sbimikesmullin:
`js.Lib.eval(functionName)` errors if `functionName` does not exist

i meant to have this behavior in #1609 but it escaped me then.

### DIFF
--- a/openfl/external/ExternalInterface.hx
+++ b/openfl/external/ExternalInterface.hx
@@ -40,11 +40,22 @@ import openfl.Lib;
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


mikesmullin:
Grr this will not work either as external function is not defined by openfl. Have to use untyped js type of...

sbimikesmullin:
ok i tested it.

`Type["typeof"](fn) == ["TFunction", 5, __enum__: Object, toString: function]`

so we're good to merge this.

PR #1611, opened 2017-06-14
External Interface function check

sbimikesmullin:
we've tested it with and without external functions at this point, so this should be the last pr on it.

we'll test future PRs for a few weeks before submitting.


jgranick:
:shipit:

PR #1616, opened 2017-06-21
Make bitmapdata draw work with filters when readable is false

Beeblerox:
Currently there is no support for filters in bitmapdata's draw() method.
This pull request adds this feature for cases when bitmapdata isn't readable (so this feature works only with gl renderer).


Beeblerox:
@jgranick if you accept this change then i'll make pull request to make copyPixels work (partially) for cases, when bitmapdata isn't readable

PR #1620, opened 2017-06-27
Sorting of the eventlisteners with same priority doesn't preserve the order they are added for html5 target

NenadBojkovski:
As it is written in the [documentation](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) of the Array's sort method in JavaScript, the order of the elements that have the same value against which the sorting is performed is not garanteed. 
Namely, what we have experienced is when we have multiple event listeners for the same event with the same priority(=0), they are not executed in the order they are added. If you check the list after sort method, at some point, the list is gonna get shuffled. It doesn't happen on every sort call.

### DIFF
--- a/openfl/events/EventDispatcher.hx
+++ b/openfl/events/EventDispatcher.hx
@@ -366,6 +366,7 @@ class EventDispatcher implements IEventDispatcher {
 		this.list = list;
 		
 		active = true;
+		isCopy = false;
 		index = 0;
 		
 	}
--- a/openfl/events/EventDispatcher.hx
+++ b/openfl/events/EventDispatcher.hx
@@ -74,7 +74,7 @@ class EventDispatcher implements IEventDispatcher {
 				
 			}
 			
-			__addListinerByPriority(list, new Listener (listener, useCapture, priority));
+			__addListenerByPriority(list, new Listener (listener, useCapture, priority));
 						
 		}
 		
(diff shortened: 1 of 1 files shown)


MSGhero:
Throwing this out there, but is `haxe.ds.ArraySort.sort()` what you're intending to implement? It's a stable sort.

NenadBojkovski:
If you chech the code carefully there is no explicit sorting. But a listener is added to the proper position in the listeners' list based on priority value. This is more efficient than sorting evertime you add a listener. That's one thing.
The second and more important thing we try to solve is that on html5 target in chrome when you add multiple listeners with same priority value, the sort method at some point will shuffle the array. Couldn't find why I cannot debug the js array's sort method.

MSGhero:
Well what you wrote is essentially `insertionSort` operating on a mostly sorted list, which is fine. I was guessing based off the original code, and either way is fine.

One note is that it is spelled `listener` with an `e`, not an `i`.

NenadBojkovski:
Oh, it's a typo, thanks that you've noticed, I'll fix that.

Regarding the sorting, I won't agree either that this is an insertion sort. It is not a sorting algorithm at the first place. If the list that is sent is unsorted, the method won't sort it. It is much simpler than a sorting algorithm and more efficient. We are taking advantage here that we have complete control over the array and the fact that the elements(listeners) are not changed after being added, so we add them to the list in a sorted manner.

PR #1622, opened 2017-06-27
Rename `clearDirtyFlags` to less ambiguous `hasDirtyFlags`.

sbi-jeff-cook:


jgranick:
Thanks, I've renamed to `clearRenderDirty` which should be a little clearer. `clearDirtyFlags` suggests it might clear other dirty flags, such as `transformDirty`, which it does not :smile:

PR #1633, opened 2017-07-08
fix typo; object needs to be initialized outside of for loop

sbimikesmullin:
minor issue. without this patch, you can see it tries to push to the array before it has been initialized. 

ran into that when trying to `openfl process` some of our `.swf` with this upstream version.


jgranick:
I just hit this today, I'm sorry I did not read the contents of your pull request before getting stuck on this bug myself 😊

PR #1634, opened 2017-07-08
MovieClip implements Dynamic<DisplayObjectContainer>

sbimikesmullin:
since `DisplayObjectContainer` extends `DisplayObject`, i don't see how this could break anything.

but it does seem more useful in practice. any objection?

### DIFF
--- a/openfl/display/MovieClip.hx
+++ b/openfl/display/MovieClip.hx
@@ -30,7 +30,7 @@ import hscript.Parser;
 @:access(openfl._internal.symbols.SWFSymbol)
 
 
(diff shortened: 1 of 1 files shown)


jgranick:
I like how this would allow for `myClip.childClip.childClip.addChild`, but am concerned this would lead to wrongfully treating a child `DisplayObject` (such as `TextField` or `Shape`) as a `DisplayObjectContainer` and causing a runtime crash. This is all dynamic anyway, so there is risk involved, so I'm going to close this issue, but I'm on the fence, personally.

PR #1635, opened 2017-07-08
Add setContent function to Loader

sbimikesmullin:
this came in handy for us. i don't see how it could hurt but lmk if any objection.

not sure if this was some kind of protected access / security vulnerability consideration

### DIFF
--- a/openfl/display/Loader.hx
+++ b/openfl/display/Loader.hx
@@ -217,7 +217,11 @@ class Loader extends DisplayObjectContainer {
 		}
 		
 	}
-	
+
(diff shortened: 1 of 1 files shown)


jgranick:
I would be happy to hear more about the specific use case that brought this up. Perhaps we are populating the initial `content` value with a different kind of instance than your original code expected, and we can discuss that?

I want to be _very_, _very_ careful about adding any new APIs. Usually, we only add static helper methods, such as `BitmapData.fromFile` or things to help behave better with Lime such as `stage.application` and `stage.window`

PR #1636, opened 2017-07-08
translate AVM2 ActionScript3 Byte Code (ABC) to JavaScript

sbimikesmullin:
only the frame scripts, and only the easy stuff.

🚧 👷 **CAUTION!** This branch is still experimental and should be thoroughly tested. 🏗 

💪 **GREETZ:** Thanks to @as3boyan for help with pattern matching.

PR #1637, opened 2017-07-08
extract sound to bundle mp3 files

sbimikesmullin:
- all sounds will be placed inside `.bundle/sounds/*.mp3` where the file basename will equal the linkage class name (assumes all sounds have linkage names; technically may not be always true but has been so far; can't imagine a scenario where sound is used without linkage name; could provide character id as fallback for that scenario if proven possible)
- sounds exported will also be listed inside `.bundle/library.json` `assets:Asset[]` with their name containing  a relative path from the `.bundle/` folder
- .mp3 files just work!
- .wav files need testing. may need their header bytes prefixed when writing to disk. notice that code is there but commented until we get a chance to test further.
- other sound formats are possible but were not used by our codebase so we are skipping implementation for now.
- ideally sounds should all be translated to .mp3 and/or .ogg regardless of what format they were in. perhaps some .exe tool can be provided similar to how webfonts are processed. (ffmpeg is ideal but a HUGE project)

🚧 👷 **CAUTION!** This branch is still experimental and should be thoroughly tested. 🏗 

**TODO:** currently adding support to read and cache the sounds before `.bundle` is finished loading at runtime. all sounds should be preloaded before the `.bundle/` is considered done loading, and able to be referred to and run from anywhere (globally) by their `library + ":" + linkageName`, similar to the way MovieClips are.

**NOTICE:** Depends on PR #1636 being merged first.

### DIFF
--- a/tools/Tools.hx
+++ b/tools/Tools.hx
@@ -659,6 +659,37 @@ class Tools {
 			}
 			
 		}
+
+		createdDirectory = false;
+		for (id in exporter.sounds.keys ()) {
+
+			if (!createdDirectory) {
+
+				PathHelper.mkdir (PathHelper.combine (targetPath, "sounds"));
+				createdDirectory = true;
+
+			}
+
+			var symbolClassName = exporter.soundSymbolClassNames.get (id);
+			var typeId = exporter.soundTypes.get (id);
+			trace("exporting sound id "+id+", type "+ typeId +", symbol class name "+ symbolClassName);
+			var type;
+			switch (typeId) {
+				case SoundType.MP3: type = "mp3";
+				case SoundType.ADPCM: type = "wav";
+				// TODO: this WAV header may become necessary to prepend in some cases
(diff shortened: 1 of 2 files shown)


sbimikesmullin:
this guy's project can convert SWF ADPCM format (currently mistakenly called .WAV by me) to RAW then normal software can convert it from RAW to WAV or MP3

https://github.com/alexgirao/adpcm_swf

sbimikesmullin:
nvm `jpexs.jar` is already cross-platform and does it better

`java -jar "${JPEXS}" -format sound:mp3_wav -export sound "out/jpexs/${FILE}/sounds/" "out/swf/${FILE}.swf"`

https://www.free-decompiler.com/flash/features/commandline/

we could package this in?
would still need something to transcode .wav to .mp3 tho

jgranick:
This has been merged, thank you :smile:

PR #1639, opened 2017-07-16
Added more cursors to Openfl Mouse

Jarrio:
`WAIT_ARROW` does not appear any differently to `WAIT` on Windows 10, I've left it in as I'm not sure if it is platform dependent

### DIFF
--- a/openfl/ui/Mouse.hx
+++ b/openfl/ui/Mouse.hx
@@ -53,6 +53,14 @@ import openfl.Lib;
 			case MouseCursor.BUTTON: LimeMouse.cursor = POINTER;
 			case MouseCursor.HAND: LimeMouse.cursor = MOVE;
 			case MouseCursor.IBEAM: LimeMouse.cursor = TEXT;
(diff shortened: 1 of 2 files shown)


jgranick:
Hello! Thank you for your help :smile:

I have made some changes with the most recent commit, which should allow for the additional cursor types from Lime, without changing `openfl.ui.MouseCursor` :smile:

`openfl.ui.MouseCursor` now supports automatic conversion from `lime.ui.MouseCursor`, as a result, you should be able to do the following:

```haxe
Mouse.cursor = lime.ui.MouseCursor.MOVE;
```

We still use `String` as the OpenFL `MouseCursor` type, but I've added conversion to hidden private types matching the Lime cursors, so this should give the same variety as your original code, but it won't change the API signature of OpenFL `MouseCursor`

I believe `WAIT_ARROW` might be supported on macOS, but not Windows :smile:

Jarrio:
No problem :)

PR #1644, opened 2017-07-21
[SWF] change dat extension to txt because of possible forbidden MIME type

loudoweb:
I have several projects for big accounts that can't work with .dat file because of forbidden MIME type. But it works with .txt files.


jgranick:
Do you know if it works as *.bin instead? I understand that adding new MIME types to web servers in order to use OpenFL is not ideal, but I want to leave it open in the future for us to use a binary format. I wonder if this is supported widely?

loudoweb:
Hello.
It seems that I have at least one customer that forbids .dat and allows .bin. For the others, I don't know, I'm waiting for their complete list of mime type. If we use .bin, it could be great to have a fallback to .txt because the txt format is widely supported I think.

sbimikesmullin:
i have no opinion about what the file extension is, as long as it doesn't change frequently. we have a lot of assets that would have to be recompiled with this type of change. 

but i am curious about the details surrounding the experience you're reporting. what are we trying to work around--a very restrictive http proxy? i am imagining like a library system that is trying to protect its users from malicious formats. but then i wonder, what are they allowed to do with such a restrictive ruleset? and what about files without extensions? and have you considered overriding the http response header `Content-Type:`? i can't think of any browser that would hold the url's file extension as a higher precedent than that. or are we talking static targets--is it an anti-virus system that is suspicious of .dat format? 

more detail please.

loudoweb:
The issue is restricted MIME type on web server. We can't do anything against that (customer will not change anything for security reason or whatever) except trying to use different file format when problem occurs. So basically when problem occurs the server don't send the file (error 404 I think).
But I agree, maybe we should use .dat by default to not break old projects and use conditional compilation to have a different format when needed.

Issue #1402, opened 2017-01-04
Improve UTF-8 text support in `openfl.text.TextField`


jgranick:
Haxe 3.2.1 has limited UTF-8 support on some platforms, @larsiusprime has an implementation of an abstract string that should help make TextField work more appropriately, at the cost of performance. We need to integrate this support by default, or make it available behind a define, for applications using a broader character set

jgranick:
Possibly related: https://github.com/openfl/openfl/issues/1376

jgranick:
Information on a String abstract that may be a good choice for improving UTF-8 in TextField (`String` references internally would be replaced with the abstract

https://gist.github.com/larsiusprime/6cd097b9493c8173a315c68f2bc60457

pozirk:
Yes, really need that. :)
In my case for Android target, I can enter text in Cyrillic, but if I try to delete some letters, the whole string just disappears.
And after that, no matter what I type, even Latin symbols, input stays empty, while there is text in it.

pozirk:
Not fixed!
I still can not type in Cyrillic on native platforms.

RblSb:
@pozirk Could you give an example of a problem code?

pozirk:
Yes!

```
text = new TextField();
...
text.type = TextFieldType.INPUT;
```

Now launch it on Android or Windows or any native target, I suppose, and try to enter Cyrillic letters.

PR #1645, opened 2017-07-21
The text field stops to work when you write an accent or dieresis in a vocal letter (CPP)

aW4KeNiNG:
If you write "á" or "ö" then the text field doesn't work, because the `"á".length` is 2 in CPP. Then the `endIndex > __text.length` is satisfied in [TextField.hx L495](https://github.com/openfl/openfl/blob/develop/openfl/text/TextField.hx#L495). `__text` is UTF8String, so it has the good length.

### DIFF
--- a/openfl/text/TextField.hx
+++ b/openfl/text/TextField.hx
@@ -483,8 +483,8 @@ class TextField extends InteractiveObject {
 		var endIndex = __caretIndex > __selectionIndex ? __caretIndex : __selectionIndex;
 		
 		replaceText (startIndex, endIndex, value);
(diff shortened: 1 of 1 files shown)


RblSb:
Oh, thank you! I tried to fix it, but could not find the reason.
Now input should work with any languages, judging by this test:
Https://gist.github.com/RblSb/303c1986d13fb792782519bad62c7ec8
Related to #1402

aW4KeNiNG:
Did it work for you?

RblSb:
Yep

PR #1646, opened 2017-07-24
Bitmapdata drawing on HDPI displays fixed

NenadBojkovski:
Currently if you bitmapdata draw a shape or a textfield on HDPI(retina) displays only a quarter of it is drawn. This happens because the canvas scales according to the display's dpi but not scaled down when drawn. This PR fixes the scaling issue.


jgranick:
Merged, thank you :smile:

Issue #1648, opened 2017-07-25
TextField with GlowFilter doesn't respect parent sprite scaling


starry-abyss:
openfl 5.1.5, lime 5.2.1, haxe 3.4.2
targets windows, neko, html5 are affected
target flash scales as expected (but has differences in colors? :-/)

Here I have two textfields (with and without the filter) and one bitmap. On window resize the main sprite is scaled, but the textfield with the filter is out of sync with the other one. The same filter on Bitmap is updated OK.

```haxe
package;
import openfl.Lib;
import openfl.display.Bitmap;
import openfl.display.BitmapData;
import openfl.display.Sprite;
import openfl.events.Event;
import openfl.filters.BitmapFilterQuality;
import openfl.filters.GlowFilter;
import openfl.text.TextField;
import openfl.text.TextFormat;

class Main extends Sprite 
{

	public function new() 
	{
		super();
		
		var glowFilter = new GlowFilter(0x00aaff, 0.9, 4, 4, 2, BitmapFilterQuality.HIGH);

		var b1 = new Bitmap(new BitmapData(400, 400, false, 0xffff0000));

		b1.x = 200;
		b1.y = 0;

		b1.filters = [ glowFilter ];
		
		addChild(b1);
		
		var eventNameFormat = new TextFormat(null, 32, 0xff0000ff);


		var tf = createText();
		var tf2 = createText();

		tf.filters = [ glowFilter ];
		
		addChild(tf);
		addChild(tf2);

		stage.addEventListener(Event.RESIZE, onResize);
	}
	
	
	function createText(): TextField
	{
		var tfFormat = new TextFormat(null, 32, 0xff00aaff);
		
		var tf = new TextField();
		tf.setTextFormat(tfFormat);

		tf.text = "TEST TEXT";

		tf.selectable = false;

		tf.width = 200;
		tf.height = 40;

		return tf;
	}
	
	function onResize(event: Event)
	{
		var baseGameWidth = 800;
		var baseGameHeight = 480;
		
		var aspect = baseGameWidth / baseGameHeight;
		
		var xScale = Lib.application.window.width / baseGameWidth;
		var yScale = Lib.application.window.height / baseGameHeight;
		
		var scale = Math.min(xScale, yScale);
		
		this.scaleX = scale;
		this.scaleY = scale;

		this.x = Lib.application.window.width / 2 - width / 2;
		this.y = Lib.application.window.height / 2 - height / 2;
	}

}
```

[...]

PR #1650, opened 2017-07-27
Fix cacheAsBitmap issues.

NenadBojkovski:
There are several issues that we found out and fixed with this PR. Some of the issues fixed:

1. Original display object's graphic doesn't get clean when bitmap data is created, so you basically have two visual elements for a display object which can get easily out of sync.
2. The view of the original display object that remains when `__cacheBitmap` is created doesn't get updated. Namely if there is a simple animation which requires transform change and if at some point the animated display object is cached as bitmap( or gets filters), `__cacheBitmap` will keep animating while the original display object graphic will freeze.
3. Texfield cannot be cached as bitmap directly (`tf.cacheAsBitmap = true`), only as a part of cached as bitmap display object container. 
4. Caching Textfield as bitmap by caching as bitmap its parent container doesn't work always. Namely only if the caching as bitmap of the parent container and creation of the textfield and adding it to the container is in the same frame. If you cache textfield's parent display object container few frames later than textfields creation and addition to the container the textfield won't be cached as bitmap.
5. Shapes are not animating when cached as bitmap.

Some of these issues are not only DOM issues but also happen using other renderers. There's an issue recently filed which is basically fixed with this PR. https://github.com/openfl/openfl/issues/1648


jgranick:
Merged, thank you :smile:

PR #1652, opened 2017-08-02
run DocumentClass constructor before adding it to stage

kasparsj:
avoids a situation where children don't have a stage reference

### DIFF
--- a/templates/haxe/ApplicationMain.hx
+++ b/templates/haxe/ApplicationMain.hx
@@ -319,9 +319,9 @@ class DocumentClass {
 				var fields = Context.getBuildFields ();
 				
 				var method = macro {
-					
-					current.addChild (this);
+
 					super ();
+					current.addChild (this);
 					dispatchEvent (new openfl.events.Event (openfl.events.Event.ADDED_TO_STAGE, false, false));
 					
 				}


jgranick:
Do you have any sample code I could use to understand the issue you were running into? This is done in this order in order to make `stage` not null in the constructor of the document class, but perhaps there was a regression in some other behavior

kasparsj:
My DocumentClass extends a SWF asset, and all the child objects are created inside the constructor. Without this fix, the child objects don't have a stage object assigned. 

I think it's unusual to have stage already assigned inside constructor. One should register an `ADDED_TO_STAGE` listener instead, and do stage dependant stuff inside the handler.

Can post an example next week.

kasparsj:
[NyanCat.zip](https://github.com/openfl/openfl/files/1207125/NyanCat.zip)

The bug exists only if the DocumentClass extends a SWF symbol.
Attached is an example (Main2) based on NyanCat.

jgranick:
Thank you for the sample! Yes, I see the issue there.

I just made a fix in `MovieClip` that fixed the issue here, while preserving the "stage is already set" behavior of the document class constructor

I had to remove `<haxelib name="swf" />` from your sample project to build. Please give my patch a try, and let me know if it works for you, or you see other issues :smile:

kasparsj:
thanks :)

PR #1656, opened 2017-08-05
Empty TextField size

RblSb:
In Flash, the size of an empty TF is 5x5 for any font and its size
screen shot 2017-08-03 at 12 16 44 qiwhw4e

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -128,8 +128,8 @@ class TextEngine {
 		
 		this.textField = textField;
 		
-		width = 4;
-		height = 4;
+		width = 100;
+		height = 100;
 		text = "";
 		
 		bounds = new Rectangle (0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


MSGhero:
When does 100x100 come up? Isn't that the default size when there is text?

jgranick:
Maybe the size is 4x4 if you use text auto-size? @RblSb Are you using auto-size, or is this a vanilla `new TextField`?

Thanks!

RblSb:
@jgranick I forgot that I'm using TextFieldAutoSize, you're right. 4x4 will be with any of its value except NONE.
So, if you remove the text == "" check [here](https://github.com/RblSb/openfl/blob/46fd03f42ba809cffb77fe84e26c203e506cf1ba/openfl/_internal/text/TextEngine.hx#L1372), then empty textfield will calculate the size correctly by calling getLineMeasurements () i think.

MSGhero:
I guess it's the same either way. It would be more consistent to use `getLineMeasurements()` on the empty text, but you already know what the results will be for each property.

jgranick:
I just fixed the behavior of `textWidth` and `textHeight` when not type `INPUT`, I believe the current behavior is correct, in the tests I am running. Please let me know if we're missing something :smile:

PR #1661, opened 2017-08-09
Break long words, round two

### LINKED ISSUE #1185: Japanese doesn't word wrap correctly in TextFields

### LINKED ISSUE #1482: Word wrap doesn't quite work correctly on HTML5
Hello,

I've noticed a peculiar behavior in HTML5 textfields (on canvas at least) --unlike flash textfields, a wordwrap-enabled textfield will not wrap unless a space exists in the text. 

wordwrapping

The code:

```haxe
public function new() 
{
	super();
	var tf:TextField = new TextField();
	tf.type = TextFieldType.INPUT;
	tf.border = true; tf.borderColor = 0;
	tf.wordWrap = true;
	addChild(tf);
}
```

MSGhero:
Closes #1185, closes #1482
#1186 updated for OpenFL 6 and Lime 5.

Relies on #1660 for properly testing multi-byte characters, like Japanese. Probably relies on #1066 if you plan on testing with different text formats.

I've tested every case mentioned in #1185 and #1186 plus more. The only thing I've found that is inconsistent with Flash is [here](https://github.com/openfl/openfl/issues/1185#issuecomment-321135713), which isn't really important and probably isn't even documented to begin with.

cc: @larsiusprime if you have any ability of testing this or giving me test cases.

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -929,6 +929,58 @@ class TextEngine {
 			
 		}
 		
+		inline function breakLongWords (endIndex:Int):Void {
+			
+			var tempWidth = getTextWidth(text.substring(textIndex, endIndex));
+			
+			while (offsetX + tempWidth > width - 2) {
+				
+				var i = 1;
+				
+				while (textIndex + i < endIndex + 1) {
+					
+					tempWidth = getTextWidth(text.substr(textIndex, i));
+					
+					if (offsetX + tempWidth > width - 2) {
+						
+						i--;
+						break;
+						
+					}
+					
+					i++;
+					
+				}
+				
+				if (i == 0) i = 1;
+				
+				nextLayoutGroup(textIndex, textIndex + i);
+				layoutGroup.advances = getAdvances(text, textIndex, textIndex + i);
+				layoutGroup.offsetX = offsetX;
+				layoutGroup.ascent = ascent;
+				layoutGroup.descent = descent;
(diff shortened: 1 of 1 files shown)


MSGhero:
This might be for another issue, but the Japanese full stop `。` doesn't act like a period. It will get bumped onto a new line all by itself if the text is too long to fit otherwise.

It might be because of all that empty space in the right half of the character versus how small a period is. I'm not sure if periods are explicitly handled in OpenFL, but this (and other) full stop characters would have to be handled there.

PR #1670, opened 2017-08-11
Fix Raspberry Pi freeze by adding alpha > 0 check in shader

gepatto:
Somehow the Raspberry Pi freezes if we don't check for `color.a > 0.0`  in the colorTransform part of the fragmentShader. This is weird because it even happens when there is no colorTransform in use.

Setting `gl_FragColor = vec4 (0.0, 0.0, 0.0, 0.0)` in an else statement doesn't seem to be necessary.


gepatto:
Sorry, I couldn't commit from my Pi, and made changes on my Mac but in the wrong spot...
Making a new pull request ...

PR #1682, opened 2017-08-21
Previous wordwrap fix reverted due to inefficiency, new one implement…

NenadBojkovski:
…ed. Other DOM texfield improvemnets done


MSGhero:
I guess I should test this, but why did you revert the `<br>` htmlText linebreak? I didn't immediately see a place where that was handled again.

MSGhero:
This also reverted baseline alignment, and breaking long words no longer works if there are no spaces in the string (going by eye, haven't tested yet). Was it necessary to completely revert versus changing a few lines?

jgranick:
Maybe more than one change was reverted?

NenadBojkovski:
There is a new implementation of word-wrapping(breaking long words) which works faster than the old solution. Therefore I reverted the old implementation because it was interfering the new one. The bad performance for the old solution seems to be because of the extensive usage `context.measureText` method which is expensive. 
The <br> is removed because it doesn't make sense to be there. Namely, the TextEngine class is not designed to work on htmlText. The html text will be treated as a normal string so tags will be considered also in the measuring of the lines which will give wrong results. Hence, we never pass htmlText to TextEngine class.
In case of DOM Textfield, the htmlText is HtmlParser-parsed first, converting all the formatting html tags to textFormatRanges and the text converted to plain text which is later passed to TextEngine class.

MSGhero:
Ok, but you also removed baseline alignment, which has nothing to do with breaking long words. And now, text such as `"aaaaaaaaaaaaaaaaa"` or `"aaaaaaaaaaaaaaa\naaaaaa"` will not wrap properly since you only handle this within the space-wrapping while loop.

I'm fine with optimizing `breakLongWords()` because I totally just threw that together, but your current PR is breaking things that have just been fixed. It seems like your only concern is with `__context.measureTextWidth()` on DOM, so why not change that and leave everything else as-is?

As far as the `<br>` stuff, what was the point of changing `getLineBreakIndex()` in #1665 then? Or was that just an oversight? What about on other targets besides DOM js, does it still work there?

NenadBojkovski:
1. Yes the \<br> tag was an oversight. It was added as a DOM textfield improvements and was working only for DOM textfields. So neko and other targets should work without any problem.
2. Thanks for these two examples "aaaaaaaaaaaaaaaaa" or "aaaaaaaaaaaaaaa\naaaaaa". They won't work with my implementation. My test cases doesn't cover those examples therefore I didn't caught them. I'll also write a unit test for the TextEngine class so we're gonna make sure that whenever we change the class is still working.
3. Regarding `alignBaseline` method, it actually does more things then aligning. Most of its behaviour is still in the code but only scattered on different places. It has been removed also due to inefficiency. If you take a look at it there is a loop through all layout groups.  This method is called in several places but most importantly in a while loop in `breakLongWords` method. Imagine if you have a textfield with long lines and a lot of text. It is super slow. We have a credits window in our game which has a huge html text -around 700 layout groups. It takes 15 seconds to open the window with old solution and 2 seconds with this one. 

So it is not that easy just to optimise the old solution. What I could promise is that I will try to optimise it to see if I can get decent performance out of it. If I need to do major changes to it, I'd rather implement the missing things in my solution.

MSGhero:
I can add a Boolean so that it skips baseline alignment when it's not necessary. Based on what you've described, that seems like what you're really looking for, in addition to the optimized `measureTextWidth()` on DOM.

NenadBojkovski:
@MSGhero I've just committed an optimisation on the existing word-wrapping solution. Please take a look and let me know if there are any concerns.

MSGhero:
Oh yeah, that's a good way to handle it. I can test this out later, and I'll also look at making a comprehensive unit test for text.

Your tabs though lol. VS Code must have deleted them all because you have a ton of changes.

NenadBojkovski:
I don't think that it is necessary to be a visual test. TextEngine particularly is a data class, calculating and storing the text metrics given the restrictions of the text field(width, height, wordwrap, etc). How I would do the unit test for TextEngine class is by providing the text and textfield parameters to the test and check if the produced layout groups correspond to the expected ones. 
My only concern is if the measuring text with canvas context will work for the tests, if not we could use mocks I guess.

jgranick:
Not sure if you guys have seen this:

http://community.openfl.org/t/html5-textfield-not-fixed-yet/9454/6

Looks like the calls to get the layout repeatedly may be damaging the layout?

MSGhero:
Some stuff came up, I'll be testing this in a few days.

jgranick:
Do you think this would be related to: http://community.openfl.org/t/html5-assigning-large-strings-to-textfield-causes-slowdown/9501 ?

MSGhero:
Ok update:

Slight regression on every target, except DOM.
image

It seems like one extra "a" can fit on each line, but it's breaking early. I tried this on normal HTML5 and Neko. I'll look into it a bit.

MSGhero:
What is the status on this? The `ascent` change is good but the DOM changes break every other target.

NenadBojkovski:
On what change do you mean specifically ?

MSGhero:
Everything except the `ascent` changes I guess. When I tested those, only DOM gave the appropriate results.

NenadBojkovski:
That's weird because most of the code is behind `dom` compilation flag. It would be great if you could send an example project(or test) so I could use it to investigate what is actually wrong with this PR ?

MSGhero:
A long string of “a” or any letter will highlight the issue I think, it’s been a while since I’ve tested this.

Compare that to flash and to the existing class and you’ll see that this change is wrong for every wrapped line on every target but DOM. One extra character can fit on each line, but it’s wrapping early except on DOM.

NenadBojkovski:
I actually have found the issue. The `getCharOffsetAtWidth()` returns a zero-based index. Namely if the character at positions[0] is the last character in the line, the `getCharOffsetAtWidth()` will return 0, but the character at positions[0] is at offset 1 from the current one`textIndex`. Therefore we need to add 1 to get the actual character offset

PR #1689, opened 2017-09-02
Haxe4 support for open.utils.ByteArray

hklindworth:
openfl.utils.ByteArray extended haxe.io.Bytes and accessed private class field b. This is no longer possible with Haxe4 in some situations: https://github.com/HaxeFoundation/haxe/issues/6543#issuecomment-326502444

Therefore, this pull request changes openfl.utils.ByteArray to embed haxe.io.Bytes. It also avoids unwanted functions derived from haxe.io.Bytes and therefore improves compatability with Flash.

### DIFF
--- a/openfl/display3D/textures/CubeTexture.hx
+++ b/openfl/display3D/textures/CubeTexture.hx
@@ -3,7 +3,7 @@ package openfl.display3D.textures;
 
 import haxe.Timer;
 import lime.graphics.opengl.GL;
-import lime.utils.ArrayBufferView;
+import lime.utils.DataPointer;
 import lime.utils.UInt8Array;
 import openfl._internal.stage3D.GLUtils;
 import openfl._internal.stage3D.SamplerState;
@@ -107,7 +107,7 @@ import openfl.utils.ByteArray;
 		#if js
 		if (byteArrayOffset == 0) {
 			
-			uploadFromTypedArray (@:privateAccess (data:ByteArrayData).b, side);
+			uploadFromTypedArray (@:privateAccess (data:ByteArrayData).bytes.getData(), side);
 			return;
 			
 		}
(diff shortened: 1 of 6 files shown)


hklindworth:
I reverted some of the changes to increase compatibility & performance.

jgranick:
I tried the Haxe 4 preview today, and it appeared to work without modifications to `ByteArray`. Have you tried this version? Do you know if this is still necessary?

One of the motivations for the current code is to allow for `Dynamic` usage:

```haxe
var byteArray = new ByteArray ();
var dynamicReference:Dynamic = byteArray;
var bytes:Bytes = dynamicReference;
```

This is not ideal, but if you somehow end up treating the `ByteArray` as a `Dynamic`, I believe our current code will be more compatible with `Bytes` than using a different architecture?

hklindworth:
For me it still fails with Haxe4 preview and openfl develop when being in macro context:
```
/Dev/haxe/openfl/openfl/utils/ByteArray.hx:293: characters 18-25 : haxe.io.Bytes has no field b
/Dev/haxe/openfl/openfl/utils/ByteArray.hx:814: characters 3-4 : Unknown identifier : b
/Dev/haxe/lime/lime/system/ThreadPool.hx:25: characters 22-49 : Unknown<0> cannot be constructed
```
Repo: https://github.com/hklindworth/eval-issue-openfl

Regarding compatibility I was thinking more about compatibility with AdobeAir target, where it is also not possible to treat ByteArray as haxe.io.Bytes.
To improve compatibility with previous openfl versions I added already some backward compatibility methods, I could add even more to further increase compatibility.

hklindworth:
Should I adjust this pull request to the new texture classes or do you rather want to follow another approach for haxe4?

jgranick:
I will try and test Haxe 4 and your sample again, and see if I can repro

jgranick:
I was able to get Haxe 4 to compile going a different way, but I'd like to consider this again if it really seems necessary. If possible I'd like to use our current approach, though (even though it is a bit hacky) because it allows you to use `ByteArray` in place of `Bytes`, even when being treated as a `Dynamic` value :smile:

hklindworth:
Thx, alright.

I will keep that branch, just in case some day another compatibility issue arises.

PR #1691, opened 2017-09-05
TextEngine.hx fix for textfield wordwrap=true & size=0 causing an endless loop

mmauri:
This PR fixes the edge case where a Textfield has size=0 and wordwrap=true causes an endless bug.
This is a common case in HaxeUI in the process to pre-setup the layout.
Steps to reproduce:

var tf: TextField = new TextField();
tf.width = 0;
tf.wordWrap = true;
tf.text = "|";
this.addChild(tf);

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -1260,7 +1260,7 @@ class TextEngine {
 							
 						}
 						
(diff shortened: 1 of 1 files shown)


mmauri:
I close it, because there is another error, TextEngine also hangs when width is <4

PR #1692, opened 2017-09-05
TextEngine.hx fix for textfield wordwrap=true & size<4 causing an endless loop

mmauri:
This PR fixes the edge case where a Textfield has size<4 and wordwrap=true causes an endless bug in  breakLongWords inline method.
This is a common case in HaxeUI in the process to pre-setup the layout, or in our case for Textfield sizes generated from code for charts.

Steps to reproduce bug:

var tf: TextField = new TextField();
tf.width = 3;
tf.height = 200;
tf.wordWrap = true;
tf.text = "THIS IS A TEST";
this.addChild(tf);

### DIFF
--- a/openfl/_internal/text/TextEngine.hx
+++ b/openfl/_internal/text/TextEngine.hx
@@ -1260,7 +1260,7 @@ class TextEngine {
 							
 						}
 						
-						breakLongWords(endIndex);
+						if (width > 0) breakLongWords(endIndex);
 						
 						nextLayoutGroup (textIndex, endIndex);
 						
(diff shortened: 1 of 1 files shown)


MSGhero:
In Flash, can you even set the width to be less than 4? 4 should be the minimum I would think. I'm not in a position to try it out right now.

MSGhero:
Seems like you can. Good catch!

mmauri:
Thanks....
I saw yesterday openfl in a macbook with retina display, and all the textfields were blurred.
Will try to investigate more

PR #1709, opened 2017-09-12
Shape and bitmap masking

Greg209:
Shape and bitmap masking using a masking shader. Two textures are used in the shader - the object being rendered and the mask.


guruas3:
I've tried your PR in my project and it doesn't work for MovieClip.graphics mask neither for SWF mask.

PR #1728, opened 2017-10-07
ATF cubemap support

hklindworth:
Extracted the ATF reader so it can now also be used to load ATF cubemaps


jgranick:
Awesome, thank you :)

Issue #1451, opened 2017-01-25
Getting another MovieClip instance with Assets.getMovieClip() breaks previous instance (regression)


Tearff:
I use swf assets in my project (`type="swflite" preload="true"`), loading the entire timeline of every library:
```clip = Assets.getMovieClip('libraryName:');```

At some point in the last 1 or 2 public releases, the behavior of this changed: instead of being able to create multiple instances of the same clip (as expected), the previous instance now disappears from the screen and its children all become `null`.

**Comment by jgranick** [comment]
This is working for me:

```bash
openfl create NyanCat DuplicateTest
cd DuplicateTest
```

and editing "Main.hx"

```haxe
package;


import openfl.display.Sprite;
import openfl.Assets;


class Main extends Sprite {
	
	
	public function new () {
		
		super ();
		
		var cat = Assets.getMovieClip ("library:");
		cat.x = -100;
		addChild (cat);
		
		var cat2 = Assets.getMovieClip ("library:");
		cat2.x = 100;
		addChild (cat2);
		
	}
	
	
}
```

and running

```bash
openfl test neko
```

Do you know how I might replicate the problem? :smile:

Thank you!

PR #1746, opened 2017-11-01
Fix wrong movieclip being created

starburst997:
On android we hit a very weird issue where sometimes the movieclip would be empty (or have the wrong number of children), digging and digging we couldn't figure out why, it didn't makes sense. Turns out that while debugging, I notice that the wrong symbol was used during the movieclip creation but the parent definitely had the correct symbol when it tried to create the child. 

Somehow we must have one of our extension that use thread on android, a callback is called, some movieclip get created at the same time and that static variable get changed.

I think it's much simpler like that and less error prone, was there a reason why it was using static variable?

### DIFF
--- a/openfl/_internal/symbols/SpriteSymbol.hx
+++ b/openfl/_internal/symbols/SpriteSymbol.hx
@@ -31,11 +31,6 @@ class SpriteSymbol extends SWFSymbol {
 	
 	private override function __createObject (swf:SWFLite):MovieClip {
 		
-		#if !macro
-		MovieClip.__initSWF = swf;
-		MovieClip.__initSymbol = this;
-		#end
-		
 		var symbolType = null;
 		
 		if (className != null) {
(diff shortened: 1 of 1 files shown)


starburst997:
Now that I think about it if our extension is having the callback called from another thread, that's a much bigger issue and we wouldn't need that PR haha, anyway maybe that would fix this issue? #1451 ??

jgranick:
The reason for the static variable is to make the `MovieClip` children already created in the `MovieClip` constructor.

For example:

```haxe
class MyCustomMovieClipBaseClass extends MovieClip {

    public function new () {
        
        super ();
        trace (numChildren);
        trace (getChildByName ("foo"));
        
    }
    
}
```

This only works if we can execute the initialization code for the SWF symbol inside of `super()`, this static variable is the workaround to implement it :smile:

starburst997:
I see! Makes sense! Anyway the bigger issue was that we somehow was having code being called on a separate thread and we were able to figure this out because of this haha

PR #1752, opened 2017-11-08
implements DynamicDisplayObject

sbijoshj:
Adds a new extern class that has all the possible child display object members for type checking. This allows code ported from AS3 to stay very similar-looking (e.g. `myClip.child.child.play()`) and of course fail at runtime if the instances don't actually have those members, just like AS3.

### DIFF
--- a/openfl/display/DynamicDisplayObject.hx
+++ b/openfl/display/DynamicDisplayObject.hx
@@ -1,0 +1,354 @@
+package openfl.display;
+
+import lime.graphics.cairo.Cairo;
+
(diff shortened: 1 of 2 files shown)


sbijoshj:
Questions for anyone who wants to pick this up to work on: 
Can we simplify/reduce duplicate code by extending DisplayObject?
Does it also catch that we have a custom resolve function, so that `.child` calls `getChildByName ("child")`?

jgranick:
Haxe 4 removes support for `implements Dynamic` 🙁

Issue #1711, opened 2017-09-14
Ascent in HTML5 text renderers


MSGhero:
`layoutGroup.ascent` is never taken into account in `CanvasTextField` and `DOMTextField`. It's fine in `CairoTextField`.

@jgranick I see notes about hacks (`// Hack, baseline "top" is not consistent across browsers`), so that's probably out of my depth.

Test case:

```haxe
var tf = new TextField();
tf.wordWrap = true;
tf.border = true;
tf.text = "bbbbb";
tf.setTextFormat(new TextFormat(null, 20), 2, 3);
addChild(tf);
```

The bottoms of the b's should align.

jgranick:
We should have ascent hard-coded into our builds, now, on HTML5, so that we can use these values at runtime to help fix poor (read: non-existent) text measurement APIs at runtime

PR #1771, opened 2017-11-25
[WIP] Fix multiformat words

### LINKED ISSUE #1675: TextFormat breaks word wrapping
Repro on latest dev:
```haxe
var tf = new TextField();
tf.text = "here is some fantastic\ntext";
tf.setTextFormat(new TextFormat(null, 8), 15, 16);
tf.wordWrap = true;
addChild(tf);
```

image
Flash vs Neko.

The issue is that the new format causes the ["word-wrap due to spaces in the text" while loop](https://github.com/openfl/openfl/blob/develop/openfl/_internal/text/TextEngine.hx#L1117-L1361) to exit early. It should either handle the new format itself (or call a function), or there should be a way for the overall loop to re-enter the space-wordWrap condition, even if there are no more spaces in the text.

MSGhero:
Closes #1675, non-DOM parts of #1682 

~~(I have not extensively tested this one like I did the others. I broke basic text multiple times, so some normal things may currently be broken. I haven't tested justify, center, right, among other things.)~~ Good 2 go, hopefully.

Fixes the last known layout issue with text formats. Now, it word wraps correctly by placing a whole word at once instead of format-by-format. Also corrects multiformat justified text.

Testing on HTML5 will look funky due to #1711


MSGhero:
Looks like `justify` is the only one messed up. And I can't find the merge conflict in the file...?

MSGhero:
Maybe not the best named variables and functions, but hey... Ready to be tested and merged.

MSGhero:
@jgranick is anything holding this one up?

MSGhero:
Breaks on RTL text... will work on it

MSGhero:
I've done a dozen different things on this branch, so I'm going to split it up into several PRs once I finish justified RTL.

PR #1794, opened 2018-01-05
Check tile integrity before removing the tile

miltoncandelero:
Extrapolated from DisplayObjectContainer's removeChild()

---

I just encapsulated the whole thing with an if.
I don't understand why the git code looks so messy. 

Sorry :(


jgranick:
Thanks! :smile:

PR #1795, opened 2018-01-05
Only call update on dirty display objects

hklindworth:
In big scenes the scene traversal via DisplayObjectContainer.__update can have significant performance impact, even though most of the scene did not change. This change only traverses the parent objects of changed objects (to reach them) and only calls __update on the changed objects and their children.

This pull requests increases performance by around 30% in my benchmark.

### DIFF
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -123,6 +123,8 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if openf
 	private var __scrollRect:Rectangle;
 	private var __transform:Matrix;
 	private var __transformDirty:Bool;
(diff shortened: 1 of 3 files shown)


mastef:
I tested this, but it didn't produce nice results for us for some reason. When walking with a character through a 2d map, the children objects didn't update until the character stopped - so basically the whole game was stuttering. ( Imagine walking through 2D Zelda, and while you walk - the houses stay where they were. Once you stop, they then suddenly move into the new on-screen position )

hklindworth:
Thx for testing!

Can you give me some additional information about the hierarchy and the involved classes? There might be some call to __setUpdateDirty(); missing.
You can also try to call __setUpdateDirty(); and see if this helps as a workaround.

PR #1811, opened 2018-02-06
Textfield not being updated when cached as bitmap or has filters

NenadBojkovski:


jgranick:
Closed in favor of 43a6f0598cd3ba0998ce2d02441e3284999d1047

PR #1812, opened 2018-02-07
Improve docs search box focus keyboard handling

ttencate:
- Do not focus the search box initially, so we can use arrow keys and
  such to scroll the page.
- Only focus the search box if the pressed key actually looks like an
  attempt to type something (as opposed to, e.g., scroll the page)
- In line with common practice (e.g. Gmail), make / a shortcut for
  focusing the search box.
- Make Esc defocus (blur) the search box so we can use arrow keys to
  scroll the page again.
- Stop depending on nonstandard "keypress" event in favour of "keydown".


jgranick:
Awesome, thank you!

PR #1814, opened 2018-02-09
Font metrics in html5

### LINKED ISSUE #908: Font size calculations inacurate
Hi

The following scenario shows inconsistent results:
1. Create an empty test app
2. Load all the fonts via Assets.list (use a few fonts from google fonts as test)
3. Create a TextField for each font and showcase the font's tallest characters
4. Draw a rectangle around each TextField making use of the "textHeight" and "textWidth" properties (make it 2px bigger on all sides).

Source code is available at https://github.com/johanvdb/openfl-test-fonts

I realise that this is partially being discussed under #288, but I believe that this is wider than just html5, and while the biggest problems are currently being experienced in html5, I hope to see all platforms behave the same at the end of this exercise.

Here are my findings:

```
openfl test mac -Dlegacy
```

MacBook Pro with OSX 10.9.5
screen shot 2015-10-28 at 1 49 01 pm

```
openfl test mac 
```

MacBook Pro with OSX 10.9.5
screen shot 2015-10-28 at 1 56 43 pm

```
openfl test html5
```

Chrome Version 46.0.2490.80 (64-bit)
screen shot 2015-10-28 at 1 59 24 pm

```
openfl test html5
```

Safari 7.1.6
screen shot 2015-10-28 at 2 00 36 pm

```
openfl test html5 -Ddom
```

Chrome Version 46.0.2490.80 (64-bit)
screen shot 2015-10-28 at 2 05 58 pm

```
openfl test html5 -Dwebgl
```

Chrome Version 46.0.2490.80 (64-bit)
screen shot 2015-10-28 at 2 07 26 pm

```
openfl test neko
```

screen shot 2015-10-28 at 2 08 30 pm

```
johantheitguy$ haxelib list
actuate: [1.8.6]
box2d: [1.2.3]
flixel-addons: [1.1.0]
flixel-demos: [1.1.2]
flixel-tools: 1.1.0 [1.1.1]
flixel: [3.3.11] 3.3.8
hxargs: [3.0.0]
hxcpp: 3.2.102 [3.2.180]
layout: 1.2.0 [1.2.1]
lime-samples: 2.3.0 [2.6.0]
lime: 2.4.3 2.4.4 2.4.5 2.6.0 2.6.1 2.6.2 [2.6.8]
nape: [2.0.19]
openfl-bitfive: 3.0.0 3.3.2 [3.3.3]
openfl-samples: 2.2.2 3.1.0 3.3.0 [3.3.1]
openfl: 3.0.1 3.1.0 3.3.1 3.3.2 3.3.3 [3.3.8]
swf: 1.8.8 1.9.1 2.0.2 [2.1.1]
```

### LINKED ISSUE #1801: TextField rendering workaround for Firefox webgl/canvas
TextField webgl/canvas rendering is seriously broken on Firefox, due to a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=737852) that causes canvas text drawn with `baseline = "top"` to look different than in all other browsers (who's bug it is can be debated, but is irrelevant).

There's a hack that seems to be addressing exactly this issue, by moving down text by `size * 0.185`.

https://github.com/openfl/openfl/blob/4f4fa13b14c14eefffb70f7f6acae88dff4d3231/src/openfl/_internal/renderer/canvas/CanvasTextField.hx#L205-L209

However, the effectiveness of this hack is limited, the offset that needs to be applied depends on the font: some fonts need a higher, lower, or even negative offset! This often makes TextField almost unusable on Firefox, we need a better solution.

A much more reliable workaround, which however requires to know the font's `ascender`, is to draw with `baseline = "alphabetic"`, and apply an offset of `size * ascender`. I created a [demo page](https://chatziko.github.io/simulate-baselinetop.html) comparing `baseline = "top"` with this workaround (using opentype.js to get the font's ascender). The result appears identical on all browsers I tested (except firefox, of course).

I also implemented the workaround in `CanvasTextField.hx`, with the ascender hard-coded for my fonts, and it seems to work great!

To use this, of course, there's the major problem of finding the font's ascender. I see a few solutions:

1. Compute font metrics of embedded fonts during build, and make them available at runtime (they might actually be useful for other text-related math).

2. Compute the ascender at runtime. Not sure if there's an easy way though, and using opentype.js is probably an overkill).

3. Let the developper provide a map of ascender values for his fonts. Not great, but better than the current situation.

I'd go with (1), and I'm willing to implement it if I get some feedback about the best way to proceed.

chatziko:
The lack of font metrics in html5 is reponsible for at least two serious issues (#908, #1801). Unfortunately, there doesn't seem to be a reliable method to get font metrics in html5, especially if we want to support webfonts of any format added dynamically at runtime outside our control.

This PR mitigates this issue by:
 1. making font metrics automatically available for fonts known at build time
 2. allowing the developer to manually add font metrics for dynamically embedded fonts

In both cases, font metrics are used to solve both #908, #1801. If not available, the bahaviour stays the same as before. The changes also make it simple to add automated font metric computation in the future, if some reliable method is found.

Detailed summary of the changes (for both openfl and lime):

- font metric properties of `lime.text.font` are now writeable
- for fonts known at build time, font metrics are automatically made available by setting the properties in the generated `__ASSET__*` classes in `ManifestResources.hx`.

- added `openfl.text.Font.registerFontInstance`, allowing to manually add metrics for fonts not known at build time, as follows:  (also demonstrated in this [demo](http://chatziko.github.io/firefox-workaround/Export/html5/bin/))

      var font = new Font(fontName);
      font.ascender = 600;
      font.ascender = 500;
      font.unitsPerEM = 1000;
      Fonts.registerFontInstance(font);

- if font metrics are available:
  * `TextEngine` uses them to properly compute `textHeight` (fixes #908). As a bonus, this makes the corresponding methods quasi-identical for html5 and lime_cffi, so less `#if`s.

  * `CanvasTextField` uses them to correctly position text in Firefox (fixes #1801).

- if in the future we find a way to compute metrics automatically for web fonts, we can just add it to `lime.text.Font` and they'll be used automatically for the above purposes.

**Note:** the corresponding lime PR openfl/lime#1152 needs to be merged at the same time.

### DIFF
--- a/src/externs/core/openfl/openfl/text/Font.hx
+++ b/src/externs/core/openfl/openfl/text/Font.hx
@@ -76,6 +76,13 @@ extern class Font extends LimeFont {
 	public static function registerFont (font:Class<Dynamic>):Void;
 	
 	
+	/**
+	 * Registers a font instance in the global font list.
+	 *
+	 */
+	public static function registerFontInstance (instance:Font):Void;
+	
+	
 }
 
 
--- a/src/openfl/_internal/renderer/canvas/CanvasTextField.hx
+++ b/src/openfl/_internal/renderer/canvas/CanvasTextField.hx
@@ -12,6 +12,7 @@ import openfl.events.Event;
 import openfl.filters.GlowFilter;
 import openfl.geom.Matrix;
 import openfl.geom.Rectangle;
+import openfl.text.Font;
 import openfl.text.TextField;
 import openfl.text.TextFieldAutoSize;
 import openfl.text.TextFormat;
@@ -27,6 +28,7 @@ import js.html.ImageData;
 @:access(openfl._internal.text.TextEngine)
 @:access(openfl.display.Graphics)
 @:access(openfl.geom.Matrix)
+@:access(openfl.text.Font)
 @:access(openfl.text.TextField)
 
 
@@ -36,6 +38,7 @@ class CanvasTextField {
 	#if (js && html5)
 	private static var context:CanvasRenderingContext2D;
 	private static var clearRect:Null<Bool>;
+	private static var shownHackWarning:Bool;
 	#end
 	
 	
@@ -211,7 +214,25 @@ class CanvasTextField {
 						
 						if (applyHack) {
 							
-							offsetY = group.format.size * 0.185;
+							var font = TextEngine.findFontVariant (group.format);
+
+							if(font != null && font.ascender != null && font.unitsPerEM != null) {
+
+								context.textBaseline = "alphabetic";
+								offsetY = group.format.size * font.ascender / font.unitsPerEM;
+
+							} else {
+
+								// no font metrics, ballback to the old hack, but show warning
+								if(!shownHackWarning) {
+									shownHackWarning = true;
+									trace('canvas/webgl text might appear broken on this browser. Consider setting the ascender and unitsPerEM for "${font != null ? font.name : group.format.font}".');
+								}
+
+								context.textBaseline = "top";
+								offsetY = group.format.size * 0.185;
+
+							}
 							
 						}
 						
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -194,7 +194,11 @@ class TextEngine {
 	
 	private static function findFont (name:String):Font {
 		
-		#if (lime_cffi)
+		#if (js && html5)
+		
+		return Font.__fontByName[name];
+		
+		#elseif (lime_cffi)
 		
 		for (registeredFont in Font.__registeredFonts) {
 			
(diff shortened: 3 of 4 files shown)


chatziko:
Thanks for merging openfl/lime#1152 !

I just rebased this PR to the latest `develop` to resolve conflicts with 1c2fdc89ed861a1d2905943e79a576f056a74469.

Note that this PR needs to be merged to really close #908 and #1801.

jgranick:
THANK YOU, again, for all your help. This is now merged -- sorry for the very long delay! I hesitated at first because 1.) I was concerned that fonts without encoded metrics would not work (which you did resolve), 2.) I worried about `Font.registerFontInstance` (although I totally see the use-case for it) and 3.) the diff on this didn't clearly show how you preserved the existing logic for native, while allowing HTML5 to be more unified. 

I've merged it as you had it, with the exception that `Font.registerFont` accepts `Dynamic`, with support for both font classes or instances in the same method. This is really the only way you can reasonably register fonts on non-Flash platforms (where it does not need to be embedded to work with the font renderer), and I added code for Flash so it becomes a non-op there

Thanks again :smile:

PR #1825, opened 2018-02-21
Vertical scrolling for multiline input textfield implemented

NenadBojkovski:
Currently we don't have any vertical scrolling of the multiline input textfields. So if you add more lines of text than textfield can display, they won't be displayed. Normally the text should scroll as more text is added. 
Additionally vertical and horizontal scrolling with up/down arrow keys is implemented and scrolling when the characters are being removed with backspace and delete keys.


MSGhero:
Addressed in #2108, but this helped inspire some of the fixes!

PR #1828, opened 2018-02-23
Static init functions code need to be attribuable to their class

elsassph:
Once generated to JS, code from `__init__` needs to look like:
```
com_Foo.staticProp = initialValue;
com_Foo.doStaticInit();
if (Array.prototype.indexOf == undefined) {
   // some init code attributed to Array
}
```
Code analysis is broken by `__init__` code like:
```
if (typeof window != "undefined") {
   // one or more lines of static init logic
}
var foo = [1,2,3];
com_Foo.staticProp = foo;
```

### DIFF
--- a/src/externs/core/openfl/openfl/Vector.hx
+++ b/src/externs/core/openfl/openfl/Vector.hx
@@ -2114,6 +2114,12 @@ abstract Vector<T>(VectorData<T>) from VectorData<T> {
 	
 	private static function __init__ () {
 		
+		__stub__();
+		
+	}
+	
+	private static function __stub__() {
+		
 		untyped __js__ ("var prefix = (typeof openfl_VectorData !== 'undefined');
 		var ref = (prefix ? openfl_VectorData : VectorData);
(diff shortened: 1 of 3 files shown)


jgranick:
Unfortunately, this fixes one problem, to create another. Statics are defined after `__init__` on the hxgenjs project:

```js
// Init

VectorData.__stub__();

// Statics

VectorData.__stub__ = function() {
```

I'll message you about it

elsassph:
Maybe just move the `__init__` function after the stub?

jgranick:
I merged hafl of the fixes -- I have a feeling this was the one that caused the problem in the first place.

Yeah, I wondered that as well, but the modules generated do not suffer from the circular dependencies issues I had before using another approach, I am not sure if the order between static init and static methods is relevant, or not. Another option would be to consider putting static init inside a closure of some kid, such as:

```haxe
MyClass.__init__ = function () {
    // code
} ();
```

or something similar? We would have to handle scope, and again, I'm concerned about breaking some important edge case

PR #1836, opened 2018-03-07
Fixed event-bubbling of DisplayObject

MCHammel:
It may happen that 
`var result = super.__dispatchEvent (event);`
causes the parent to be removed.
By storing it temporarily we can continue bubbling, just as in flash.


jgranick:
Good catch, thanks! :smile:

PR #1842, opened 2018-03-15
reset the __worldTransformInvalid flag on __update

nadako:
this can e.g. save some milliseconds in `__hitTest` when dragging stuff, because before `__update` would update all the transforms without resetting the flag, so `__getWorldTransform` would call `__update` again


jgranick:
Good catch, thanks 😀

PR #1847, opened 2018-03-19
don't do redundant child traversing in DisplayObjectContainer/SimpleButton.__getFilterBounds

nadako:
This extracts the functionality to get actual graphics/bitmap boundaries from `__getBounds` into `__getOwnBounds` method.

It saves some CPU time when moving `cacheAsBitmap`'d sprites, because before `DisplayObjectContainer.__getFilterBounds` (called by `__updateCacheBitmap`) would call `__getBounds`, which is overriden to traverse all the children, and then traverse the children again to call `__getFilterBounds` on them. Now it just calls the `__getOwnBounds` to get rid of the first traversal.

PS I also changed `__getRenderBounds` in the same way, although I haven't found any actual usages of that function...

### DIFF
--- a/src/openfl/display/Bitmap.hx
+++ b/src/openfl/display/Bitmap.hx
@@ -83,7 +83,7 @@ class Bitmap extends DisplayObject {
 	}
 	
 	
-	private override function __getBounds (rect:Rectangle, matrix:Matrix):Void {
+	private override function __getOwnBounds (rect:Rectangle, matrix:Matrix):Void {
 		
 		if (__bitmapData != null) {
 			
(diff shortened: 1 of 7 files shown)


nadako:
Looks like this actually introduces weird performance regression in some case, don't merge yet, I'm investigating...

jgranick:
If we can assume that the bounds are not dirty (as it should trigger the dirty flag if they are changed) then perhaps we can 1.) check for `__cacheBitmap` in `DisplayObject` and use that for bounds if present and then 2.) bail in `DisplayObjectContainer` `__getBounds` if `__cacheBitmap` is present (so it doesn't recurse over children

I think this would work without a new method name

nadako:
anyway, I just figured out that this works wrongly when the object with the filter doesn't contain any graphics itself (everything is in its children), so filters extend an empty rectangle which then leads to bad things :)

PR #1849, opened 2018-03-20
DisplayObjectContainer/SimpleButton.__get*Bounds rework

nadako:
This is another take on optimizing `DisplayObjectContainer`'s `__getFilterBounds` (used by `__updateCacheBitmap`), as well as the normal `__getBounds` (which are basically the same).

Before, the `get*Bounds` function would temporarily set current object's `__worldTransform` to the given matrix and recursively update its children, and then reset everything back, recalculating the transform and traversing the children again, which is again done by the children.

Now we calculate the child world transform matrix locally and pass it to recursive `__get*Bounds` calls without modifying any state and doing excessive traversal. This saves quite some CPU time, for example, when moving complex `cacheAsBitmap`d objects.

### DIFF
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -1390,6 +1390,17 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if openf
 		
 	}
 	
+	private static inline function __calculateAbsoluteTransform (local:Matrix, parentTransform:Matrix, target:Matrix):Void {
+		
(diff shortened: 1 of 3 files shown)


jgranick:
Thank you! The render refactor has had a similar focus -- allowing our world transforms to remain correct, and to not change for logic like this. Great idea

nadako:
By the way, the `__updateChildren` doesn't seem to be used anymore, so you might want to clean it up. I didn't remove it in this PR because it originates from our own branch where `BitmapData.draw` still uses it, but I can prepare another one later if you want.

PR #1867, opened 2018-04-06
small filter check refactoring 

nadako:
Originally I was fixing the usage of `filters` getter in `__updateCacheBitmap` in our OpenFL branch, but I see it's already also fixed in d2d85a526ada0120c66f749b5fb122bd97ae7959.

Although I also did some refactoring that extracts null+length check into a `__hasFilters` inline function instead of duplicating it all over the code, which might be useful. I also use that new function in `__updateCacheBitmap`s instead of just checking for null, which is more consistent with other places. And if we don't want the length check, we can now simply change this function, so I think it makes sense.

### DIFF
--- a/src/openfl/_internal/renderer/cairo/CairoTextField.hx
+++ b/src/openfl/_internal/renderer/cairo/CairoTextField.hx
@@ -224,7 +224,7 @@ class CairoTextField {
 					
 					var usedHack = false;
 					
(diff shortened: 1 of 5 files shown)


jgranick:
Thanks! I decided to simplify it a different way -- `__filters` is null unless it's length is greater than zero, so `__filters.length > 0` is redundant. Makes sense to keep this consistent!

Thanks again

PR #1874, opened 2018-04-16
dispose texture+buffer in BitmapData.disposeImage, call disposeImage when re-rendering Graphics/TextField, delete previous buffer when recreating one for BitmapData

nadako:
This will properly delete buffer/texture when re-rending TextField and Graphics, and will also dispose the old buffer when BitmapData buffer is re-created. This is good, because I don't think we can rely on GL resources being auto-freed on GC collection.

### DIFF
--- a/src/openfl/_internal/renderer/canvas/CanvasGraphics.hx
+++ b/src/openfl/_internal/renderer/canvas/CanvasGraphics.hx
@@ -1253,6 +1253,10 @@ class CanvasGraphics {
 			
 			var width = graphics.__width;
 			var height = graphics.__height;
+
+			if (graphics.__bitmap != null) {
+				graphics.__bitmap.disposeImage ();
+			}
 			
 			if (!graphics.__visible || graphics.__commands.length == 0 || bounds == null || width < 1 || height < 1) {
 				
--- a/src/openfl/_internal/renderer/canvas/CanvasTextField.hx
+++ b/src/openfl/_internal/renderer/canvas/CanvasTextField.hx
@@ -68,6 +68,10 @@ class CanvasTextField {
 			var width = graphics.__width;
 			var height = graphics.__height;
 			
+			if (graphics.__bitmap != null) {
+				graphics.__bitmap.disposeImage ();
+			}
+			
 			if (((textEngine.text == null || textEngine.text == "") && !textEngine.background && !textEngine.border && !textEngine.__hasFocus && (textEngine.type != INPUT || !textEngine.selectable)) || ((textEngine.width <= 0 || textEngine.height <= 0) && textEngine.autoSize != TextFieldAutoSize.NONE)) {
 				
 				textField.__graphics.__canvas = null;
--- a/src/openfl/display/BitmapData.hx
+++ b/src/openfl/display/BitmapData.hx
@@ -505,6 +505,17 @@ class BitmapData implements IBitmapDrawable {
 		
 		readable = false;
 		
+		if (__texture != null) {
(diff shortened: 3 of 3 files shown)


src/openfl/_internal/renderer/canvas/CanvasGraphics.hx
```diff
@@ -1253,6 +1253,10 @@ class CanvasGraphics {
 			
 			var width = graphics.__width;
 			var height = graphics.__height;
+
+			if (graphics.__bitmap != null) {
+				graphics.__bitmap.disposeImage ();
+			}
```
jgranick: In the following lines after this call, we check if our graphics are visible. If not, we flag our bitmap for disposal.

If we will render, we determine the size. If we can re-use our previous bitmap data, we do. If not, we allocate a new one. Calling `disposeImage` here will work against these optimizations

src/openfl/_internal/renderer/canvas/CanvasTextField.hx
```diff
@@ -68,6 +68,10 @@ class CanvasTextField {
 			var width = graphics.__width;
 			var height = graphics.__height;
 			
+			if (graphics.__bitmap != null) {
+				graphics.__bitmap.disposeImage ();
+			}
+
```
jgranick: This is similar to the `CanvasGraphics` case, I believe.

src/openfl/display/BitmapData.hx
```diff
@@ -505,6 +505,17 @@ class BitmapData implements IBitmapDrawable {
 		
 		readable = false;
```
jgranick: The purpose of `bitmapData.disposeImage` (I wonder if there's a better name?) is to mark an object as intended to be hardware only. The next time it becomes eligible to render, the software surface (stored in the Lime `image` object) is disposed, and only the hardware texture remains. This means that you can no longer `getPixel`, `setPixel` or run similar operations on the `BitmapData`, but it does save some memory. Disposing the hardware surface seems to go against this purpose?

src/openfl/display/BitmapData.hx
```diff
@@ -916,6 +927,11 @@ class BitmapData implements IBitmapDrawable {
 			
 			// __bufferAlpha = alpha;
 			// __bufferColorTransform = colorTransform != null ? colorTransform.__clone () : null;
+
```
jgranick: Perhaps this might make sense, although this case is only run when `__buffer` is `null` already or if the GL context was restored, in which case your buffer object is gone anyway, so I think we could skip it

src/openfl/_internal/renderer/canvas/CanvasGraphics.hx
nadako: Not sure I understand, in the following lines we either set the `__bitmap` to `null` or recreate it with `BitmapData.fromCanvas`, so it always makes sense to release the resources of the previous bitmap?

src/openfl/_internal/renderer/canvas/CanvasTextField.hx
nadako: Yes, same comment as above ^^

src/openfl/display/BitmapData.hx
nadako: Oh, I thought it was a way to dispose the referenced texture. Then maybe the code should go elsewhere?

src/openfl/display/BitmapData.hx
nadako: Hmm, I guess you're right...

PR #1877, opened 2018-04-20
HTML5: buttonMode sets correct frame labels in MovieClip (_up, _over, _down)

rewb0rn:
I added a behavior in HTML5 to copy the behavior of Flash to show the correct frame labels when buttonMode is enabled. (_up, _over, _down)

I also tried to copy your code style and annotations, furthermore the code complexity only increases when the user sets buttonMode to true, projects that don't use the setting should not be affected at all.

### DIFF
--- a/src/externs/core/openfl/openfl/Vector.hx
+++ b/src/externs/core/openfl/openfl/Vector.hx
@@ -7,7 +7,7 @@ import openfl.utils.ByteArray;
 @:multiType(T)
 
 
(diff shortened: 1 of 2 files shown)


jgranick:
Thank you so much! https://github.com/openfl/openfl/commit/b21a7bee0ef96bec9b562c1adc131e96f04b9693

I merged your improvements, then I cleaned it up a little and got the mouse down / mouse up behavior working like your TODO. Thank you so much for the help :smile:

rewb0rn:
Ha nice! Thank you!

PR #1880, opened 2018-04-24
HTML5: Restart playing child MovieClips in subsequent cycles of a MovieClip play()

rewb0rn:
When a child of a MovieClip is added in a frame and removed in a later frame, the next time the MovieClip completes a playing cycle and comes back to the frame where the child is added, the child should start playing again (if it is a MovieClip). This PR makes sure that frame children of MovieClips are set back to the initial "playing" state when they are removed.

Previously, the HTML5 target was not behaving like Flash because the MovieClip child only played once and would not start playing again if it was stopped by a script and then created again in the next cycle.

### DIFF
--- a/src/openfl/display/MovieClip.hx
+++ b/src/openfl/display/MovieClip.hx
@@ -365,6 +365,13 @@ class MovieClip extends Sprite #if openfl_dynamic implements Dynamic<DisplayObje
 					
 					if (instance.displayObject == child) {
 						
+						//set MovieClips back to initial state
+						if (Std.is(child, MovieClip))
+						{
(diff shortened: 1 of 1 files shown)


Gama11:
You should probably use an unsafe cast instead of a safe cast here, the latter has a runtime overhead (which is a bit pointless if you're already checking the type beforehand). See also:

https://haxe.org/manual/expression-cast-safe.html

rewb0rn:
Okay thanks for pointing that out, I will fix it.

rewb0rn:
Done, let me know if it is okay like this or if there is something else.

PR #1882, opened 2018-04-26
feature/ifdef-conditional, conditional compilation with webpack

andresa88:
Allows us to add triple slash conditionals to the generated npm javascript classes that almost mirrors the functionality of the haxe conditional compilation. You need to add the ifdef-loader to a project's webpack loader list. So here's the updated lib/_gen/openfl/display/MovieClip.js init closure that includes the conditional.

```
{
  MovieClip.__parent_fps = true;
  /// #if (typeof swf_parent_fps === 'undefined' || !swf_parent_fps) && (typeof swflite_parent_fps === 'undefined' || !swflite_parent_fps)
  MovieClip.__parent_fps = false;
  /// #endif 
}
```

webpack.common.js

```
module: {
  rules: [
    {
      test: /\.js$/,
      use: [
        { loader: "ifdef-loader", options: {swflite_parent_fps: true} }
      ]
    },
    {
      test: /\.tsx?$/,
      use: [
      { loader: 'ts-loader' }
      ]
    }
   ]
}
```

A slightly modified hxgenjs ClassGenerator is required for the sole purpose of removing the trailing semicolons that the Haxe compiler adds to the end of any `__js__("...")` output. ifdef-loader errors out if it encounters these semicolons. So short of fixing this in the Haxe compiler, this can be dealt with using  hxgenjs.

```
untyped __js__("/// #endif") 

haxe outputs in the js:
/// #endif;
```

The modifications in CustomClassGenerator  are from line 189 to 204.

### DIFF
--- a/scripts/build.hxml
+++ b/scripts/build.hxml
@@ -9,6 +9,8 @@ ImportAll
 -cp ../node_modules/hxgenjs/src
 -cp ../node_modules/tink_macro/src
 -cp ../node_modules/tink_core/src
+-cp custom
+--macro CustomHxGenJSSetup.setup()
 --macro genjs.Generator.use()
 # -D hxextern
 # -D tsextern
--- a/scripts/custom/CustomClassGenerator.hx
+++ b/scripts/custom/CustomClassGenerator.hx
@@ -1,0 +1,224 @@
+package;
+
+import haxe.ds.Option;
+import haxe.macro.JSGenApi;
+import haxe.macro.Type;
+import genjs.processor.*;
+import genjs.generator.*;
+
+#if tink_macro
+using tink.MacroApi;
+#end
+using haxe.io.Path;
+using StringTools;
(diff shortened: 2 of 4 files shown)


jgranick:
The conditionals changed to use dashes instead of underscores, Haxe automatically converts our defines (using dashes) into underscores. `-Dswf-parent-fps` is the real preferred format, so figures it makes sense to support this on HTML5.

I'm using this convention for the conditional code here:

https://github.com/openfl/openfl/blob/develop/src/openfl/display/MovieClip.hx#L43

I think this leads to much cleaner source code (even if we still export both versions of the code to most targets). The `inline` should allow C++ compilers to optimize it out, and at the risk of slightly added weight to Haxelib-style HTML5 builds, I think it's probably a good way to handle this.

We can handle more defines this way, now, if we run across more we want :smile:

PR #1884, opened 2018-05-01
feature/npm es2015 modules

andresa88:
Added node script that generates the es2015 equivalent modules of the commonjs modules in the lib/_gen and lib/openfl directory. Incorporated script into the "build-lib" package.json script.

Modified package.json "module" field. Added "types" and "sideEffects", the latter allowing webpack 4 to produce smaller bundles.


andresa88:
I'll be testing this now. I found a bug with the es2015 generator script.

andresa88:
I moved all the es modules to lib_esm/ so that they're apart from the commonjs modules in lib/. 

Most of the require() calls in the lib_esm/_gen/ modules are now import calls, but I still need to figure out what to do with the pako, holwer, and file saver require() calls. None the less, webpack is able to bundle the app just fine even with those require() calls mixed in.

jgranick:
@andresa88 Thank you!

What would be the right way to test this? What are the key benefits over the CommonJS modules?

Did you look at using hxgenjs to do the ES2015 modules instead? What do you think are the benefits of the script? Does the converter script come from somewhere, or did you write it from scratch?

andresa88:
No problem! Thanks for releasing the npm version.

1. What would be the right way to test this?
I'm not sure the best way, but I've been using it and fixing errors as they pop up, so that's how I've been testing it.

2. What are the key benefits over the CommonJS modules?
I was going to say smaller bundle size, but it might not be a big difference. I'll get back to you on this.

3. Did you look at using hxgenjs to do the ES2015 modules instead? 
That would be the ideal place to do this, but I wanted to do this quickly, so I wrote the convert script from scratch. At the time it seemed that it would just be a bunch of string replacement operations, but then I ran into circular dependency issues that the commonjs implementation have. I fixed the ones I encountered but there still may be more.

4 and 5. 
I wrote it myself, but like you, I think hxgenjs is ultimately the place this should go to keep the build process consistent between commonjs and es2015.

jgranick:
So for testing, should I just use the TypeScript OpenFL samples? That will use the ES6 modules instead of the CommonJS ones, or should I change the tsconfig or anything else? What about the path mapping? Should I remove something in the default project config to test how that works? Did you need the "lime/" mapping in there, or the specific "openfl/utils/Assets" rather than "openfl/utils"?

I tested, and I didn't see much size difference between CommonJS and ES modules when packed with Webpack, but there's an obvious difference in that ES modules can be loaded at runtime from a browser, so that's an obvious advantage. I almost wonder if we need CommonJS, or if everything can understand ES modules now

I got hxgenjs running, but it took some tweaking since it has a dependency on tink_macros, which I try to avoid to keep our package lighter. It was slower than your script, but if its stable, it shouldn't matter, but that's where I wonder/worry about testing. Perhaps we could set up testing so we run over both types of modules

andresa88:
I need to fork your generator-openfl project and make the changes I've made to my tsconfig and webpack config files to get it all working with es6 modules. But I'll post below the changes I made to both.

I ran some tests myself using the latest openfl 8.1.1 library and got the following when using webpack 4.

Note: I had to make some fixes to the generator script due to the updates in 8.1.1, so I'll push those later.

1. Default project setup, so using the commonjs modules and direct imports such as: 
import Sprite from "openfl/display/Sprite"
892 KB

2. Using es6 modules and direct imports. 46 KB reduction in size of bundle.
846 KB

3. Using es6 modules and index module import like:
import { Sprite, Stage } from "openfl";
846 KB

4. Using commonjs modules and index module import. Largest bundle size.
1.2 MB

So no matter what import style you go with, you get about 46 KB reduction in the bundle size when you use es6 modules compared to the default commonjs setup. You can use the easier import syntax with es6 modules, import { .. } from "openfl", whereas if you do the same with commonjs, the bundle size increases to 1.2 MB because webpack cannot prune unused cjs modules. You need to use webpack 4 along with the sideEffects package.json flag or else 3. will output a 1.2 MB bundle too.

So the benefit of using  import { Sprite, Stage } from "openfl" from a development perspective is that auto import seems to work better in VSCode compared to when you use direct imports. If a module is not linked from the entry app.ts module, then you just need to add: 

import { } from "openfl";

at the top of the file to allow the auto import of any openfl module.

And here are the changes I made to configuration files. We can remove both "paths" from tsconfig.json and the openflPaths() alias from webpack if we don't use any direct imports like: import Sprite from "openfl/display/Sprite". So I keep those in for backwards compatibility.


tsconfig.json - Need to use "module: es2015" for webpack tree shaking to work and moduleResolution needs to be set to "node" or else it's "classic" by default when es2015 is used for module. I removed "openfl" entry in paths and kept "openfl/*" for backwards compatibility.
```
{
  "compilerOptions": {
    "target": "es5",
    "module": "es2015",
    "moduleResolution": "node",
    "baseUrl": ".",
    "paths": {
      "openfl/*": ["node_modules/openfl/lib/openfl/*"]
    },
    "lib": ["es2017", "dom"],
    "sourceMap": true
  },
  "files": ["src/app.ts"]
}

```


webpack.common.js - Added openflPaths() path mapping for backwards compatibility.

```
const path = require ('path');
const openflPaths = require('openfl/lib_esm/path-mapping');

module.exports = {
	entry: "./src/app.ts",
	output: {
		path: path.resolve (__dirname, "dist"),
		filename: "app.js"
	},
	resolve: {
		alias: {
			...openflPaths()
		},
		extensions: [
			'.ts',
			'.tsx',
			'.js'
		]
	},
	module: {
		rules: [
			{
				test: /\.tsx?$/,
				loader: 'ts-loader'
			}
		]
	}
};
```


webpack.prod.js - I added "mode: production", required by webpack 4 and removed UglifyPlugin since webpack 4 now minifies for us.

```
const webpack = require ('webpack');
const merge = require ('webpack-merge');
const common = require ('./webpack.common.js');

module.exports = merge (common, {
	devtool: "source-map",
	plugins: [
		new webpack.DefinePlugin ({
			'process.env.NODE_ENV': JSON.stringify ('production')
		})
	],
	mode: "production"
});

```


openfl/package.json

```
...
"module": "lib_esm/openfl/index.js",
"sideEffects": ["lib_esm/openfl/utils/AssetLibrary.js"],
"types": "lib/openfl/index.d.ts",
...
```


app.js
```
import { Sprite, Stage } from "openfl";

class App extends Sprite {

	constructor () {
		super ();

	}
}

var stage = new Stage (550, 400, 0xFFFFFF, App);
document.body.appendChild (stage.element);

```

jgranick:
I was struggling to find information about ES6 on browsers + Node.js, a lot of documents point to using *.mjs instead of instead of *.js, but one project I saw used *.esm.js instead. Would one of these solutions be reasonable for putting both module types in the same "lib" directory, or I wonder about only doing ES6 modules, except I'm afraid that might have negative effects for backward compatibility, but ES6 modules seems to obviously be the future

andresa88:
So I took a look at hxgenjs and it doesn't seem to output es2015 modules. It does output es6 language features like "class", but it still includes commonjs "require()" calls right? So it outputs commonjs + es6 modules.

My script outputs es2015 modules (import and export calls) but keeps the es5 prototypical style class notation that hxgenjs outputs by default. But man, it looks daunting to get es2015 module output working in hxgenjs, Haxe is not my strong suit which is why I just went with the script route.

jgranick:
Oh no, that's fine. I think the script you wrote is fine, and I'm unsure whether doing full ES6 output would really make sense outside updating the compiler to support it natively.

I opened up a pull request for an ES6 Haxe proposal here:

https://github.com/HaxeFoundation/haxe-evolution/pulls

I appreciate your input and feedback on how we can continue to make things behave more nicely for ES6 and TypeScript development :smile:

PR #1885, opened 2018-05-02
HTML5: Improve Performance for colorTransform reset

rewb0rn:
With my previous PR the colorTransform is now reset everytime the frameObject is created in the MovieClip. With this PR the colorTransform is only created again if the existing colorTransform is not already the default transform, i.e. this commit improves the performance by creating less objects.


jgranick:
Thanks! :smile:

PR #1886, opened 2018-05-03
Fix tilemap after being forced into canvas renderer

Misiur:


jgranick:
Thank you! I just implemented this slightly differently -- I'll think about putting this in a helper private method

PR #1889, opened 2018-05-07
Filters fail to complain due to non-existing properties in macro context

NenadBojkovski:


src/openfl/filters/GlowFilter.hx
```diff
@@ -126,6 +127,7 @@ import openfl.geom.Rectangle;
 		__glowShader.uColor.value[1] = ((color >> 8) & 0xFF) / 255;
 		__glowShader.uColor.value[2] = (color & 0xFF) / 255;
 		__glowShader.uColor.value[3] = alpha;
+		#end
```
Gama11: I'm not sure what problem you're trying to solve here exactly, but wouldn't the compiler still complain if it comes across this code in `!macro`? There's still a `return __glowShader;`, and it won't know what `__glowShader` is.

PR #1894, opened 2018-05-18
Enable off-stage input text

MSGhero:
On Flash, `textField.stage` can be null while the input TextField still receives events and updates its caret and such, which requires a check against `stage.focus`. The `stage != null` checks in TextField.hx prevented that on non-Flash targets.

A static reference to the stage is used here (via `Lib.current`) to avoid `this.stage` while still having access to `stage.focus`.


MSGhero:
How do you get multiple stages? I wasn't aware that was a thing

jgranick:
You will have multiple stages if you create multiple windows on the deskop, or (since HTML5 embeds share a static context) multiple HTML5 embeds on the same page.

The latter would be great if we could fix it by creating separate variable contexts somehow

MSGhero:
Is it possible to do on flash? Not sure how to handle this otherwise

MSGhero:
@jgranick looking at this again, does multiple stages really matter if all of these changes revolve around `focus`? As in, wouldn't Window 0's stage lose focus when you select Window 1, so TextField 0 could never refer to Stage 1?

MSGhero:
Talking to the void, but this might all be possible from within TF's `FocusEvent` callback. There would perhaps need to be a new `stage` property added to the event on not-Flash, but that would give an off-stage TF all the info it needs to work like in Flash.

MSGhero:
Not really interested in addressing this anymore

PR #1901, opened 2018-05-25
Fix es2015 generator and circular dependency detection

andresa88:
I updated generate-es2015.js, the build-lib-esm npm script and the lib/index.d.ts and lib-esm/index.js modules. Now when build-lib-esm is run it will also update the index modules under lib-esm/ when those in lib/ have been updated.


andresa88:
Hold off on merging this. I'm going to separate the generate-es2015 call into two separate scripts in package.json. "build-lib-esm" that outputs the modules to lib-esm/_gen/ and another for generating the ones under lib-esm/openfl/ which would need to be called manually, but rarely since those don't change often.

andresa88:
I added initial circular dependency detection option to the script. It looks for potential run-time errors that would then need to be dealt with by updating the circular dependency section of the createEsmModule() function.

PR #1902, opened 2018-05-28
fix the behavior of the visible property when overrode

dpomier:
when overriding the property `visible`, setting `super.visible` to `false` does not behave as expected; for instance the instances of a class which extends DisplayObject will still be rendered if `get_visible` is overrode to always return `true`

this PR preserve the behavior of `super.visible` to behave like in flash when `visible` is overrode


jgranick:
Thank you! :smile:

PR #1904, opened 2018-05-29
fix regression from ec50f5f

dpomier:
As `URLVariables` is an abstract (since ec50f5f6bfaf213fc2dcde183068d08b7b170b3a) we cannot use `Std.is`. It will always be false with haxe 3.4.7 and throw an exception at runtime with haxe 4.0.0-preview.3 (html5)

There is maybe a better way to fix this so feel free to change it. In the meantime this do the job


jgranick:
Great! Thank you

PR #1908, opened 2018-05-31
Don't fire MOUSE_LEAVE event in the mouse button is down - that's how…

NenadBojkovski:
Don't fire MOUSE_LEAVE event in the mouse button is down - that's how flash behaves


jgranick:
Does the event fire when the mouse button is released outside the window?

PR #1911, opened 2018-06-01
Moved ifndef block in MovieClip.hx to constructor

andresa88:
In the hxgenjs generated MovieClip.js, the static variable __useParent is initialized to false after the __init__() scripts are executed thus overriding whatever value __useParent is set to in __init__(). This seems to follow the Haxe specification in which static variable initialization comes after __init__() invocation. 

So the only solution I could think of was to move the ifndef block to the constructor.

I also swapped the dashes with underscores in the swflite_parent_fps defines. May have been a typo, a reason for it or something more sinister ;)


jgranick:
Does this do it?

https://github.com/openfl/openfl/commit/9f61c0592d9f78d4ac68ca4d62786c711b9df391

This way, we won't initialize the value, so that the `__init__` value is not overridden.

When you pass `-D swf-parent-fps` in Haxe, it converts it to `#if swf_parent_fps`, as `#if a-b-c` is not valid Haxe code. We prefer dashes over underscores, and that's how we document them. I thought it would be better to use dashes, unless we _really_ wanted to support both, but maybe it's okay to just support `swf-parent-fps`?

andresa88:
Yep, that works too! 

Oh I didn't realize that was the behavior in Haxe, I see now. So then the problem here is that variable names with dashes causes the ifndef loader to error out since it evaluates the conditionals as JavaScript. Here is the relevant snippet in the ifndef-loader that evaluates the conditionals. In our case, the "condition" param would be set to the following string:

 "(typeof swf-parent-fps === 'undefined' || !swf-parent-fps) && (typeof swflite-parent-fps === 'undefined' || !swflite-parent-fps)"

So the answer would be to replace any dashes found in the variables supplied to ifndef-loader in the webpack config with underscores. I do want to stay consistent with what's documented in terms of defines, so I'll see how we can insert the logic into the loader.

```
function evaluate(condition: string, keyword: string, defs: object): boolean {

   const code = `return (${condition}) ? true : false;`;
   const args = Object.keys(defs);

   let result: boolean;
   try {
     const f = new Function(...args, code);
     result = f(...args.map((k) => defs[k]));
      //console.log(`evaluation of (${condition}) === ${result}`);
   }
   catch(error) {
      throw `error evaluation #if condition(${condition}): ${error}`;
   }

   if(keyword === "ifndef") {
      result = !result;
   }

   return result;
}
```

andresa88:
Another solution that doesn't require a change to ifndef-loader. We can instead use this:

```
MovieClip.__useParentFPS = true;
/// #if (typeof defines == 'undefined' || (!defines["swf-parent-fps"] && !defines["swflite-parent-fps"]))
MovieClip.__useParentFPS = false;
/// #endif 
```

And our ifndef configuration:

```
{
  test: /\.js$/,
  use: [
	{ 
          loader: "ifdef-loader", 
          options: {
            defines: { 
              "swflite-parent-fps": true 
            } 
          }
        }
  ]
}

```

jgranick:
Hmm, this one is tricky.

In the documentation, it uses "ifdef-verbose" or "ifdef-triple-slash" as defines to control it's own behavior. Sad that this style of defines doesn't propagate properly into the source, I see what you mean now, because it does not result in a valid JavaScript expression.

I wonder if there's a good solution for this in `ifdef-loader` (or another library?). Perhaps it should replace dashes with underscores, similar to the Haxe behavior, so `swf-parent-fps` becomes `swf_parent_fps` and we `/// #if swf_parent_fps` similar to how you would do it in Haxe.

Is your custom `evaluate` function converting a "swf-parent-fps" in the webpack config to underscores, or is it modifying the evaluation behavior within the source files?

jgranick:
What are you feelings on the DefinePlugin:

https://github.com/webpack/docs/wiki/list-of-plugins#defineplugin

https://stackoverflow.com/a/29851256/1193912

andresa88:
That could work too. The plugin is included with webpack so that's a plus. To your point, we would just have to add an undefined check in case ENV is not defined.

if (typeof ENV != 'undefined' && ENV.debug) {
    console.log('Yo!');
}

So in our case we could use this:

MovieClip.__useParentFPS = true;
if (typeof ENV == 'undefined' || (!ENV['swflite-parent-fps'] && !ENV['swf-parent-fps'])) {
  MovieClip.__useParentFPS = false;
}


So ifdef-loader would be useful in cases where we want to conditionally strip out dead code from the final bundle. So like if we have the following esm style imports

```
import { Game1 } from "./game1";
import { Game2 } from "./game2";

/// #if (typeof game != 'undefined' && game == 'game2')
Game2.start();
/// #else
Game1.start();
/// #endif
```

If "game" is set to game2, the bundle will not include the game1 module nor any of its dependencies, only game2. Whereas if we used the DefinePlugin, both modules would be included in the bundle, at least according to a comment in that SO link.

BUT we could also use the dynamic import() expression to deal with this situation and strip out both modules and their dependencies from the bundle until they're loaded with import(). So we have that option as well.

jgranick:
I tried to merge both approaches:

https://github.com/openfl/openfl/commit/bede8e2913fc32519c220c7ec1f270c3390fc232

Let me know if you have any issues :smile:

PR #1913, opened 2018-06-05
Don't try to "sync" BitmapData's Image before uploading it as a texture.

nadako:
This would cause it to be drawn to a freshly-created Canvas which is heavy (esp. on MS Edge) and unnecessary because texImage2D supports uploading directly from Image elements.

PS also remove the unused `__sync` method.

### DIFF
--- a/src/openfl/display/BitmapData.hx
+++ b/src/openfl/display/BitmapData.hx
@@ -1051,10 +1051,6 @@ class BitmapData implements IBitmapDrawable {
 			
 		}
 		
-		#if (js && html5)
-		ImageCanvasUtil.sync (image, false);
-		#end
-		
 		if (image != null && image.version > __textureVersion) {
 			
 			var internalFormat, format;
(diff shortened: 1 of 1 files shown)


nadako:
This is the second part of the change we had to make to avoid creating canvases for all the `BitmapData`s loaded over the network and upload textures directly from HTML image elements. The first part is here: https://github.com/openfl/lime/pull/1199

jgranick:
There is a possibility that an `Image` could need to be synchronized between an internal canvas and data buffer. For example, the type might be `DATA`, but the user might call `copyPixels`, which is MUCH faster as `context.drawImage` instead of modifying `UInt8Array` values, but when they go to upload, we need to commit these changes back to image data so it's not out-of-date.

Perhaps we need an `image.type == IMAGE` case in the `sync` method:

https://github.com/openfl/lime/blob/develop/lime/graphics/utils/ImageCanvasUtil.hx#L432

That could avoid performing a conversion for objects without any pending changes, while still allowing for the existing behavior?

jgranick:
I added an update to the `sync` method in Lime, here:

https://github.com/openfl/lime/commit/f01e980114e9b900330825fec8a8dd022b464807#diff-9e56bd57d787d3db5ec513d8de92facb

I think this may help prevent the case where it's an HTML5 Image, with no canvas or data, but still sync otherwise

PR #1914, opened 2018-06-05
dispatch MOUSE_MOVE events once per frame or when other input event (click/keypress) happens

nadako:
As described/discussed in https://github.com/innogames/openfl/pull/26, this helped a lot in our game by saving a lot of CPU time not processing redundant mouse move events between frames.

So far we didn't notice any issues with this, so we'd like to propose this for the official OpenFL.


MSGhero:
When I was testing with flash at a very low FPS, MOVE events were being generated at like 60 FPS or something. I put a trace in enterframe and another as a MOVE callback. The MOVE one was firing multiple times in between enterframe ones. 

I don’t see it as a big deal, but just so you know.

nadako:
So, what is gonna happen with this PR? Still works great for us!

jgranick:
I added `-Dopenfl-always-dispatch-mouse-events` as a workaround if someone wants to disable this behavior, meanwhile, I think it's worth building off of these changes, since I know hit testing may be expensive. Thanks :smile:

PR #1915, opened 2018-06-07
Perfomance improvement for GLTilemap

Beeblerox:
This is the first step to make Tilemap rendering faster.
In current OpenFL version there is happening constant buffer uploading at every call to `flush()` method.
I've added simple check and now buffer uploads only at shader switch cases.

The next step (in next PR that i'll try to send to you) is to reupload only necessary part of buffer to gpu via `gl.bufferSubData()` instead of  `gl.bufferData()` calls


Beeblerox:
yay!

PR #1916, opened 2018-06-08
Add individual tile blending support for GLTilemap

Beeblerox:


Beeblerox:
@jgranick ok, i think i've cover all cases for blend modes for tiles on OpenGL renderer and added support for Cairo, Canvas and Flash as well. Could you review it, please?

PR #1928, opened 2018-06-28
Removed setting MovieClip child objects' parent and stage properties on creation

aaronneugebauer:
Setting parent and stage properties immediately (rather than being set in the DisplayObjectContainer.addChildAt method) prevented the Event.ADDED_TO_STAGE event from being fired.

### DIFF
--- a/src/openfl/display/MovieClip.hx
+++ b/src/openfl/display/MovieClip.hx
@@ -611,8 +611,6 @@ class MovieClip extends Sprite #if (openfl_dynamic && haxe_ver < "4.0.0") implem
 							
 							if (displayObject != null) {
 								
-								displayObject.parent = this;
-								displayObject.stage = stage;
 								instance = new FrameSymbolInstance (frame, frameObject.id, frameObject.symbol, frameObject.depth, displayObject, frameObject.clipDepth);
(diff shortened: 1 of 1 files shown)


jgranick:
Thank you @aaronneugebauer for your help, I am sorry for the delay in being able to dig further into this.

We do this currently so that the SWF-based children exist in custom SWF extension classes.

For example:

```haxe
class MyClass extends MySWFSymbol {
    
    public function new () {
        super ();
        trace (this.numChildren); // should already include the SWF children
    }
    
}
```

If we do not, the children are added after the constructor.

That said, we do dispatch an extra `ADDED_TO_STAGE` event in some spots to reproduce the Flash behavior for the event.

Here's an example:

https://github.com/openfl/openfl/blob/66cfc08a452361fffe4833267d82448dbe64db21/src/openfl/display/Stage.hx#L822

Perhaps we need to dispatch this somewhere?

jgranick:
I'm committing this, let me know if this has issues, or helps your use-case?

https://github.com/openfl/openfl/commit/aa16aa8c1549fe1dbe7e2ed3453bbb34814663ac

Thanks :smile:

PR #1930, opened 2018-06-29
improving TextField cacheAsBitmap behavior

rewb0rn:
During my tests I found that my initial change correctly fixed the positioning of the TextField, however the size of the Bitmap was too small in some cases. It seems that the __getBounds logic was changed and does not work correctly anymore.

By adjusting the matrix instead of the bounds rectangle in __getBounds(), the size of the Bitmap is corrected as well as the positioning of the matrix in regards of the TextField offset values. Therefore my previous change has to be undone. The positioning is still fixed, but now also the size of the rectangle seems to be correct.


PXshadow:
Your a legend thank you so much! This fixes so much for me!

PR #1934, opened 2018-07-04
Bitmap pixelsnapping implemented

NenadBojkovski:


nadako:
@NenadBojkovski should we close this for now as our fork now has a more complete implementation of that?

NenadBojkovski:
Yes we can do that and file new PR with the complete solution.

PR #1943, opened 2018-07-09
fix canvas colorTransform

dpomier:
When compiling with -Dcanvas, the children are not well displayed when there is a colorTransform applied. This PR try to fix it.

I tested it with the following project (c.f. [TFColorEffect.zip](https://github.com/openfl/openfl/files/2176294/TFColorEffect.zip)):

```haxe
class Main extends Sprite {

	public function new () {
		
		super();
		
		var _mc = Assets.getMovieClip("mylib:Button");
		
		addChild(_mc);
		
		_mc.alpha = 0;
		
		haxe.Timer.delay(function () { _mc.alpha = .4; }, 200);
		haxe.Timer.delay(function () { _mc.alpha = 1; }, 500);
		
	}
}
```
`mylib:Button` is a MovieClip containing a MovieClip with a brightness effect containing a TextField. At run-time the TextField is visible in webgl but not in canvas.

I didn't test it in dom but it would be interesting to try.

Maybe it is not the right way to fix this as I don't have a deep understanding of what is happening here. In the meantime it does fix some of our buttons in game.

### DIFF
--- a/src/openfl/display/DisplayObjectContainer.hx
+++ b/src/openfl/display/DisplayObjectContainer.hx
@@ -11,6 +11,7 @@ import openfl.errors.RangeError;
 import openfl.errors.TypeError;
 import openfl.events.Event;
 import openfl.events.EventPhase;
+import openfl.geom.ColorTransform;
 import openfl.geom.Matrix;
 import openfl.geom.Point;
 import openfl.geom.Rectangle;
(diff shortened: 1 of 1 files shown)


jgranick:
Okay, this test appears to be working properly with the current development version of OpenFL.

I have tried to improve how caching works with `colorTransform`

dpomier:
Effectively, nice!

PR #1952, opened 2018-07-13
properly handle half-pixel coordinates for rectangle masks

nadako:
sometimes when rendering objects with a scrollrect (or tilemaps) at half-pixel coordinates, the width/height for `gl.scissor` wasn't correctly expanded and the object was missing one-pixel line from top or right, because the `__scissorRect` function didn't consider the position for calculating width/height. this PR fixes that

PS: another commit fixes the annoying whitespace trailing in vscode

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     "editor.trimAutoWhitespace": false,
+    "files.trimTrailingWhitespace": false,
     "editor.insertSpaces": false,
     "codedox": {
         "commentprefix": "    ",
--- a/src/openfl/display/OpenGLRenderer.hx
+++ b/src/openfl/display/OpenGLRenderer.hx
@@ -927,14 +927,14 @@ class OpenGLRenderer extends DisplayObjectRenderer {
 			__gl.enable (__gl.SCISSOR_TEST);
 			
 			var x = Math.floor (clipRect.x);
-			var y = __flipped ? Math.floor (__height - clipRect.y - clipRect.height) : Math.floor (clipRect.y);
-			var width = Math.ceil (clipRect.width);
-			var height = Math.ceil (clipRect.height);
+			var y = Math.floor (clipRect.y);
+			var width = Math.ceil (clipRect.right) - x;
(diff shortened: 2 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "editor.trimAutoWhitespace": false,
+    "files.trimTrailingWhitespace": false,
```
nadako: BTW, would you consider trimming trailing whitespace in openfl/lime code? I find it very annoying to maintain the current aligned tabs.

jgranick: Could you give an example of trailing whitespace? Do you mean between lines, or on the same line?

jgranick: Oh, so the editor is removing whitespace when saving a file, as compared to while editing?

jgranick:
The `__flipped` property is used when using a scissor rectangle on the main framebuffer, versus render-to-texture. Is this no longer needed, or should it be adapted to be included? Thanks :smile:

nadako:
> The __flipped property is used when using a scissor rectangle on the main framebuffer, versus render-to-texture. Is this no longer needed, or should it be adapted to be included? Thanks 😄

It's still handled, but now directly in the `gl.scissor` call. So we have it a bit separated: first we expand the rectangle, then pass it (possibly flipped) to the GL call.

.vscode/settings.json
nadako: Trailing whitespace is a whitespace (tab or space) to the right of a non-whitespace character, e.g.

```
someCall()<space><space>
```

in OpenFL/Lime codebases specifically there are a lot of whitespaces on the empty lines, like
```
<space>function f() {
<space><space>
<space><space>doSomeCall();<space>
<space><space>
<space>}
```

trimming trailing whitespace means that the text above will be saved as:
```
<space>function f() {

<space><space>doSomeCall();

<space>}
```

nadako: (the empty line themselves are preserved of course, but they will be empty instead of having a number of whitespace characters)

PR #1953, opened 2018-07-13
don't call __update in __updateCacheBitmap

nadako:
> adapted from https://github.com/innogames/openfl/pull/10

since `__updateCacheBitmap` is always called from rendering functions, which are always called after the update, we don't need to traverse again

also, fix returning `rect` to the pool and remove now unused `matrix` var


jgranick:
What about using `bitmapData.draw`?

nadako:
> What about using bitmapData.draw?

It calls `__update` itself:

https://github.com/openfl/openfl/blob/7bb0a04cd62b8c11fc2413a3384250ede523ef8c/src/openfl/display/BitmapData.hx#L530-L534

nadako:
are there any issues with this?

PR #1964, opened 2018-07-20
orphan display objects cleanup fixes

nadako:
> from https://github.com/innogames/openfl/pull/57

- cleanup and dispose cache bitmap
- cleanup child display objects recursively
- always cleanup removed children even if we return early and don't actually do the rendering

---

this will properly (and timely) dispose cache bitmaps and their textures in the GPU. this greatly minimizes the impact of small-ish memory leaks, where the object is removed from the stage, but the reference to it is still hanging somewhere. we just had a nasty case with a minor event listener we forgot to remove retained a display object with huge cacheAsBitmap textures that were never cleaned up and took a lot of GPU space

### DIFF
--- a/src/openfl/display/DisplayObjectContainer.hx
+++ b/src/openfl/display/DisplayObjectContainer.hx
@@ -995,6 +995,8 @@ class DisplayObjectContainer extends InteractiveObject {
 	@:noCompletion private override function __renderCairoMask (renderer:CairoRenderer):Void {
 		
 		#if lime_cairo
+		__cleanupRemovedChildren ();
+		
 		if (__graphics != null) {
 			
 			CairoGraphics.renderMask (__graphics, renderer);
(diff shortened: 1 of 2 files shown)


src/openfl/display/DisplayObjectContainer.hx
```diff
@@ -676,6 +676,38 @@ class DisplayObjectContainer extends InteractiveObject {
...
+			
+			if (orphan.stage == null) {
+				
+				orphan.__cleanup ();
+				
+			}
+			
+		}
+		
+		__removedChildren.length = 0;
```
nadako: As also mentioned in https://github.com/innogames/openfl/pull/57#discussion_r203802882:

I have a gut feeling that using simple array for `__removedChildren` and just setting it to an empty one here would perform better than this whole `Vector` machinery, but this is a bit off-topic for this PR.

PR #1966, opened 2018-07-23
[FIX] Typescript syntax errors

Linard11:
There were some syntax errors in the typescript types that probably happend during the automatic conversion of the .hx files.

This would be a temporarily fix that worked for us to get typescript to transpile without errors but its best to find out what exactly went wrong during conversion (if it was done automatically and not by hand)


jgranick:
Thank you!

PR #1996, opened 2018-08-23
Restore binding to non-default framebuffer

justin-espedal:
Allow a framebuffer other than the default to be restored to after temporarily binding other framebuffers.

This approach explicitly enforces a desired buffer to be restored to.


justin-espedal:
Closing this because `Context3D`'s `setRenderToTexture` and `setRenderToBackBuffer` now provide an adequate solution.

Issue #1984, opened 2018-08-14
TextField.htmlText splits words with html-formatting


YvesScherdin:
If a word at the textfield's border should be taken to the next line completely, but has some html-text-formatting applied, the word gets split.
The textfield adds all still-fitting letters of that word into that line and puts the rest into next line.

Example-Code:

```
myTextField.wordWrap = true;
myTextField.htmlText = "Test Teeeeest Teeeeeeeeeeestt <font color='#ffffff'>Teeeeeeeeeeeeeeeeeeeeeest</font>";
```

Result: (left side "with", right side "without" html-formatting)

![htmltextformattedwordlinebreak_short](https://user-images.githubusercontent.com/6424614/44104202-8b952c30-9fee-11e8-93bf-1ede7f489fb4.jpg)

MSGhero:
Does it happen if you bold the text with a textformat instead?

YvesScherdin:
yes. bold has no effect on the bug.

YvesScherdin:
Here, I created a minimalistic app for testing:

[Test_TextFieldLineBreaks.zip](https://github.com/openfl/openfl/files/2289586/Test_TextFieldLineBreaks.zip)

YvesScherdin:
And I also enhanced the app for testing... it provides live textfield content editing with a few test cases I encountered so far, which should make the testing easier, if that helps.

[Test_TextFieldLineBreaks.zip](https://github.com/openfl/openfl/files/2314604/Test_TextFieldLineBreaks.zip)

MSGhero:
Oh I see. I might have fixed this in a PR I did already. It got merged and reverted, so I’ll have to re-PR it. It “broke” `defaultTextFormat`, which doesn’t behave correctly anyway, so I probably need to fix that first before my PR gets accepted.

MSGhero:
I checked and #2003 fixes this, as far as I played with your example.

YvesScherdin:
okay, cool. I tried this out with the development branch with my example. There it works for me too. Yet, I was unable to compile it with my bigger project. But I assume that just that branch is maybe not stable currently.

MSGhero:
Is this one closed now?

YvesScherdin:
Yes, thanks!

PR #2003, opened 2018-09-08
Multi-format word fixes, round 3

### LINKED ISSUE #1675: TextFormat breaks word wrapping

MSGhero:
Closes #1675, #1984 I think. Redo of #1771 and #1942. Multiple formats within the same word behave now, I hope.

The diff is kind of a mess, but this PR ignores `formatRange.end` when word wrapping. All text between `textIndex` and the next space/break/end of text is placed at once, and all format changes, word wrapping, and breaking of long words are handled within each of those placements.

### DIFF
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -893,7 +893,7 @@ class TextEngine {
 					
 					var advance;
 					
-					if (i < text.length-1) {
+					if (i < text.length - 1) {
 						
 						// Advance can be less for certain letter combinations, e.g. 'Yo' vs. 'Do'
 						var nextWidth = __context.measureText (text.charAt (i + 1)).width;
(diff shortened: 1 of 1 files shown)


MSGhero:
Just for info, `defaultTextFormat` fixes, `left/rightMargin` implementation, and RTL need this first.

MSGhero:
Ahh I found an existing bug I might as well fix: `breakIndex` should have a `+1` in the same way that `spaceIndex` does throughout the space handling chunk. I'm getting empty layout groups instead of the `\n` being included in the previous group.

MSGhero:
Alright, I think this is to a point where it doesn't introduce any new issues.

PR #2032, opened 2018-10-15
Add VideoTextureEvent

peteshand:


peteshand:
Also add private __seeking property to netStream for html5 target. This allows video textures to update while seeking a paused video

src/openfl/net/NetStream.hx
```diff
@@ -404,6 +406,17 @@ class NetStream extends EventDispatcher {
...
+			__seekTimer.stop();
+			__seekTimer = null;
+		}
+		__seekTimer = Timer.delay(() -> {
+			if (__seeking == true) {
+				__seeking = false;
+				__connection.dispatchEvent (new NetStatusEvent (NetStatusEvent.NET_STATUS, false, false, { code : "NetStream.Seek.Complete" } ));
+			}
+		}, 50);
+
```
jgranick: What is the reason for adding a timer here? I suppose that seeking does not occur immediately, but I am concerned about adding a timer without connection to the state of the video. What if you request to seek, but you cannot (because it exceeds the video length) or the internet is slow and it is still buffering, or what if the video is stopped or seeked a second time before the timer fires?

peteshand:
I've removed the timer and updated VideoTexture.__getTexture to switch __seeking back to false

jgranick:
That seems okay for now, we'll keep an eye on this, and find ways to improve with time. This runs the risk of maybe missing an update during a seek, but that seems safer than possibly dispatching a seek event after playback has occurred, and introducing a tricky to find logic bug later. Thank you so much for your help :smile:

PR #2043, opened 2018-10-22
Fix transform updates for canvas when cacheAsBitmap = true

dpomier:
This PR fixes the transform of a display object not being updated when the latter is cached as bitmap. You can easily reproduce it like so:

```haxe
var sprite = new Sprite();
sprite.graphics.beginFill(0x333366);
sprite.graphics.drawRect(0, 0, 80, 80);
sprite.graphics.endFill();
addChild(sprite);

sprite.cacheAsBitmap = true;
		
stage.addEventListener(MouseEvent.MOUSE_MOVE, function (e:MouseEvent) {
	
	sprite.x = e.stageX; // Doesn't update the displayed position with -Dcanvas
	sprite.y = e.stageY; 
	
});
```

the square won't follow the cursor when building with `html5 -Dcanvas`.

I tested `windows`, `neko`, `dom`, and `webgl` while I was at it, they all look fine.


jgranick:
Thank you :smile:

PR #2045, opened 2018-10-23
fix context3D.present() double call bug

peteshand:


jgranick:
There should be code here to mark as cleared:

https://github.com/openfl/openfl/blob/a09f3579e074ac6a61322457c280ec3d3b3a6bf5/src/openfl/display3D/Context3D.hx#L236-L240

The idea is that we need to clear the color for it to count

Perhaps we need to ignore `present` if it hasn't been rendered yet? Is that how Flash works? (calling `present` fives times still looks the same?)

PR #2046, opened 2018-10-23
Implementation for Left/Right margins in TextFormat

lmodlinski:
Main idea is to replace all the hardcoded offsets of 2 (and their sums for both sides of 4) to actual margins or margins sum. I've updated everything I believed is required to update. I would appreciate your feedback on this :)

Example I've tested on:
```haxe
package;

import openfl.text.TextFormatAlign;
import openfl.text.TextField;
import openfl.display.Sprite;

class Main extends Sprite {
    public function new() {
        super();

        var format:TextFormat = new TextFormat('Arial', 40, 0xFF0000);
        format.leftMargin = 10;
        format.rightMargin = 10;
        format.align = TextFormatAlign.JUSTIFY;
        format.underline = true;

        var field:TextField = new TextField();
        field.defaultTextFormat = format;
        field.backgroundColor = 0x00FF00;
        field.background = true;
        field.multiline = true;
        field.wordWrap = true;
        field.width = 450;
        field.height = 100;
        field.text = 'A b c d e f g h i j k l m n o p r s t u w x y z a b c d e f';

        this.addChild(field);
    }
}
```

It's HTML5 on the left | Flash on the right, on every attached screenshot. Underscore was just left for decoration and it sometimes hints where is the issue just looking at its length :)

margins 10 10 width 450 aligned left
margins 10 10 width 450 aligned left

margins 250 250 width 450 aligned center
margins 250 250 width 450 aligned center

margins 10 10 width 450 aligned justify
margins 250 250 width 450 aligned justify

margins 250 250 width 450 aligned left
margins 250 250 width 450 aligned left

margins 100 25 width 450 aligned center
margins 100 25 width 450 aligned center

### DIFF
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -644,7 +644,28 @@ class TextEngine {
 		bottomScrollV = 0;
 		maxScrollH = 0;
 		
+		var offsetX = 2.0;
+		var offsetY = 2.0;
+		var offsetWidth = 2.0;
+		var margins = 4.0;
+		
+		var groupFormat;
+		
 		for (group in layoutGroups) {
+			groupFormat = group.format;
+			
+			offsetX = 2.0;
+            		offsetWidth = 2.0;
+
+            		if(groupFormat.leftMargin != null) {
+                		offsetX = groupFormat.leftMargin;
+            		}
+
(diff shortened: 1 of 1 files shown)


MSGhero:
@lmodlinski long delay :) but I will get refreshed on the various text PRs and get this merged in one way or another

PR #2063, opened 2018-11-08
[WIP] RTL support

MSGhero:
RTL support. Checks `__textLayout.direction`, which is currently inconvenient to set by the user (requires Reflection iirc).

An old branch that I brought up to current dev. Super WIP, I haven't even tried compiling.

I'm sure there are plenty of edge cases, but I'd rather get a mostly working implementation in than waiting for every last one to get fixed.

### DIFF
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -920,6 +920,13 @@ class TextEngine {
 				
 			}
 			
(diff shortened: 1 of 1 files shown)


MSGhero:
I'll update this once the API settles down a bit.

MSGhero:
API hasn't settled down yet 😄 but I have a plan

PR #2065, opened 2018-11-09
Feature/stage_fullscreenSourceRect

m0rkeulv:
This pull request adds support for [fullScreenSourceRect](https://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/display/Stage.html#fullScreenSourceRect). I have tested it on Html5, Neko, Windows and Android and as far as i can tell it all behaves the same way Flash and Air targets. 

i did have to comment out a few lines in OpenGLRenderer as the renderer would clear the entire screen.
can anyone confirm if the scissor stuff  done when offset X or  Y is more than 0  actually works ?


jgranick:
Thanks! I’ll take a look. Have you tried -Dcanvas or -Ddom on HTML5, or -Dcairo on native? It’s okay if they’re still TODO :smile:

m0rkeulv:
html5 (working)
html5 -Ddom  (working)
html5 -Dcanvas  (Not working)

neko (working)
neko -Dcairo (not working)

Seems it need some more work :)

m0rkeulv:
html5 (working)
html5 -Ddom (working)
html5 -Dcanvas working)

neko (working)
neko -Dcairo (working)

I did however notice some artifacts for both  the Cairo and Canvas renderer.
Switching between fullscreen with sourceRect and normal would cause DisplayObjects with filters to get some strange artifacts, it almost looks like the  filter effects are smudged out or looped. it looks like some texture coordinates are messed up or something. but i thought these where software renderers?

m0rkeulv:
Hm...
so i went back to the develop branch and rebuilt my project, and it seems these artifact already occurred for these renderers  when switching to fullscreen and/or back from it. I guess the fullscreenSourceRect part is working as intended.

m0rkeulv:
@jgranick 
As far as i know everything should be working now.
i have tested:
html5 (working)
html5 -Ddom (working)
html5 -Dcanvas working)

neko (working)
neko -Dcairo (working)

Issue #113, opened 2013-09-06
flash.events.KeyboardEvent.hx crash


hoseyjoe:
Flash output breaks openfl 1,0,6

VerifyError: Error #1058: Illegal operand type: uint must be int.

A fix would be changing line 51 from
var keyCode : Int;
to
var keyCode : UInt;

PR #2092, opened 2018-12-06
Fix incorrect KeyboardEvent types in Flash externs

Gama11:
OpenFL seems to use `Int` instead of `UInt` in its Flash externs (`KeyboardEvent` in this case, but probably others too). This happened to not cause any issues so far, but that changed with Haxe 4 preview 5, where the following causes a crash on the Flash target:

```haxe
package;

import openfl.events.KeyboardEvent;
import openfl.display.Sprite;

class Main extends Sprite {
	public function new() {
		super();
		stage.addEventListener(KeyboardEvent.KEY_DOWN, function keydown(e:KeyboardEvent) {
			switch (e.keyCode) {
				case 39:
					trace(e.keyCode);
				case 40:
					trace(e.keyCode);
			}
		});
	}
}
```

If you compare the `-D dump=pretty` output between preview 4 and current builds, you can see that the switch-case is no longer generated with a temporary variable (probably due to an improvement in Haxe's static analyser / var fusion):

If the extern uses the proper types (like it does in the extern from the standard library), this isn't an issue.

### DIFF
--- a/externs/flash/flash/events/KeyboardEvent.hx
+++ b/externs/flash/flash/events/KeyboardEvent.hx
@@ -10,15 +10,15 @@ extern class KeyboardEvent extends Event {
 	public static var KEY_UP (default, never):String;
 	
 	public var altKey:Bool;
-	public var charCode:Int;
+	public var charCode:UInt;
 	public var ctrlKey:Bool;
 	public var commandKey:Bool;
 	public var controlKey:Bool;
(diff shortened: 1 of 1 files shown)


Gama11:
Note: it seems this already came up once 7 major OpenFL versions ago, but was never fixed? Or the fix was reverted since? https://github.com/openfl/openfl/issues/113

PR #2093, opened 2018-12-07
We need this to keep our override_blend_mode data

JoeCreates:
I think we still need this back, otherwise blend modes stop working in flixel again and overrideBlendMode has no effect. Note that the default case on the switch calls data.destroy().


jgranick:
Yep, looks like you're right. Thank you :smile:

PR #2099, opened 2018-12-13
Replace internal representation of Graphics api enums to String

mrchnk:
There was a bug in npm distribution of library:

- swf converted to swflite with `openfl` tool using integer values for that types
- js implementation of swflite library expects that values to be string

That leads to exception in CanvasGraphics.createGradientPattern() method and not displaying any swflite files contains gradient fill at all.

So I guess making that values type platform-independent should be clearer and produce less type mismatch.

Unfortunately that change will break file format compatibility for swflite.


mrchnk:
I pull another request to `develop` branch #2100
I guess that request should be rejected.

PR #2108, opened 2018-12-22
Scroll to selection

MSGhero:
Closes #1467, #1825

Fixes `scrollV`, `maxScrollV`, and `bottomScrollV` behavior. Also adds "scroll to selection" as you type in an input field, move the caret with the arrow keys, or use `setSelection()`.

It could be optimized I bet.


MSGhero:
Another alternative is for LRUD text navigation to use `setSelection()` instead of manually setting the values. It would reduce the number of lines a bit, functionality would be the same.

MSGhero:
Whoops... dev is targeted now

MSGhero:
Alright this works now

src/openfl/text/TextField.hx
```diff
@@ -51,84 +53,84 @@ import js.html.DivElement;
 #end
 
 /**
-	The TextField class is used to create display objects for text display and
```
Gama11: Looks like there were some changes to the doc comment style you'd have to merge manually.

MSGhero:
I've been running this for a month without issues.

PR #2112, opened 2018-12-27
Fix GL error for cube textures

player-03:
Currently if you try to use `CubeTexture`, OpenGL throws `INVALID_OPERATION` because you passed a cube-format texture (rather than a 2D texture) to `__bindGLTexture2D()`. `Context3D` already has some code to check the texture format and call the correct function, so I simply copied it over.


jgranick:
Thank you :smile:

PR #2120, opened 2019-01-09
Fix issue with drawQuads calculating incorrect graphics bounds

JoeCreates:
I believe the calculation of the graphics bounds (minX, maxX etc) was incorrect if any drawQuads rects are not at position 0, 0. This resulted in bounds sometimes being way larger than necessary, or so small that some quads didn't display when they were supposed to.

If I'm not mistaken, the `rects` are positions in the texture, not the position of the final tile, and so the x and y should not be used when determining how to expand the graphics bounds.


JoeCreates:
Wrong branch, sorry. First attempt at trying a PR directly from GitKraken.

Gama11:
Fyi, GitHub allows changing the target branch these days.

PR #2121, opened 2019-01-09
DrawQuads should not use rect position for graphics bounds

JoeCreates:
I believe the calculation of the graphics bounds (minX, maxX etc) was incorrect if any drawQuads rects are not at position 0, 0. This resulted in bounds sometimes being way larger than necessary, or so small that some quads didn't display when they were supposed to.

If I'm not mistaken, the `rects` are positions in the texture, not the position of the final tile, and so the x and y should not be used when determining how to expand the graphics bounds.


jgranick:
Thanks, that looks great. You are right, it's being transformed twice for bounds calculations, which is just the wrong behavior :disappointed_relieved:

PR #2125, opened 2019-01-19
Return the actual object in ByteArray.readObject with AMF

chatziko:
format's AMF reader returns an enum (format.amf(3).Value) wrapping the returned value to easily get the object's type. Although the use of enums is cool, ByteArray.readObject should return the actual object, not an enum, this is how it works for all other encodings (and also in Flash).

```
var obj = { a: 1 };
trace(obj);
// output: { a : 1 }

var ba = new ByteArray();
ba.objectEncoding = ObjectEncoding.HXSF;
ba.writeObject(obj);
ba.position = 0;
var newobj = ba.readObject();
trace(newobj);
// output: { a : 1 }

var ba2 = new ByteArray();
ba2.objectEncoding = ObjectEncoding.AMF3;
ba2.writeObject(obj);
ba2.position = 0;
var newobj2 = ba2.readObject();
// output: AObject({a => AInt(1)},null)
```

This PR recursively unwraps the enum and returns the actual object.

### DIFF
--- a/src/openfl/utils/ByteArray.hx
+++ b/src/openfl/utils/ByteArray.hx
@@ -1,6 +1,8 @@
 package openfl.utils;
 
 
+import haxe.Constraints.IMap;
+import haxe.ds.ObjectMap;
 import haxe.io.Bytes;
 import haxe.io.BytesData;
 import haxe.io.BytesInput;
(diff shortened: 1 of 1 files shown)


chatziko:
Thanks for merging!

I can write something. Btw to run the unit tests you just do ```haxelib run munit test``` inside ```test/haxelib/unit```? I get

```
Type not found : massive.munit.Async
```

and it only works after I comment out this import from all the files it appears. Using munit 2.3.2.

jgranick:
You can do it one of two ways:

```
cd openfl/test/haxelib/unit
haxelib run munit test
```

or

```
cd openfl
hxp unit-test
```

The `hxp` command is meant to be an easier shortcut. We have a little Haxe script file that helps run any other commands we need for the unit/integration/functional testing. We're using a fork of the munit library at the moment, with fixes to async testing.

```
git clone https://github.com/openfl/munit
haxelib dev munit munit/src
```

I think that's the error you're experiencing

PR #2140, opened 2019-02-26
Read bitmap from ShaderBuffer instead of Shader

starry-abyss:
When using drawQuads one can attach the same shader in shader fill several times in a frame.
For the most part, data are read from ShaderBuffer, but UVs are screwed (because bitmap width/height is read for wrong bitmap) unless one creates an unique Shader for each shader fill.
This change makes shared shaders work again, at least in a mini project of mine.

**Before:**

default

default

**After:**

default

default

### DIFF
--- a/src/openfl/_internal/renderer/context3D/Context3DGraphics.hx
+++ b/src/openfl/_internal/renderer/context3D/Context3DGraphics.hx
@@ -74,7 +74,7 @@ class Context3DGraphics
 					}
 					else
 					{
-						bitmap = c.shaderBuffer.shader.__bitmap.input;
+						bitmap = c.shaderBuffer.inputs[0];
 					}
 
 				case DRAW_QUADS:


starry-abyss:
Is the change good enough to be merged, @jgranick ?

src/openfl/_internal/renderer/context3D/Context3DGraphics.hx
```diff
@@ -74,7 +74,7 @@ class Context3DGraphics
 					}
 					else
 					{
-						bitmap = c.shaderBuffer.shader.__bitmap.input;
+						bitmap = c.shaderBuffer.inputs[0];
```
jgranick: It probably is right to use the cached inputs instead of referencing the shader again -- but what if the shader has more than one input? Is it always index 0?

Thanks :smile:

starry-abyss: From my understanding, we only use `bitmap` in this function to calculate UVs. And what we pass to `drawQuads` (first argument) provides only one set of UVs per quad...
So to support multiple UVs here we first need to make a new `drawQuads`? :thinking:

jgranick: A custom shader can take from multiple shader inputs. It will only have one set of texture UVs, but it still is possible for someone to sample from more than one texture.

For example, say you had a color texture and an alpha texture, and sampled from both to create the final result. There are many more complex things a user can achieve. Most shaders will probably have the one input only, but they may have more. Maybe we can iterate over inputs to find the matching one?

starry-abyss: Yes, I intended to use multi-texturing too. But I thought OpenFL was using UVs from the first bitmap?

> Maybe we can iterate over inputs to find the matching one?

What is the concept of "matching input"? How OpenFL is intended to determine which input holds the UVs we need? Is there a flag the user passes to mark some input as primary?

jgranick:
Please try the latest commit. You are right that we should use a cached bitmapData, but we do need to make sure it's the correct input :smile:

starry-abyss:
Thank you, works for my simple test case! :)
So it was indeed some cryptic flag that pointed to the right input, would took a long time if I had to search it myself...

PR #2143, opened 2019-03-02
Fix hasRect() always false for non square tiles

pixelsyntax:
fix: Tileset.hasRect() always returns false for non square tiles due to comparing rect.width == tileData.height


pixelsyntax:
getRectID() was similarly affected with the same typo

PR #2153, opened 2019-03-12
Fix infinite loop in TextEngine

### LINKED ISSUE #2124: Infinite loop in TextEngine.getLayoutGroups()
I encountered that issue with html5 target on latest haxelib and npm distributions (8.8.0 at moment)

I extracted reproducible project with problematic swf asset at https://github.com/mrchnk/openfl-issue-textengine

That loop:
https://github.com/openfl/openfl/blob/8.8.0/src/openfl/_internal/text/TextEngine.hx#L1083

OS (Windows 10 x64)
Haxelib version 3.3.0
Haxe version 3.4.7

haxelib list
```
openfl-issue-textengine> haxelib list
actuate: [1.8.9]
box2d: [1.2.3]
format: [3.4.1]
hxcpp: [4.0.4]
hxp: [1.0.5]
layout: [1.2.1]
lime-samples: [7.0.0]
lime: 7.2.0 [7.2.1]
openfl-samples: [8.7.0]
openfl: 8.7.0 [8.8.0]
```

dpomier:
This PR fixes #2124

### DIFF
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -1233,7 +1233,7 @@ class TextEngine
 			{
 				// if a line break is the next thing that needs to be dealt with
 
-				if (textIndex <= breakIndex)
+				if (textIndex < breakIndex)
 				{
 					setFormattedPositions(textIndex, breakIndex);
 					placeText(breakIndex);


jgranick:
I had to revert this because it broke the unit testing, but perhaps there is another way to prevent the infinite loop?

MSGhero:
Can you reopen 2124 so we don’t forget about it? My initial guess is that this issue is really an unhandled html text case.

Gama11:
Huh, you should have the permissions to do that yourself?

MSGhero:
I’m on my phone :)

PR #2155, opened 2019-03-13
Adding SWFLite.symbolsByClassName to speed up searching for an asset by className

sobotklp:


jgranick:
Do you think that the SWFLite file format should include the `symbolsByClassName` data, or should it be generated at runtime when the SWFLiteLibrary is first launched or when a symbol name lookup is first requested?

sobotklp:
to be honest, I'm not sure :). Generally, I prefer to err on the side of better runtime performance, but I could be convinced either way. It does seem redundant to add this mapping to the SWFLite .bin files, since the data is already there, just not in a convenient data structure.

PR #2156, opened 2019-03-16
Don't crash when weird text layout bugs arise

MSGhero:
Feel free to suggest a better warning message.

### DIFF
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -956,7 +956,7 @@ class TextEngine
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


Gama11:
How about:

>You found a bug in OpenFL's text code! Your project has automatically been zipped and sent to Joshua Granick (@singmajesty) so we can fix this.

MSGhero:
I don't think people in copyright-land would appreciate sending their project off without permission.

Gama11:
I guess I forgot the sarcasm tag. x)

jgranick:
It's like winning the lottery!

PR #2160, opened 2019-03-25
scale9Grid: added support in CanvasBitmap and Context3DBitmap

joshtynjala:
I tried to include CairoBitmap and DOMBitmap as well, but I ran into difficulties with those two. I'm sure that someone more familiar with the OpenFL/Lime rendering system, and those platforms, would have an easier time with them than I did.

In the case of CairoBitmap, I am not familiar with Cairo at all. The current rendering seems to render the image all at once, and I could not find a way to draw it in nine separate parts, like CanvasBitmap, or to specify some kind of vertices to stretch internal parts of the image, like Context3DBitmap.

I tried rendering DOMBitmap to canvas when scale9Grid was specified. I was thinking that I could reuse much of the code I had already written CanvasBitmap. However, there seemed to be smoothing issues that caused opacity issues at the seams.

### DIFF
--- a/src/openfl/_internal/renderer/canvas/CanvasBitmap.hx
+++ b/src/openfl/_internal/renderer/canvas/CanvasBitmap.hx
@@ -39,7 +39,71 @@ class CanvasBitmap
 
 			if (scrollRect == null)
 			{
-				context.drawImage(bitmap.__bitmapData.image.src, 0, 0, bitmap.__bitmapData.image.width, bitmap.__bitmapData.image.height);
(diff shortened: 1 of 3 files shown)


joshtynjala:
@jgranick Thank you!

jgranick:
@joshtynjala We're currently looking at migrating code into `Context3DGraphics` and `CanvasGraphics` for scale-9 in `sprite.graphics` similar to Flash. 

Happy to discuss more on https://discord.gg/tDgq8EE

PR #2170, opened 2019-04-25
Fix issue with incorrect cached bitmap dimensions on DisplayObject resize when filter applied

grayhaze:
When a `DisplayObject` has a filter applied, such as a `ShaderFilter`, the dimensions of the cached bitmap for the object don't always match the dimensions of the object when they change. This change fixes this problem by always using the filter's bounds as the dimensions for the cached bitmap.


grayhaze:
This relates to this issue I described in the flixel channel of the Haxe Discord, which concerned a shader not appearing correctly when resizing a HaxeFlixel camera object: https://discordapp.com/channels/162395145352904705/165234904815239168/569870006465200138

I'll try to put together a pure OpenFL-based example over the weekend.

grayhaze:
Closing for now, as it seems that the original behaviour was intended to prevent constant allocation of new buffers every time an object changes size. @jgranick has suggested some possible alternative approaches around modifying the texture coordinates passed to the shader, but I'm not sure how to tackle that myself.

PR #2171, opened 2019-04-26
count scrollRect in __worldTransform calculation

mrchnk:


ubald:
Thanks 👏🏻 that did it for me! I had the same issue.

jgranick:
If the purpose is to trigger `__updateCacheBitmap`, is there another we can make that occur without changing the world transform? I believe that altering this may break issues with things like mouseX/mouseY values and other logical position issues that are uniquely interesting about `scrollRect`

ubald:
I would need to check if my issues with scale and `cacheAsBitmap` are related to this change or not. Other than that I haven't found issues with the solution here. I just came back to an old project of mine (CPP, on Linux) and I'm trying to get it working with the latest versions of lime/openfl.

PR #2174, opened 2019-05-02
bugfix : New logic for swf asset classes caused generated classes to extend null

m0rkeulv:
Classes generated for normal SWF Movieclip assets  ends up extending null with current code.
ex:
```
class MySwfAsset extends null {
```

i am not sure if the `if` statement was intended to be inside the `case` statement (never seen that used before) but moving it down to the "case logic" part seems to solve the problem.

### DIFF
--- a/scripts/Tools.hx
+++ b/scripts/Tools.hx
@@ -229,9 +229,11 @@ class Tools
 					var superClassData = swf.data.abcData.resolveMultiNameByIndex(classData.superclass);
 					switch (superClassData.nameSpace)
 					{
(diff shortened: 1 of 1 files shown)


Gama11:
I'm not sure why that would make a difference, that's simply a pattern guard...

https://haxe.org/manual/lf-pattern-matching-guards.html

m0rkeulv:
hm... i can't reproduce the bug anymore, could be a problem caused by not properly rebuilding tools after i switched branch.  now after rebuilding tools i cant seem to reproduce it.

PR #2191, opened 2019-05-29
Fixed runtime error in VideoTexture - "canplay" handler wasn't being cleaned up

TomByrne:
Fixed issue where "canplay" event was coming through after VideoTexture has been disposed, throwing a runtime error:

TypeError: Cannot read property '__video' of null


Gama11:
Something seems to be wrong with indentation (files uses tabs, but your changes use spaces).

Note that OpenFL uses [haxe-formatter](https://github.com/HaxeCheckstyle/haxe-formatter). If you're using VSCode, you can just press Ctrl+S to re-format the file.

TomByrne:
Ok, I've resaved with auto-formatting

PR #2209, opened 2019-07-04
A new take on GlowFilter GL path.

thunerli:
GL implementation of all 4 variations of GlowFilter; glow, inner glow, knockout glow and inner knockout glow.


MSGhero:
This is cool

thunerli:
Pushed DropShadowFilter as well

ianharrigan:
Does this also work in conjunction with scrollrects? It seems that was usually what was causing a number of issues (at least for me)

PR #2211, opened 2019-07-05
fixed culling for Graphics::drawTriangles

Klug76:
Fixed wrong behavior for culling.NONE polygons (both-sided). For such polygons we got random flicked culling (depend on previous culling flag).
This screenshot before fix:
before_fix
After fix:
after_fix

### DIFF
--- a/src/openfl/_internal/renderer/context3D/Context3DGraphics.hx
+++ b/src/openfl/_internal/renderer/context3D/Context3DGraphics.hx
@@ -744,6 +744,9 @@ class Context3DGraphics
 								case NEGATIVE:
 									context.setCulling(BACK);
 
+								case NONE:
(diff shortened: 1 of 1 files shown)


jgranick:
Awesome thanks

Klug76:
I'm slightly unsure about this code below:
```
							if (culling != NONE)
							{
								context.setCulling(BACK);
							}
```
it would be better to remember the previous value of context:culling and restore it here?

jgranick:
I think it would be better (long-term) if each draw call specified the type of culling it expects and then we can apply the change only if the context is not already using the same type of culling.

This code is here because other draw calls are not aware (currently) of the culling type and just generally expect it to use back face fulling by default

PR #2219, opened 2019-07-17
Adding a sorting function to TileContainers

miltoncandelero:
Not sure about the ITileContainer modification. Can be removed and still works fine.


Gama11:
Looks like the indentation is inconsistent with the other code. You can just run haxe-formatter over it to fix it.

jgranick:
Thanks!

I just renamed to `sortTiles` and made a few aesthetic changes... also I set it to flag the render as dirty after running it

PR #2241, opened 2019-08-22
Fix the problem that Chinese cannot be rendered properly.

rainyt:
I think that Haxe 4.0 has improved UTF-8 characters. This fixes the
Chinese rendering problem of C++ (Android & IOS).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,31 @@
+8.9.2 (08/20/2019)
+------------------
+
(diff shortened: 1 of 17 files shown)


MSGhero:
Here is the Harfbuzz source that needs to get called: https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-buffer.cc#L1534-L1592

This solution only accounts for the last two lines, which are the most important but don't account for every case. See #2300

PR #2280, opened 2019-10-22
less gl commands calls

Beeblerox:
@thunerli after this commit https://github.com/openfl/openfl/commit/e9f04643a893b08102e30ec22e76bfb34f3a7335 the number of gl commands calls have increased significantly.
And with this change i've decreased the number from around 12900 to 11300 in my game.
Could you review it?

### DIFF
--- a/src/openfl/_internal/renderer/context3D/Context3DRenderer.hx
+++ b/src/openfl/_internal/renderer/context3D/Context3DRenderer.hx
@@ -764,9 +764,6 @@ class Context3DRenderer extends Context3DRendererAPI
 		if (__numClipRects > 0)
 		{
 			__numClipRects--;
-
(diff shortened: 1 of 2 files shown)


thunerli:
There was an issue with scrollRect, we weren't flushing the batcher at the right time. These three commits work together to fix the scrollRect in GL path: e9f04643a893b08102e30ec22e76bfb34f3a7335 044926965a571a7e6c67274b832403df94c6fee6 and d677dae9bc9a8706f0f4067412de67519d61ecc7
Reverting the first commit can break scrollRect again, could you please confirm that's not the case?

thunerli:
Tested this out, looks good.

Beeblerox:
@thunerli thanks!

PR #2281, opened 2019-10-22
CUSTOM_RENDER_OBJECT: a new type in DisplayObjectType

Klug76:
``DisplayObjectType::CUSTOM_RENDER_OBJECT``: a new type of object is added so that no other rendering methods are called during custom rendering.
For example, if a custom object is inherited from ``DisplayObject``, it may not need to call the ``__renderShape`` method before calling ``RenderEvent``.

### DIFF
--- a/src/openfl/_internal/renderer/DisplayObjectType.hx
+++ b/src/openfl/_internal/renderer/DisplayObjectType.hx
@@ -11,6 +11,7 @@ enum DisplayObjectType
 	TEXTFIELD;
 	TILEMAP;
 	VIDEO;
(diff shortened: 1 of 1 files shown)


jgranick:
What if the object uses `__type = null`?

Klug76:
``__type = null`` looks  simpler, but now it will not work due to this condition: ``if (object != null && object.__type != null)``.  This condition does not allow rendering at all. In addition, a switch with an null enum value will throw an exception in c++ target:
null_enum_switch_exception

jgranick:
I just merged it in as `CUSTOM`

PR #2283, opened 2019-10-23
Bitmap should use width/height getters of DisplayObject

Beeblerox:
This fixes the case when Bitmap is rotated with 90/270 degrees, which leads to "swap" between Bitmap's width and height values

### DIFF
--- a/src/openfl/display/Bitmap.hx
+++ b/src/openfl/display/Bitmap.hx
@@ -132,16 +132,19 @@ class Bitmap extends DisplayObject
 
 	@:noCompletion private override function __getBounds(rect:Rectangle, matrix:Matrix):Void
 	{
+		var bounds = Rectangle.__pool.get();
 		if (__bitmapData != null)
 		{
-			var bounds = Rectangle.__pool.get();
 			bounds.setTo(0, 0, __bitmapData.width, __bitmapData.height);
-			bounds.__transform(bounds, matrix);
-
-			rect.__expand(bounds.x, bounds.y, bounds.width, bounds.height);
-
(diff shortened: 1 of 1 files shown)


Beeblerox:
plus i've made changes to width/height setters so they produce the same result as on flash target

Beeblerox:
my simple test for result comparison is following:
```
package;

import openfl.display.Bitmap;
import openfl.display.BitmapData;
import openfl.display.Sprite;

class Main extends Sprite 
{
	public function new() {
		
		super ();
		
		var bd:BitmapData = new BitmapData(200, 100, true, 0xffff0000);
		
		var b1:Bitmap = new Bitmap(bd);
		b1.rotation = 30;
		var b2:Bitmap = new Bitmap();
		b2.rotation = 30;
		
		trace("b1.width: " + b1.width);
		trace("b1.height: " + b1.height);
		
		trace("b1.scaleX: " + b1.scaleX);
		trace("b1.scaleY: " + b1.scaleY);
		
		trace("b2.width: " + b2.width);
		trace("b2.height: " + b2.height);
		
		trace("b2.scaleX: " + b2.scaleX);
		trace("b2.scaleY: " + b2.scaleY);
		
		b1.width = 300;
		
		trace("after width set: ");
		
		trace("b1.scaleX: " + b1.scaleX);
		trace("b1.scaleY: " + b1.scaleY);
		
		b2.width = 300;
		
		trace("b2.scaleX: " + b2.scaleX);
		trace("b2.scaleY: " + b2.scaleY);
	}	
}
```

PR #2292, opened 2019-11-08
Fix for rightMargin with center-justified text.

justin-espedal:
I wrote this after only skimming the code and getting a basic understanding. It seems `leftMargin` isn't needed here because it's later accounted for by the calls to `getBaseX()` in `placeFormattedText()`. Without this, center-justified text is off-center when a right margin is specified.


MSGhero:
Nice catch! The format aligns could use a looking over after the margins and indents PR, for sure.

PR #2293, opened 2019-11-09
Fix preventDefault() on a mouse wheel event

Klug76:
This change enables ``preventDefault()`` on the mouse wheel event.
Now user can scroll the data only in a certain list, not the entire web page.


Klug76:
Note: see https://github.com/haxelime/lime/issues/743

Issue #2230, opened 2019-08-05
Double-click selectable textfield should select charachters under


intoxopox:
Double-clicking inside a textfield should select any contiguous characters beneath the point clicked. Currently, it does not on non-flash targets.

intoxopox:
Any status updates or insights on this?

MSGhero:
It's just not implemented yet. I imagine the solution would be to add a double click event, use getCharIndexAtPoint in the callback, look for the previous and next space/newline/etc, and then setSelection on those indices to highlight the text.

intoxopox:
Thinking about tackling this, if no one else is on it yet.

MSGhero:
Go for it

intoxopox:
Think I got it. Do I just submit a PR and reference this issue number?

MSGhero:
Yep and you're in luck, I just hooked up the internet to my dev PC. So I'll be able to test your changes once you post the PR (and generally code more often).

PR #2298, opened 2019-12-03
Added doubleclick select word behavior, fixing issue #2230

intoxopox:
Added doubleclick select word behavior, fixing issue #2230


src/openfl/text/TextField.hx
```diff
@@ -3187,6 +3191,54 @@ class TextField extends InteractiveObject
...
+					if (DisplayObject.__supportDOM)
+					{
+						if (__renderedOnCanvasWhileOnDOM)
+						{
+							__forceCachedBitmapUpdate = true;
+						}
+					}
+					else
+					#end
+					{
```
MSGhero: The { } don't line up with anything in the case of !js

intoxopox: Oops. Fixed it. While I was in there, I noticed the same error in stage_onMouseMove, btw. Unless you see a reason to leave it be, I'll fix that one too.

MSGhero: I'll just do that one to keep the PR clean if it ever needs to be reverted

src/openfl/text/TextField.hx
```diff
@@ -3059,6 +3063,55 @@ class TextField extends InteractiveObject
 			scrollV -= event.delta;
 		}
 	}
+	
+	@:noCompletion private function this_onDoubleClick(event:MouseEvent):Void
+	{
+		if (__textEngine.selectable) {
```
MSGhero: I would do `selectable` (i.e. `this.get_selectable()`), just in case the getter ever needs to be modified

src/openfl/text/TextField.hx
```diff
@@ -3059,6 +3063,55 @@ class TextField extends InteractiveObject
...
+					pos = txtStr.lastIndexOf(c, startPos - 1);
+					if (pos > leftPos) leftPos = pos + 1;
+
+					pos = txtStr.indexOf(c, startPos);
+					if (pos < rightPos && pos != -1) rightPos = pos;
+				}
+				
+				if (leftPos != rightPos)
+				{
+					__selectionIndex = leftPos;
```
MSGhero: `setSelection()` will adjust the scrollV as well which you need, instead of manually setting the indices

MSGhero:
Triple click selects the whole line it seems :) and it's a true triple click, not double clicking already highlighted text or whatever. Not sure how that works, and that can be a separate issue

src/openfl/text/TextField.hx
intoxopox: Sure thing.

MSGhero:
Thanks!

Issue #2237, opened 2019-08-21
Problem with cyrillic symbols on Android


flashup:
Hello, I've updated lime, openfl and haxe to the last versions and tried to run TextMetrics project from `openfl-samples`.
I added some cyrillic text to strings in sources.
Flash, html5, neko render text properly, but android renders wrong.
```
Haxe RC-3
Lime 7.6.0
OpenFL 8.9.2
HXCPP 4.0.19
```
Flash:
![flash](https://user-images.githubusercontent.com/2893619/63424747-c29ef500-c417-11e9-81f4-085398ec0817.jpg)

Neko:
![neko](https://user-images.githubusercontent.com/2893619/63424748-c468b880-c417-11e9-85c4-3dd5ab472b62.jpg)

Android:
![android](https://user-images.githubusercontent.com/2893619/63424751-c6327c00-c417-11e9-98a4-41924d12960a.jpg)

MSGhero:
Can you try a Windows/Mac compile with your current versions? There's an issue with rc4 and 5, and I want to see if this is the same or different.

flashup:
@MSGhero what I need to do?

MSGhero:
Instead of building android, build windows or Mac and see if the same problem happens

flashup:
Same lib versions, couldn't compile iOS:
```
Compiling group: runtime
xcrun --sdk iphoneos12.4 clang++ -D_CRT_SECURE_NO_DEPRECATE -DHX_UNDEFINE_H -c -stdlib=libc++ -O2 -arch armv7 -isysroot /Applications/Xcode_10.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS12.4.sdk -miphoneos-version-min=5.0 -fobjc-arc -DOBJC_ARC -DENABLE_BITCODE=YES -fembed-bitcode -Wno-parentheses -Wno-null-dereference -Wno-unused-value -Wno-overflow -Wno-trigraphs -Wno-bool-conversion -fno-stack-protector -DIPHONE=IPHONE -DIPHONEOS=IPHONEOS -DSTATIC_LINK(static) -DHXCPP_ARMV7 -DHXCPP_VISIT_ALLOCS(haxe) -DHX_SMART_STRINGS(haxe) -DHXCPP_API_LEVEL=400(haxe) -I/usr/local/lib/haxe/lib/hxcpp/4,0,19/include -fexceptions -fstrict-aliasing ... tags=[haxe]
[...]

PR #2300, opened 2019-12-07
Fix UTF16 text on Haxe4

### LINKED ISSUE #2264: Cyrillic text issue after updating to haxe 4.0.0-rc.5 (in openfl TextField)
After upgrading haxe I get correct traces of Cyrillic strings’ lengths and values, but the TextField itself looks like:
Screenshot 2019-09-21 at 20 21 43

However tracing the textField.text gives a correct result in the console.

This is on mac/ios targets. `toUpperCase()` and `length` work as expected for Cyrillic strings, but the text looks like:
Screenshot 2019-09-22 at 03 15 47

On neko target `toUpperCase()` and `length` don't work correctly, but the text is displayed ok:
Screenshot 2019-09-22 at 03 17 14

Here is the full code to reproduce it:
```
	var _body:TextField = new TextField();
        addChild(_body);
	_body.text = 'фыва йцук';
```
Should I downgrade back to 3.4.7 version or there might be another way around?

MSGhero:
Closes #2264, #2237 

Proper UTF16 support via the Harfbuzz API's intended function. I can't easily test Mac, Linux, or HL, so I'm not sure the conditional is still correct, if someone else can check. See #2264 for a good example: the text will render incorrectly if this change is required.

### DIFF
--- a/src/openfl/_internal/text/TextLayout.hx
+++ b/src/openfl/_internal/text/TextLayout.hx
@@ -137,12 +137,13 @@ class TextLayout
 			__hbBuffer.script = script.toHBScript();
 			__hbBuffer.language = new HBLanguage(language);
 			__hbBuffer.clusterLevel = HBBufferClusterLevel.CHARACTERS;
-			// #if ((haxe_ver < "4.0.0") || neko || mac || linux || hl)
+			#if (neko || mac || linux || hl)
+			// other targets still uses dummy positions to make UTF8 work
+			// TODO: confirm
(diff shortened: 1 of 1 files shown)


flashultra:
@MSGhero I just test on iOS and this fix cyrillic text ( mention in  #2264 ) , but still have problem with symbols like 😀 . Is there something specific for ios ?

MSGhero:
There's currently no support for emojis 😢 it's on the list though 📑

PR #2303, opened 2020-01-02
added BlendMode.ERASE support for BitmapData.draw()

Klug76:
Added ``BlendMode.ERASE`` support for ``BitmapData.draw()`` only, ``DisplayObject::blendMode=ERASE`` still not supported.


Klug76:
This PR is not applicable after the current backend refactoring. I (temporarily) cancel it.

PR #2324, opened 2020-03-10
Adding weakReference support for EventListeners when targeting native.

m0rkeulv:
In order to improve GC performance on native targets i have added support for weak references for EventListeners so that the `useWeakReference` flag actually do something.

i kept it simple and only made one class that handle both strong and weak references, if preferred i could change it into  2 distinct classes, WeakReference , StrongReference or do other changes to it.

### DIFF
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -1,5 +1,6 @@
 package openfl.display;
 
+import openfl.utils.ReferenceProxy;
 #if !flash
 import openfl._internal.renderer.cairo.CairoBitmap;
 import openfl._internal.renderer.cairo.CairoDisplayObject;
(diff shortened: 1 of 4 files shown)


jgranick:
When I last tested weak references with HXCPP they were cleaned up improperly.

I'm trying to remember but I think there are two kinds of references which I will refer to as "top down" and "bottom up" references.

The "top down" reference would be something like a static map that is holding a list of references to otherwise unreferenced objects. The "top" is holding references down to other objects. A weak map should keep these references but these were getting killed in my previous tests

The "bottom up" reference would be something like children of an object referring back up to the parent. In this approach the parent was never collected even though no "parent" references referred to it. 

I could be remembering incorrectly but I had instances where both 1.) objects I expected to remain disappeared and 2.) objects I expected to be collected remained in memory... in other words, the _wrong_ objects were collected.

Perhaps HXCPP changes, changes in how we handle garbage collection in Lime or both have brought us to the point that we can reliably expect weak references on the C++ target. If so, this would be great news!

m0rkeulv:
I don't know the exact HXCPP implementation details, but the way i expect weakRef to work is that if the WeakRef object is the last/only object with a given reference the GC is allowed to free that object and weakRef.get()  will return null.  except that it should behave just like any other object when it comes to GC.

i have been using this implementation for a while (probably 3 weeks or so) and not seen any problems so far. nor has there been any change in errors on google PlayStore or complaints from my players.

joshtynjala:
I added weak references in the following commits for 9.3.0, so I'm closing this PR.

de3a3b6372f106ce7cdcc106de5d4502b18f065e
e6f089ccfe6349a223a79f0b58cf70dde96f0415
2f0cbdb20bc890477523c8d9e401e6e6e248b420

PR #2326, opened 2020-03-11
Update TouchEvent.hx

ConfidantCommunications:
I removed the reference to GestureEvent classes since they don't exist.


joshtynjala:
Closing because I don't want to remove docs that may be relevant again in the future, if implementations are added. It is still technically accurate for the air target anyway.

PR #2328, opened 2020-04-06
Fix UTF16 on Linux

MSGhero:
Confirmed with two people that Linux should be using the UTF16 fix from #2300


MSGhero:
Fixes Android as well it seems.

PR #2330, opened 2020-04-08
FileReference.upload implementation

kasparsj:
preliminary upload implementation.

```
var file = new FileReference();
var uploadURL = new URLRequest();
uploadURL.url = "http://example.com/";
// important to override idleTimeout, by default it's 30 seconds
uploadURL.ildeTimeout = 120000; // two minutes
file.upload(uploadURL);
```


ROBERT-MCDOWELL:
Does OpenFL not already implemented upload?

jgranick:
@joshtynjala Raised an important issue in the comments

We can only merge code that is compatible with an MIT license. I apologize for the direct merge rather than a review:

 1.) Please submit code that is MIT
 2.) Do not include any comments (such as @author, @see, etc)
 3.) Use `openfl.net._internal.UploadPostHelper` rather than `openfl._internal.utils.UploadPostHelper`

Thanks!

PR #2332, opened 2020-04-14
Inline Graphics __calculateBezierCubicPoint & __calculateBezierQuadPoint

intoxopox:
Inline __calculateBezierCubicPoint & __calculateBezierQuadPoint on targets except js to avoid method call in curveTo and cubicCurveTo.


intoxopox:
Oops... I didn't mean to pollute with all these commits. A gitKraken mistake on my part.

Gama11:
Should be ok if GitHub's squash-merge option is used. :)

PR #2339, opened 2020-05-15
Fix for the wrong behavior of HOME/END keys

mikhailitsky:
Incorrect behavior of HOME and END buttons. Currently, whenever HOME/END is pressed, this results in selecting a fragment of text from the current cursor position to the beginning/end of the line. The correct behavior would be just moving the cursor to the beginning/end of the line without selecting any text, whereas text selection should happen when HOME/END is pressed together with Shift. The proposed change fixes this issue.

### DIFF
--- a/src/openfl/text/TextField.hx
+++ b/src/openfl/text/TextField.hx
@@ -1525,29 +1525,13 @@ class TextField extends InteractiveObject
 
 	@:noCompletion private function __caretBeginningOfLine():Void
 	{
-		if (__selectionIndex == __caretIndex || __caretIndex < __selectionIndex)
-		{
-			__caretIndex = getLineOffset(getLineIndexOfChar(__caretIndex));
-		}
-		else
-		{
-			__selectionIndex = getLineOffset(getLineIndexOfChar(__selectionIndex));
-		}
+		__caretIndex = getLineOffset(getLineIndexOfChar(__caretIndex));
 	}
 
 	@:noCompletion private function __caretEndOfLine():Void
 	{
-		var lineIndex;
-
-		if (__selectionIndex == __caretIndex)
-		{
-			lineIndex = getLineIndexOfChar(__caretIndex);
-		}
-		else
-		{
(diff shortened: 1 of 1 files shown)


MSGhero:
Not sure when this will get merged, but take a look at https://github.com/openfl/openfl/tree/text/home-end-lrud-fix and see if there's any differences. I didn't PR this branch yet because I haven't really tested it

mikhailitsky:
Hi, sorry, I'm just making the first baby steps in Git. Is this the commit
you are suggesting to take a look at?
[image: image.png]

MSGhero:
I can't see your image, but it's this one: https://github.com/openfl/openfl/commit/11c09bdf365adac2612c9fc78915f9a3c81a9acf

MSGhero:
Closing this one as well, but your changes are like exactly the same as mine for the Home/End part, so nice job!

mikhailitsky:
> Closing this one as well, but your changes are like exactly the same as mine for the Home/End part, so nice job! #2345

In my fix there were also calls of updateScrollH after HOME and END presses. Can you check if this is still needed in your fix. This was needed for when the text was wider than the control and pressing HOME/END was not scrolling the string into the visible boundaries.

mikhailitsky:
Also, check if 

```
if (__caretIndex < 0)
{
	__caretIndex = __text.length;
	__selectionIndex = __caretIndex;
}
```
is needed in __caretBeginningOfLine () and __caretEndOfLine (), because pressing HOME/END in a textless text field would prevent the cursor from blinking without these lines.

MSGhero:
Yes, so for your first comment, I split scrollV/H and navigation into 2 PRs, so that testers could focus on a single change. `updateScrollH()` is added in the scrolls PR. My goal is to put `updateScrollH()` into `setSelection()`, but I didn't want to add too much to the PR.

For the second, those lines already exist in `startCursorTimer()`, which gets called in `setSelection()`. So that should be covered.

PR #2340, opened 2020-05-18
Fix for the position of the cursor

mikhailitsky:
When there is no text in the text field, the blinking cursor should be positioned according to the alignment setting of the text field. Without this fix, the cursor is blinking on the left regardless of the alignment setting.


MSGhero:
Addressing in #2427

PR #2342, opened 2020-05-18
Unescaping should happen after parsing

mikhailitsky:
HTML unescaping should happen after parsing, not before, because unescaping before parsing defeats the whole purpose of escaping. The purpose of escaping is to prevent the HTML service characters from being parsed as service characters. For example, when I assign a `"&amp;gt;"` to TextField.htmlText, I'm expecting to see a `"&gt;"` on the screen, not a `">"` (which is the case now).


MSGhero (changes requested):
Changes make sense. Needs to be cherry picked onto a post-9.0 branch though

MSGhero:
Updated in #2428

PR #2344, opened 2020-05-20
Fixes for bottomScrollV and maxScrollV

mikhailitsky:


MSGhero:
I'm going to close this because I have a much more comprehensive and tested branch, and this looks like it may break a few things. It's definitely appreciated, though.

I'll tag you in the new PR so you can try it out.

mikhailitsky:
> I'm going to close this because I have a much more comprehensive and tested branch, and this looks like it may break a few things. It's definitely appreciated, though.
> 
> I'll tag you in the new PR so you can try it out. #2346

Hi, the current fix in `get_bottomScrollV ()` still has some issues. Mainly, the problem here is that the bottomScrollV value is returned as a 0-based index, whereas according to the Flash specs it must be a 1-based line number. In order to see how the current fix misbehaves, try the following code:

`var msg : String = "1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15";
var txt : TextField = new TextField ();

txt.border = true;
txt.borderColor = 0x000000;
txt.type = TextFieldType.INPUT;
txt.multiline = true;
txt.wordWrap = true;
txt.x = 30;
txt.y = 30;
txt.width = 100;
txt.height = 15;
txt.text = msg;
		
addChild (txt);`

and try scrolling the text by incrementing/decrementing  scrollV by 1 and monitor the bottomScrollV value - it will be getting the wrong values. I've figured that my previously suggested fix had some issues as well, so I've worked it out further and now seem to have nailed it. I will create a new pull request with a suggested fix.

Also, maxScrollV needs fixing too, as it does not produce the correct values. My previously suggested fix for this function had an issue as well, so I have reworked it and will provide it in this new pull request too along with explanations.

MSGhero:
It's not zero-based, it messes up in the edge case when the height of the TextField is less than the height of a character +4. `bottomScrollV` seems fine otherwise. `maxScrollV` seems fine, do you have a test case?

mikhailitsky:
Right, the above code produces the edge case, which does not work as expected. I have created a new pull request where this works as expected: https://github.com/openfl/openfl/pull/2348. However, just to make sure we are on the same page - is the bottommost visible line supposed to be displayed even when it is just partially visible? I was assuming yes, and my suggested fixes in this new pull request reflect exactly that.

MSGhero:
No, it’s not supposed to be visible. It used to be in OpenFL, and I think that caused confusion, but it should only show full lines with the exception of the current line at scrollV.

mikhailitsky:
Ok, that makes sense. I've updated my suggested solution at https://github.com/openfl/openfl/pull/2348 accordingly.

PR #2346, opened 2020-05-20
scrollV and scrollH fixes

MSGhero:
Fixes scrollV and brings scrollH to parity with Flash.

With every font and font size I tested, scrollH gets bumped in increments of 24 px, so that is captured here.


MSGhero:
@mikhailitsky

PR #2348, opened 2020-05-20
Fixes for bottomScrollV and maxScrollV, reiterated

mikhailitsky:
1. bottomScrollV: according to the Flash specs, this value is a 1-based line number, so the 0-based 'i' from the loop must be incremented by 1 to become a 1-based line number. Also '<' vs. '<=' - the first line that partially crosses the 'height - 4' boundary should be deemed bottomScrollV, that's why there should be a '<' instead of a '<='.
2. maxScrollV: this is basically just the reverse of the above algorithm. The same is applicable to the 1-based line number and the '<'. Also, there was a code that was removing a trailing line feed. Not sure what was the logic behind this, but if I had, for example, a multiline editable control with the text having a trailing line feed and the text height exceeding the text field height, I would never be able to scroll down enough to reach the line feed and edit it because of this code, which is wrong. Also, that code was removing only 1 trailing line feed, so if there were more than one trailing line feeds, only the last one would be removed, while the other one(s) would remain anyway.

### DIFF
--- a/src/openfl/_internal/text/TextEngine.hx
+++ b/src/openfl/_internal/text/TextEngine.hx
@@ -1784,18 +1784,19 @@ class TextEngine
 		else
 		{
 			var tempHeight = 0.0;
-			var ret = numLines;
+			var ret = scrollV;
 			var firstVisible = scrollV - 1;
 			
 			for (i in firstVisible...lineHeights.length)
 			{
-				if (tempHeight + lineHeights[i] < height - 4)
+				if (tempHeight + lineHeights[i] <= height - 4)
 				{
 					tempHeight += lineHeights[i];
+					ret = i + 1;
 				}
 				else
 				{
-					return (i + 1);
+					break;
 				}
 			}
 			
@@ -1812,20 +1813,20 @@ class TextEngine
 		}
 		else
 		{
-			var i = numLines - 1, tempHeight = 0.0;
+			var ret = numLines, tempHeight = 0.0;
 
-			while (i > 0)
+			for (i in lineHeights.length - 1...0)
 			{
-				if (tempHeight + lineHeights[i] < height - 4)
+				if (tempHeight + lineHeights[i] <= height - 4)
 				{
 					tempHeight += lineHeights[i];
-					i--;
+					ret = i + 1;
 				}
 				else
 					break;
 			}
 
-			return (i + 1);
+			return ret;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


MSGhero:
A lot of the behavior you describe as "wrong" is the correct implementation according to Flash. I updated my PR to fix the edge case you brought up.

mikhailitsky:
I have updated my suggested solutions here to reflect the requirement to display only fully visible functions.

I referred to the removal of 1 trailing line feed as wrong because it is not the way it works in Flash. It does not allow to edit the last line if it consists of a '\n' only. There also was a problem in maxScrollV caused by i += 2. I did not save the test case reproducing this, but that was the reason why I started to dig into maxScrollV in the first place. If I stumble upon that test case, I will post it here. In the meantime, check out the fixes I'm suggesting. I believe they are cleaner as there is a clear transition from a 0-based vector/array index to a 1-based line number, which eliminates the need to spaghettify the code with extra checks. My suggested solutions make in turn TextField::__updateScrollV() as simple as this:

```
@:noCompletion private function __updateScrollV():Void
	{
		__layoutDirty = true;
		__updateLayout();

		var lineIndex = getLineIndexOfChar(__caretIndex);

		if (lineIndex == -1 && __caretIndex > 0)
		{
			// new paragraph
			lineIndex = getLineIndexOfChar(__caretIndex - 1) + 1;
		}

		if (lineIndex + 1 < scrollV)
		{
			scrollV--;
		}
		else if (lineIndex + 1 > bottomScrollV)
		{
			scrollV++;
		}
		else
		{
			// TODO: can this be avoided? this doesn't need to hit the setter each time, just a couple times
			scrollV = scrollV;
		}		
	}
```

which is not only simpler, but faster.

(my suggested fixes require __updateScrollV to be changed this way to work right).

MSGhero:
I'll retest trailing newlines and all other edge cases, but what I saw before was that Flash does not include the final trailing newline in maxScrollV. I also thought iterating backwards caused a compiler error...

mikhailitsky:
You mean this line?

`for (i in lineHeights.length - 1...0)`

I'm using Haxe 4.0.5, it seems to be compiling it flawlessly.

MSGhero:
image
It should be compatible with Haxe 3.

MSGhero:
I was wrong about maxScrollV, so I deleted the part subtracting 1 from it. The rest doesn't account for many cases, such as when the scroll needs to update by more than 1 line or the case explicitly commented in maxScrollV. One of the scrollV issues came from a reversion, so I put that back in as your PR does. I'll commit the fixes in the following days, but I've checked that all of your core issues have been addressed. I think you still misunderstand what bottomScrollV actually represents, so I confirmed my fixes against Flash.

PR #2349, opened 2020-05-21
Added Tab Indent support and replaced showText with showGlyphs to solve distorted text rendering when selected

Dimensionscape:
Added tab indents if tabEnabled property is false


Dimensionscape:
Seems I need to make some further changes yet.

PR #2350, opened 2020-05-21
Fix socket problem when debugging  cpp targets

m0rkeulv:
Fix socket problem when debugging  cpp targets, error.Custom(EOF) would cause the connection to be closed when it should have been ignored.

`NativeSocket.socket_recv(__s, buf.getData(), pos, len);` will return "EOF"  when there is no more data when you build your code for cpp/windows. This is then wrapped  and thrown as  `Error.Custom(EOF)` and so  just ignoring `haxe.io.Eof` with the `catch (e:Eof)`  block in socket.hx wont be enough to handle end of file scenarios.
```

### DIFF
--- a/src/openfl/net/Socket.hx
+++ b/src/openfl/net/Socket.hx
@@ -1087,10 +1087,19 @@ class Socket extends EventDispatcher implements IDataInput implements IDataOutpu
 			}
 			catch (e:Error)
 			{
-				if (e != Error.Blocked)
+				switch  (e)
 				{
-					doClose = true;
+					case Error.Custom(type):
+						if(type != "EOF")
+						{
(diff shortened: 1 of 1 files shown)


MSGhero:
Wait is this why the debugger randomly exits sometimes while I’m inspecting variables?

m0rkeulv:
> 
> 
> Wait is this why the debugger randomly exits sometimes while I’m inspecting variables?
not sure how the debugger uses sockets but  i think this is more  OpenFL as its about how openFL handles sockets.

The debugger crash could be bore related to this tho https://github.com/HaxeFoundation/hxcpp/pull/908

MSGhero:
Bless

PR #2379, opened 2020-07-09
Fix Android Performance

Vortelio:
**Support:** Tilemap, Bitmap, drawTriangles, drawQuads

BunnyMark

Snapdragon 625 1920x1080
before 60fps ~9500 bunnies
after 60fps ~23000 bunnies

Asus ME173X 1280x800
before ~15fps 2000 bunnies
after ~60fps 2000 bunnies


barisyild:
Are there any performance improvements in HTML5 and Windows targets?

Vortelio:
On PC cpp / html5 target, there is no difference.
Emphasis on the CPU and not the GPU. 100% load of one core and the FPS begins to fall.
before fix: ~ 230k bunnies cpp / html5
after fix: ~ 230k bunnies cpp / html5

barisyild:
There may be a difference in computers with low processor capacity on html5/cpp platform.

Vortelio:
I think to see the difference need a powerful CPU and a weak video card.

Vortelio:
CPU: i3-5005u, GPU: Intel(R) HD Graphics 5500

cpp
before: 60fps, ~13000 bunnies, CPU: 12%, GPU: 100%
after: 60fps, ~13000 bunnies, CPU: 12%, GPU: 100%

html5
before: 60fps, ~12000 bunnies, CPU: 13%, GPU: 100%
after: 60fps, ~12000 bunnies, CPU: 13%, GPU: 100%

Apparently the original version of the shaders does not present any problems for the PC, only for mobile devices.

Vortelio:
Returned compatibility with inheritance.
Added support for drawRect.
Added accuracy for android (mediump).

Support: Tilemap, Bitmap, drawTriangles, drawQuads, drawRect

BunnyMark

**Snapdragon 625 1920x1080**
before: 60fps ~9500 bunnies
after: 60fps ~23000 bunnies 
after: + ( precision mediump float ) 60fps ~26000 bunnies

**Asus ME173X 1280x800**
before ~15fps 2000 bunnies
after ~60fps 2000 bunnies

loudoweb:
Have you tested colorTransform and filters with your shader? Is it working the same as before?

Vortelio:
I use colorTransform all the time in my application and it looks 100% the same. Of the filters I've only used BlurFilter and it works just as well as before.

m0rkeulv:
I just looked over the code quickly to see if i could understand what makes this faster, I am understanding it correctly if what you've done is to move the colorTransform calculations from the fragment shader to the  vertex shader, and then just multiply with the colorMatrix in the fragment shader ? @Vortelio 

I can see how this could make some performance improvements on devices with limited GPU power, and I don’t see any immediate drawbacks.

Vortelio:
> move the colorTransform calculations from the fragment shader to the vertex shader, and then just multiply with the colorMatrix in the fragment shader ?

Yes exactly. I think that nothing should break.

m0rkeulv:
i am a bit uncertain about why you removed  the offsets  and replaced them with multipliers,  what happens if my code uses both an offset and a multiplier?
//removed
```
tempColorTransform.redOffset = color.r;
tempColorTransform.greenOffset = color.g;
tempColorTransform.blueOffset = color.b;
```
//added
```
tempColorTransform.redMultiplier = color.r / 0xFF;
tempColorTransform.greenMultiplier = color.g / 0xFF;
tempColorTransform.blueMultiplier = color.b / 0xFF;
```

Vortelio:
> ```
> tempColorTransform.redMultiplier = color.r / 0xFF;
> tempColorTransform.greenMultiplier = color.g / 0xFF;
> tempColorTransform.blueMultiplier = color.b / 0xFF;
> ```

It will definitely not work correctly. I will do a visual test to accurately control the changes. But now it is already clear that colorOffset will work correctly only through a fragment shader, and the whole point of fixing is lost. ColorMultiplier will work correctly through the vertex shader.

Vortelio:
A little test shows that drawRect worked correctly in openFL 3.6.1.


In openFL 8.9.7, with the current implementation drawRect will fail to render correctly. It is difficult to get the correct color from the black texture. And colorMultiplier doesn't work at all.



For the colorOffset support, the pixel color is used, which means that you cannot leave the fragment shader.
colorMultipliers works great through the vertex shader.

There are many options further:

- Separate shader for drawRect.
- Different shaders with fast / slow if colorOffset is active.
- Shaders with colorMultipliers only for Tilemap, drawQuad, etc.

The current patch only works correctly with colorMultiplier, even a drawRect created from a white texture.

Vortelio:
Support: Tilemap, Bitmap, drawTriangles, drawQuads, **+bitmapFill**

drawRect is the correct color.
8 9 7-fix

Full colorTransform support:
If colorOffset is used then the standard shader is used, if colorOffset is zero then the fast shader is used.

PR #2395, opened 2020-08-04
Use lighter weight Shape instead of Sprite for DefaultPreloader parts.

intoxopox:


intoxopox:
Arg. Only the last commit was meant to go in to this very simple PR. I keep accidentally forgetting to make a clean branch for PRs. I apologize yet again.

intoxopox:
I'm going to clean this up and re-submit.

Issue #2382, opened 2020-07-17
Fix JS library script


jgranick:
The JS library script is not compiling:

https://github.com/openfl/openfl/blob/develop/scripts/build.hxml

Testing can be done running `haxe build.hxml` from the "scripts" directory

It may be missing some of the internal package paths:

```
../packages/application/src
../packages/assets/src
../packages/bitmapdata/src
../packages/displayobject/src
../packages/errors/src
../packages/eventdispatcher/src
../packages/externalinterface/src
../packages/filereference/src
../packages/filters/src
../packages/geom/src
../packages/graphics/src
../packages/input/src
../packages/loader/src
../packages/movieclip/src
../packages/printing/src
../packages/renderer-cairo/src
../packages/renderer-canvas/src
../packages/renderer-core/src
../packages/renderer-dom/src
../packages/renderer-flash/src
../packages/renderer-stage3D/src
../packages/security/src
../packages/sensors/src
../packages/shader/src
../packages/sharedobject/src
../packages/socket/src
../packages/sound/src
../packages/stage/src
../packages/stage3D/src
../packages/system/src
../packages/telemetry/src
../packages/textfield/src
../packages/tilemap/src
../packages/urlloader/src
../packages/urlrequest/src
../packages/urlstream/src
../packages/utils/src
../packages/video/src
```

It references also an [ImportAll file](https://github.com/openfl/openfl/blob/develop/scripts/ImportAll.hx) to pull in each OpenFL class name. Last I checked this file included a private class for SWF support which should be removed since SWF support has been moved to a different haxelib

PXshadow:
Hello, I'm attempting to work on this task however I'm confused how to figure out "It may be missing some of the internal package paths" if this is the case or not, also I removed swf lib and did ```haxe build``` and it seems to be building fine, I'm pretty sure I'm missing something fundamental if you could point in the right direction I'd love to help complete this task.

jgranick:
Hi @PXshadow, thank you :grin:

"It may be missing some of the internal package paths" means the build.hxml should have all of the above referenced paths in place. If it compiles without one or more of these paths then it means that (perhaps?) the ImportAll.hx file is missing a reference.

For example, if ImportAll.hx does not include `TextField` then it might compile without the path to "packages/textfield/src" but this means the JS library will be incomplete.

Once the script is compiling again without errors testing the output can be done in a separate task :smile:

PR #2397, opened 2020-08-08
Fix JS library script 

PXshadow:
builds build.hxml fixing #2382


jgranick:
I could have sworn I merged this 😱

I was confused when it threw the compile error regarding the SWF library class reference in import all 😔

PXshadow:
No worries at all, I think I missed something, but I wanted to push something to at least give you something concrete. My apologies regardless.

PR #2405, opened 2020-09-17
[WIP] Add some TextField tests

MSGhero:
WIP, documenting my current text test progress


MSGhero:
https://gist.github.com/MSGhero/848dabee2b1f210abfb252857e7fcc2a

Will get these integrated, but they all work locally against Flash except one that is probably a Flash bug

MSGhero:
Wait I messed up

PR #2416, opened 2020-10-08
Add more textfield tests

MSGhero:
Redo of additional TextField tests. These haven't been tested in the NPM environment or anything, so I'm leaving it as a draft PR for now. I tested the logic on Neko and Wincpp and compared with Flash.


MSGhero:
https://gist.github.com/MSGhero/848dabee2b1f210abfb252857e7fcc2a

PR #2420, opened 2020-11-11
Fix: propagating same rect in __getFilterBounds

alexk74:
An empty rect should be passed to every child filter bounds calculation and then it should expand the container rect if that is the case.


alexk74:
It says discord config is missing and that is why the checks have failed. I have no idea how to setup discord or if it's something that I can do.

PR #2422, opened 2020-11-19
drawRect + colorTransform fix

deadbaad:
Fixes rendering of color-transformed rectangles.

### DIFF
--- a/packages/renderer-stage3d/src/openfl/display/_internal/Context3DGraphics.hx
+++ b/packages/renderer-stage3d/src/openfl/display/_internal/Context3DGraphics.hx
@@ -33,7 +33,7 @@ import openfl.display._internal.stats.DrawCallContext;
 @SuppressWarnings("checkstyle:FieldDocComment")
 class Context3DGraphics
 {
-	private static var blankBitmapData = new BitmapData(1, 1, false, 0);
+	private static var blankBitmapData = new BitmapData(1, 1, false, 0xFFFFFFFF);
 	private static var maskRender:Bool;
 	private static var tempColorTransform = new ColorTransform(1, 1, 1, 1, 0, 0, 0, 0);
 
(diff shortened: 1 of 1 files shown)


packages/renderer-stage3d/src/openfl/display/_internal/Context3DGraphics.hx
```diff
@@ -33,7 +33,7 @@ import openfl.display._internal.stats.DrawCallContext;
 @SuppressWarnings("checkstyle:FieldDocComment")
 class Context3DGraphics
 {
-	private static var blankBitmapData = new BitmapData(1, 1, false, 0);
+	private static var blankBitmapData = new BitmapData(1, 1, false, 0xFFFFFFFF);
```
deadbaad, line 36: Offsets are applied after multipliers, so to make color fill + color transform to work properly, we need base bitmap to be whole white and color/color transform should be applied as a multiplier combo.

packages/renderer-stage3d/src/openfl/display/_internal/Context3DGraphics.hx
```diff
@@ -650,9 +650,9 @@ class Context3DGraphics
 
 								#if lime
 								var color:ARGB = (fill : ARGB);
-								tempColorTransform.redOffset = color.r;
-								tempColorTransform.greenOffset = color.g;
-								tempColorTransform.blueOffset = color.b;
+								tempColorTransform.redMultiplier = color.r / 0xFF;
+								tempColorTransform.greenMultiplier = color.g / 0xFF;
+								tempColorTransform.blueMultiplier = color.b / 0xFF;
```
deadbaad, line 655: Fill color is nothing more than another base white color transformation.

Vortelio:
I suggest adding colorOffset
https://github.com/openfl/openfl/pull/2379/files#diff-3193e2ea471e431642d9c7c0f33691ef0637bcca0618637830b0e9fb6b62f248

deadbaad:
> I suggest adding colorOffset
> https://github.com/openfl/openfl/pull/2379/files#diff-3193e2ea471e431642d9c7c0f33691ef0637bcca0618637830b0e9fb6b62f248

Sure, both seem to do effectively the same job. I see you're doing some performance optimizations by using specialized "fast" shaders, so pre-baking offsets into multiplier may be beneficial to your implementation, right? In this case I suggest adding another more specialized "solid color fill" shader (i.e. the one rendering solid color passed via vec4 uniform, but with neither texture, nor color transform uniforms), which may (or may not :)) give a performance boost by excluding a redundant load on texture sampling pipeline.

PR #2429, opened 2020-12-05
Text Renderer fixes for text selection

dimensionscape:
Fixes for text selection when aligned center or right. Previously, contrast bar was offset.


MSGhero:
Discord integration is the only thing causing a unit test failure... interesting

NipponSunrise:
According to code seems to be done. Thanks. I'll make my own ad-hoc patch later this day (EU timezone) and let you know.

PR #2430, opened 2020-12-06
Fix cursor position on align in canvas & cairo renderers.

dimensionscape:
This solution feels a little dirty, but it works. Maybe there is a better way at handling it, but I think it is a reasonable fix until then.


MSGhero:
Does it work with scrollH? You'll have to turn wordWrap off to test it

dimensionscape:
The problem was only persistent when the field was empty due to some logic in the render loop that needed something to be there to make calculations, therefore the solution **only** affects the field if there is **no text** and is ignored when the field contains glyphs. It has no direct effect on the caret index or anything like that. 

However, I did create a new issue with fields that have a border or background, and I understand why, so I need to go back and make some small change.

dimensionscape:
Ok I amended it to fix the issue with bordered/background fields.

PR #2433, opened 2020-12-07
Fix cursor alignment on empty field

dimensionscape:


dimensionscape:
Ok, this should be my last amendment to this pr. As you can see, I left END and START open so that we could add these at a later time if necessary. I don't think they actually do anything right now anyway. In the future, it will be necessary to also make a minor adjustment for bullets. Are bullets even working now?

PR #2434, opened 2020-12-07
Implementation for mouse drag scrolling via selection

dimensionscape:
Its a good start for something we've been missing. We could make it more flash like or use the render loop + mouseMove, although I didn't because it was kind of too slow. Room for improvement or modification of scroll speeds.


CanIGetaPR:
Could you squash these commits into one? You could use `git rebase -i HEAD~3` for this operation.

Issue #2377, opened 2020-07-05
Text performance and (auto) RTL


MSGhero:
I have some ideas for how to make text layout faster. It does seem to be this and not rendering that ends up being the bottleneck.

- [ ] Adding, removing, or changing text only affects the layout of the changed paragraphs and not any others, excluding y offsets. Right now, inserting a character by typing, `replaceText()`, or `appendText()` will re-layout the entire TextField when only that single paragraph needs to be looked at. This implies there is a `Paragraph` class or a `paragraphIndices` int vector since searching for line breaks each time would be too slow
- [ ] Maybe only a single line needs to be looked at at a time, but it recursively looks at other lines that were affected by the initial one, up to the entire paragraph
- [ ] From experience, it would make development faster if all of TextEngine and TextField utilized `Paragraph` and `Line` classes to keep the variables organized. I imagine there would be a lot less iteration in general. Any function that iterates over every layout group in the TextField could instead iterate over the ones in a line or paragraph
- [ ] Harfbuzz's docs say that it creates "layout strategy" objects that can be passed in as an initial heuristic. If each paragraph stored its own layout strategy and fed it into Harfbuzz, that maybe speeds things up. There could be other functions as well
- [ ] Quite a few cycles can be spent on dealing with ligatures when it's not clear that there are any ligatures to begin with. Maybe Harfbuzz reports this kind of stuff that could be taken advantage of?

I can see the `Paragraph` and `Line` classes helping make RTL implementation easier, since it would all be neatly defined in one place. It does bring up how RTL will actually be handled.

- [ ] Because of the potential for bi-directional text, the text `direction` needs to be defined for each layout group. It may make more sense for it to be defined in `TextFormat`, to give users a way to use it within the public API. RTL isn't handled by `flash.text.TextField`
- [ ] The resources exist to auto-detect whether a layout group should be RTL or LTR. It's more expensive than letting the user define it themselves, but it would allow input text to auto adjust itself. Maybe this is something the user implements instead, or maybe it occurs based on `tf.detectDirection = true;`

intoxopox:
@MSGhero With your recently merged text work, how is this checklist looking now?

MSGhero:
Like even more work 😂 Nothing gets checked off, but I can work on it now without overlapping code with another PR. I'm positive that Flash doesn't regen the whole TextField most of the time after testing some issue, and that's an immediate and massive perf boost to input text.

Issue #2229, opened 2019-08-02
[OpenFL 8.9.1] Performance issues in HTML5 targets with TextField.appendText(...) 


streiffus:
In HTML5 targets, using the appendText(...) function in a TextField instance causes a significant pause if a lot of text is on display.

I made a test case app that can be viewed here: http://tamarcurry.com/openFLTextPerformance.

From what I can tell, this is due to the getLayoutGroups() function in openfl._internal.text.TextEngine.hx making numerous calls to __context.measureText(...), iterating through several loops, and many internal functions.

We feel finding a solution would benefit the entire OpenFL community and have created a bug bounty for this issue. The details of it can be found here: https://github.com/streiffus/OpenFLTextPerformance.

larsiusprime:
Tagging @MSGhero and @jgranick :)

MSGhero:
It's like 1 ms for each function call on Flash and doesn't increase, so I guess it's not recalculating the entire field every time. Idk how to get around `measureText()` since I know nothing about JS, but we can probably make it 7 ms each append for this test case.

Pretty big scope actually, so it's on my list but not gonna resolve it right away. An issue with your solution is that the kerning can vary between characters based on the font, so caching two letter widths and measuring the width of two letters can be different. Arabic and Farsi are the extreme cases where letters actually combine.

larsiusprime:
Let it be known that I'll match @streiffus bug bounty of $50, bringing the total up to $100, as I can foresee needing this myself in a future project.

MSGhero:
Well, if anyone wants to give it a try, my solution would be for `getLayoutGroups` to take `textIndex` as an argument instead of always starting at 0. This would let you “resume” the layout at the point where appending starts. 

Gotta make sure that the previous text already on that first line you append gets its baseline and max line height set properly. Maybe Flash recalculates the entire last line, or maybe it can resume the layout function cleanly. You can tell by turning wordWrap off in the append example above, deleting any newlines in the text, and seeing if the time is constant or increases on Flash (would increase if recalculating the entire line, which is getting longer and longer now). 

Kind of a minor concern is if the first appended character is supposed to combine with the last character of the original string. I don’t fully understand how it works yet, but that could force you to recalculate the line as opposed to cleanly resuming.

MSGhero:
[...]

PR #2437, opened 2020-12-10
O(1) to O(reasonable) text insertions

MSGhero:
This is a naïve text-layout-reduction PR: only laying out text after a "placement index" aka the insertion point. There are a couple of unfinished edge cases, but I may leave it since you typically don't have aggressive text formatting when you also are inserting text frequently. More people complain about input text being unbearable, and this helps especially if the cursor is near the end of the text.

It's a little hard to evaluate the "state" of text layout at any index, so it's harder to resume from that state. There are some bigger changes I have planned that address this, but those are a ways away.

Related to #2377, #2229. I'll decide if the latter can be closed when I merge.

### DIFF
--- a/packages/textfield/src/openfl/text/TextField.hx
+++ b/packages/textfield/src/openfl/text/TextField.hx
@@ -2031,7 +2031,7 @@ class TextField extends InteractiveObject
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


MSGhero:
The word cache PR nicely addresses performance issues, and this one is too fussy for where the layout code is at right now

PR #2439, opened 2020-12-14
Prevent caret from returning to beginning on DOM

dimensionscape:


packages/textfield/src/openfl/text/TextField.hx
```diff
@@ -1376,11 +1376,20 @@ class TextField extends InteractiveObject
...
 		__updateScrollH();
+		
+		#if dom 
+		if(stage!=null && stage.focus == this && __renderedOnCanvasWhileOnDOM)
+		{
+			__stopCursorTimer();
+			__startCursorTimer();
+		}
+		#else
 		if(stage!=null && stage.focus == this)
```
CanIGetaPR: This is creating duplicate code, I would suggest in-lining the if dom to prevent code duplication.

if(stage!=null && stage.focus == this 
#if dom && __renderedOnCanvasWhileOnDOM #end)

Something like that.

CanIGetaPR (changes requested):

packages/textfield/src/openfl/text/TextField.hx
CanIGetaPR: Could also be something like

var domCaretPositionFix = #if dom __renderedOnCanvasWhileOnDOM #else true #end

MSGhero:
Now, is this a fix or a workaround for another issue somewhere else? IIRC `set_text()` was being called along with `set_htmlText()`, which in theory shouldn't happen at the same time. Would actually setting text on DOM scroll you back up to the top, and setting htmlText keep you at the bottom?

dimensionscape:
> 
> 
> Now, is this a fix or a workaround for another issue somewhere else? IIRC `set_text()` was being called along with `set_htmlText()`, which in theory shouldn't happen at the same time. Would actually setting text on DOM scroll you back up to the top, and setting htmlText keep you at the bottom?

This happens any time you stop and start the cursor in dom. In 9.0.2, this was happening simply from selecting text too. This should fix any related caret issues. Is there another way to solve this? Maybe.

MSGhero:
I guess my question is, after merging this in, should we also create a new issue, or is everything set?

dimensionscape:
> 
> 
> I guess my question is, after merging this in, should we also create a new issue, or is everything set?

Well I dont think the text setter should be getting hit on every update, but I think that is a separate problem all together. It definitely needs to be investigated.

PR #2440, opened 2020-12-15
Skip discord notify action on forks

CanIGetaPR:
Fixes the unit test action failing in fork pull requests since forks don't have the Discord webhook token secret.


CanIGetaPR:
Note, the github action for this PR will still fail since it's using the old configuration file from the parent branch.

PR #2442, opened 2020-12-22
Simplify OpenFL library

jgranick:
This simplifies the OpenFL library structure as discussed.

1. The "lib", "lib-esm" and NPM-specific files have been removed to another repository
2. All of the source files are moved again into the same directory. This simplifies access for new contributors, restores mapping with the API documentation "View Source" links, reduces reliance on macros or Lime tools in order to compile using the library
2. All of the API documentation is moved back into the source files. This is necessary based on current tooling in order to provide code completion within Visual Studio Code, HaxeDevelop and other compatible code editors
3. The OpenFL "_internal" classes are distributed within the OpenFL package (such as "openfl.text._internal" instead of "openfl._internal.text"). I am open to feedback on this. I believe this may help bring internal classes closer to the public classes, similar to the previous "packages" structure
4. Unit tests are distributed by the "packages" structure into smaller sets of functionality. This brings the focus of the intent behind the packages refactor. Each can be compiled using a single `hxp` command within its folder. Each unit can compile with or without Lime (or run twice, once in each case) or other permutations to help unit test it (for example, the "textfield" unit ideally could run once for Haxe-based layout and once for Harfbuzz/C++ based layout within the unit)

The intent is to bridge the gap between the OpenFL library of previous versions (say, OpenFL 6 or 7) before NPM support was added, which has led to confusion and overwhelm for contributors, and with the narrower "packages" structure performed for OpenFL 9, in order to cut OpenFL mentally into smaller blocks. The above structure attempts to balance both, with the following perspective:

- "openfl" as a library is for Haxelib only. It is for Haxe only. It is meant to be a library only, so "fanciness" (tools, requiring tools, build scripts, etc) should be eliminated if possible or kept to a strict minimum
- NPM support will be handled as a separate project which uses the Haxe "openfl" library as a dependency in its build process


jgranick:
Merged

Issue #2446, opened 2020-12-25
[html] scrollRect is not working properly for Bitmap with canvas back-end


mrchnk:
When using scrollRect for a Bitmap image is not cropped, but scaled into the size of scrollRect.

I created a minimal project that will reproduce the issue using openfl 8.9.6. You can replace openfl.js to test it with any other openfl version. You can checkout [sources](https://github.com/mrchnk/openfl-issue-scrollRect), or just [run it](https://htmlpreview.github.io/?https://github.com/mrchnk/openfl-issue-scrollRect/blob/master/pub/index.html).

**Actual**
![actual](https://github.com/mrchnk/openfl-issue-scrollRect/raw/master/pub/actual.png)

**Expected**
![expected](https://github.com/mrchnk/openfl-issue-scrollRect/raw/master/pub/expected.png)

**OpenFL Targets**
That is applicable **only** for **html** target with **canvas** back-end renderer

**Additional context**
I'm using the distributed version of openfl.js

PR #2447, opened 2020-12-25
Fix rendering Bitmap with scrollRect on canvas

mrchnk:
Method [CanvasRenderingContext2D.drawImage](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage) have 3 signatures

```
void ctx.drawImage(image, dx, dy);
void ctx.drawImage(image, dx, dy, dWidth, dHeight);
void ctx.drawImage(image, sx, sy, sWidth, sHeight, dx, dy, dWidth, dHeight);
```

Suitable for that case is the third with both source and destination dimensions.
I don't actually know why `dx` and `dy` are expected to be not 0, but passing the exact same `rect` to destination works well.

### DIFF
--- a/src/openfl/display/_internal/CanvasBitmap.hx
+++ b/src/openfl/display/_internal/CanvasBitmap.hx
@@ -43,7 +43,8 @@ class CanvasBitmap
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


mrchnk:
Closes #2446

PR #2448, opened 2020-12-25
Assigning a colorTransform shall flag the displayObject as dirty

fedor-strelkov:
Assigning displayObject.transform.colorTransform flags the display object as dirty.

This is a way to enable typical color adjustment method - https://books.openfl.org/openfl-developers-guide/display-programming/manipulating-display-objects/adjusting-displayobject-colors.html

### DIFF
--- a/src/openfl/geom/Transform.hx
+++ b/src/openfl/geom/Transform.hx
@@ -207,7 +207,7 @@ class Transform
 			__displayObject.__setRenderDirty();
 		}
 
-		return value;
+		return __colorTransform;
 	}
 
 	@:noCompletion private function get_concatenatedMatrix():Matrix
--- a/src/openfl/geom/Transform.hx
+++ b/src/openfl/geom/Transform.hx
@@ -199,19 +199,15 @@ class Transform
 
 	@:noCompletion private function set_colorTransform(value:ColorTransform):ColorTransform
 	{
-		if (!__colorTransform.__equals(value, false))
+		if (value != null)
 		{
 			__colorTransform.__copyFrom(value);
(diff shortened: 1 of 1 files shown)


fedor-strelkov:
```
var ct:ColorTransform = obj.transform.colorTransform;
ct.color = newColor;
obj.transform.colorTransform = ct;
```
A code like this shall effectively change the display object color. In fact it did not change the color on screen because the object was not marked as dirty. There are two ways to fix it:

1) To modify the getter to return a copy of the color transform:
```
private function get_colorTransform():ColorTransform {
  return __colorTransform.clone();
}
```
2) To set the dirty flag every time the setter is called. This approach I have chosen here as it doesn't produce a clone object.

jgranick:
Thank you!

I think we need to return a new object, because otherwise other code will also fail, like:

```haxe
var ct1 = sprite.transform.colorTransform;
ct1.blueMultiplier = 0.5;
var ct2 = sprite.transform.colorTransform;
ct2.alphaMultiplier = 0.2;
trace(ct1 == ct2); // true
```

I updated the internals so that the clone is not called internally, though now it looks like colorTransform really should be an internal property on `DisplayObject` and `Transform` is just an accessor (similar to how it works for matrix, concatenatedColorTransform, etc)

TODO :wink:

fedor-strelkov:
Joshua, the fix works well for me. Thanks!
I have commented on [your commit baffe62](https://github.com/openfl/openfl/commit/baffe62a4426fb8194bd0977b5640b79ad1237fd): need also to change the return value of the setter.

PR #2449, opened 2020-12-25
Fix major performance issues for textfield

dimensionscape:
Creating a word cache prevents expensive shaping from occurring over and over again when unnecessary

### DIFF
--- a/src/openfl/text/TextFormat.hx
+++ b/src/openfl/text/TextFormat.hx
@@ -211,6 +211,7 @@ class TextFormat
 
 	@:noCompletion private var __ascent:Null<Float>;
 	@:noCompletion private var __descent:Null<Float>;
+	@:noCompletion private var __cacheKey:String;
 
 	/**
 		Creates a TextFormat object with the specified properties. You can then
@@ -289,7 +290,10 @@ class TextFormat
 
 		newFormat.__ascent = __ascent;
 		newFormat.__descent = __descent;
-
+		newFormat.__cacheKey = __toCacheKey();
+		
+		
+			
 		return newFormat;
 	}
 
@@ -316,7 +320,16 @@ class TextFormat
 
 		if (format.__ascent != null) __ascent = format.__ascent;
 		if (format.__descent != null) __descent = format.__descent;
+		
+		__toCacheKey();
+		//Used to speed up position measurements in TextEngine by way of a cache
+	}	
+	
+	@:noCompletion private function __toCacheKey():String
+	{
+		return __cacheKey = '$font$size$bold$italic';
 	}
+	
 }
 #else
 typedef TextFormat = flash.text.TextFormat;
--- a/src/openfl/text/_internal/TextEngine.hx
+++ b/src/openfl/text/_internal/TextEngine.hx
@@ -1,6 +1,7 @@
 package openfl.text._internal;
 
 import haxe.Timer;
+import haxe.ds.StringMap;
 import openfl.display3D._internal.GLTexture;
 import openfl.utils._internal.Log;
 import openfl.Vector;
@@ -93,6 +94,7 @@ class TextEngine
 	@:noCompletion private var __textFormat:TextFormat;
 	@:noCompletion private var __textLayout:TextLayout;
 	@:noCompletion private var __texture:GLTexture;
+	@:noCompletion private var __wordCache:StringMap<Array<#if (js && html5) Float #else GlyphPosition #end>> = new StringMap();
 	// @:noCompletion private var __tileData:Map<Tilesheet, Array<Float>>;
 	// @:noCompletion private var __tileDataLength:Map<Tilesheet, Int>;
 	// @:noCompletion private var __tilesheets:Map<Tilesheet, Bool>;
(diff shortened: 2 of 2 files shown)


MSGhero:
A decent hashing function could be the font's name --> some combination of bold, italics, and font size (`size + (bold ? 1 << 31 : 0) + (italics ? 1 << 30 : 0)`) --> positions. Maybe they could be combined further by using the index within `Font.registeredFonts`, since your font size or index shouldn't be super high (`+ index << 16`). Size is a float, but whatever.

dimensionscape:
I did something similar, will add the commit in a little bit.

dimensionscape:
I decided not to use a hash function at all for now since the performance benefit was a bit moot after testing because the data set was so small. I could have chosen to make it a bit faster by using a single map, but I opted to set it up this way for a much smarter caching system later on that will clean itself up and be shared by more than one text engine/text field. Overall the benefit is enormous compared to what we were facing before. Its not really a simple problem...many engineers for google Blink's text layout team have been working on ways to improve shaping performance for years based on everything I've read and its still not perfect. We still need to fix the slowdown caused by multiple characters without a break, but I dont think it is relative to this.

MSGhero:
Maybe a limit on how big a cluster can be. I think I remember seeing 15 chars. It wouldn't be great for an entire paragraph of JP text to get cached. 

I think there are some more changes that have to be made elsewhere before script and direction can be added to this. So, pretty much done, can't wait to try it out!

MSGhero:
It's worse for text without spaces. Entire paragraphs of CJK text get cached for no reason, so memory will bloat a bit. There would probably be noticeable memory leaks if we had IME input support.

We don't have support for `textLayout.language` to define or autodetect CJK yet, so it's probably best to put a character limit on the cache. As above, chromium puts a 15 or 16 char limit on words.

Edit: minor, but `bold` and `italics` in the hash could be shortened to single letters, and they could be further combined into a single number via binary.

dimensionscape:
Yeah, I have a smarter way of doing the cache in my head which will mostly add to this, and I suppose we will have to get language support if we really want it to be efficient for CJK too. Theres really nothing wrong with entire paragraphs getting cached, thats what the cache is for. If it gets measured as a cluster, thats what its going to cache. So yeah, if we do that, we have to break clusters up into smaller sizes without separating ligatures, and uh are ligatures even handled for CJK? I don't know much about those languages.

At any rate, this has more to do with how we cluster text and not how we cache it. I cant really put a word limit on the cache, it doesnt work like that. This needs to be handled before it gets to the cache. In other words, we need to limit the range of positions asked for earlier in the logic.

As for memory considerations, right now it has no self management. Its completely dumb, except for the fact that it will not cache clusters of the same value more than once. In certain languages without spaces and line breaks, this would definitely become an issue without some kind of smarter management, including detecting languages. There is also some more going on with typing repeated characters without a break that needs to be resolved or streamlined.

I made it as simple as possible so that it was very easy to understand and less likely to cause any hard issues, but its an incremental improvement(even if it is a big one!). On top of it, I really wanted to get it out for Christmas so people could take advantage of the fix. 😄

My next move is to separate almost all of the cache logic out of text engine so that we can keep things clean, simple and compartmentalized so that the cache can evolve without touching the other text internals. I think this will be a safer way forward as we make the cache smarter and more complicated. Then I will focus on smart memory management and shared caching between multiple textfields so we can avoid duplicate caching and provide removal of unused or unnecessary caches for things like static text in a more sophisticated and intelligent way. At least for the moment, its cleaned up by GC when the textfield is no longer referenced making it a bit safer in terms of creating any sort of memory leak.

Shortening the key by getting the integer values of bold and italics instead of the full text is a good idea.

MSGhero:
I feel like every CJK character is a ligature technically, but no they should just get cached as single characters. I don't think the memory matters in general, and I didn't mean a word limit I meant `if (word.length < 16) cacheWord(); else layItOutNormally()`.

dimensionscape:
That would destroy performance again in a lot of use cases. If you get a mere handful of clusters over 16, you end up back where we started. We need to limit the cluster length and cache everything we can or its back to slow town.



Ok while not really a much of a real world example, it illustrates how text is cached as it's being written. If you type "What", the combinations of "W", "Wh", Wha", "What" are cached. You might look at that and say its wasteful, why dont we just cache the word, this is too much memory, we only need "What". But this would be an incorrect way to look at it. 
A. Many words are common. Once every combination of "What" is cached, now when we begin typing the letters of "What" again, we dont need to get the positions, we just get the cache, making subsequent use of the word what almost 0 cost when appending or inserting. 

B. Many words have common architecture. So now that we typed "What", we now have the leading cache of many other words, like "Whale". Typing "W", "Wh", "Wha" of "Whale" will have little substantive cost.

We need to do this, even for large amounts of break-less text or we kill the rest of the text's performance. However, if we minimize the cluster of text with no breaks (CJK) to be shaped, we can still cache, and in some cases, like repeated letters of the same glyph, can get reused.

This will ensure we keep the performance benefit of caching even for CJK. There may even be a more efficient way to group CJK glyphs together to get the same kind of uniform benefits of reusability within the layout process.

MSGhero:
Can this be merged now? I'd rather people get to use it than wait for incremental upgrades. Future changes would be on another branch I guess

Issue #2336, opened 2020-04-26
Graphics.lineStyle causes extra draw


isohedral:
Hi, I'm using the latest releases of OpenFL (8.9.6) and Lime (7.7.0). When drawing vector paths using Graphics, setting the lineStyle seems to cause an over-eager dot to appear at the current position.  For example:

```haxe
class Main extends Sprite
{
    public function new()
    {
        super();

        var g = graphics;
        g.clear();

        g.lineStyle( 20.0, 0x808080 );
        g.moveTo( 200, 200 );
        g.lineTo( 300, 200 );
        g.lineTo( 300, 300 );
        g.lineTo( 200, 300 );
        g.lineTo( 200, 200 );

        g.lineStyle( 10, 0x000000 );
    }
}
```

The call to `g.lineStyle( 10, 0x000000 )` makes a 10px black dot appear at one corner of the square:

![image](https://user-images.githubusercontent.com/7419269/80312301-80059780-87b2-11ea-897d-d191a12d49c6.png)

Is that the intended behaviour?  I would assume that setting line style is purely for future state and shouldn't cause any drawing to happen.

intoxopox:
I'm hitting this too on OpenFL 9.1.0 and Lime 7.8.0. Has anyone stumbled on a work-around until this is addressed?

joshtynjala:
I am still able to reproduce a similar issue, but it requires using `lineGradientStyle()`.

```haxe
var s = new Sprite();
s.x = 50;
s.y = 50;
s.scaleX = 3.0;
s.scaleY = 3.0;
addChild(s);

var g = s.graphics;
g.clear();
g.lineStyle(20.0);
g.lineGradientStyle(LINEAR, [0x808080, 0x808080], [0.3, 0.8], [0x00, 0xFF]);
g.moveTo(0, 0);
g.lineTo(50, 0);
g.lineTo(50, 50);
g.lineTo(0, 50);
g.lineTo(0, 0);
```

It reproduces on native platforms (cpp, neko, hl), but not on html5.

joshtynjala:
Okay, the fix seems easy enough.

In `CairoGraphics`, it looks like `lineStyle()` calls `closePath(true)` if `stroke && hasStroke` is true:

https://github.com/openfl/openfl/blob/9.2.0/src/openfl/display/_internal/CairoGraphics.hx#L606-L609

[...]

PR #2464, opened 2021-02-06
Fix for lineStyle causes extra draw #2336

intoxopox:
Removes the erroneous extra draw on canvas and cairo.


intoxopox:
I'm preventing an unnecessary call to cairo.lineTo here when positionX/Y & startX/Y are already the same.
Also, passing true to closePath's strokeBefore param, since hasPath, stroke, and hasStroke are all true.

intoxopox:
Any updates on hopefully merging this? It's been very thoroughly battle-tested.

PR #2466, opened 2021-02-15
Implement Shape Caching to improve performance of TextField

dimensionscape:


src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+		for (i in 0...key.length)
+		{
+			chr = key.charCodeAt(i);
+			hash = ((hash << 5) - hash) + chr;
+			hash |= 0;
+		}
+		return hash;
+	}
+
+	public function cache(formatRange:TextFormatRange, getPositions:#if (js && html5) ()->, wordKey:String = null #else TextLayout #end):Dynamic
```
MSGhero: The comma after the `if js` part is causing a compile error

src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+		{
+			return __cacheLongWord(wordKey, formatKey, getPositions);
+		}
+		else {
+			return __cacheShortWord(wordKey, formatKey, getPositions);
+		}
+		
+	}
+
+	private function __cacheShortWord(wordKey:String, formatKey:String, getPositions:#if (js && html5) ()->):Array<Float #else TextLayout):Array<GlyphPosition #end>
```
MSGhero: Compile error with the ) after the `if js` part

src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+		else
+		{
+			var formatMap = new StringMap();
+			formatMap.set(wordKey, #if (js && html5) getPositions() #else getPositions.positions #end);
+			__shortWordMap.set(formatKey, formatMap);
+		}
+		return #if (js && html5) getPositions() #else cast getPositions.positions #end;
+	}
+	
+	private function __cacheLongWord(wordKey:String, formatKey:String, getPositions:#if (js && html5) ()->):Array<Float #else TextLayout):Array<GlyphPosition #end>
```
MSGhero: Another compile error

src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+			}
+			else
+			{
+				formatMap.set(wordKey, #if (js && html5) getPositions() #else getPositions.positions #end);
+			}
+		}
+		else
+		{
+			var formatMap = new StringMap();
+			formatMap.set(wordKey, #if (js && html5) getPositions() #else getPositions.positions #end);
```
MSGhero, line 69: It is possible to store js positions as GlyphPosition objects instead of floats. Doing so here would unify the API everywhere else I imagine

src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+		__shortWordMap = new StringMap();
+		__longWordMap = new StringMap();
+	}
+
+	private static function hashFunction(key:String):Int
+	{
+		var hash = 0, i, chr;
+		for (i in 0...key.length)
+		{
+			chr = key.charCodeAt(i);
```
MSGhero, line 28: Works with Unicode as far as I can tell

src/openfl/text/_internal/ShapeCache.hx
MSGhero: Return type can be more specific than dynamic

src/openfl/text/_internal/ShapeCache.hx
MSGhero: Honestly you can probably just remove the function type and it'll auto type itself

MSGhero: I would be slightly careful of passing a function in. I don't remember if js has issues with closures, but if so that object generation could add up. May not have an effect if you never actually call the function, which is the case if it's in the cache already

src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+			hash = ((hash << 5) - hash) + chr;
+			hash |= 0;
```
MSGhero, line 30: I don't think this line does anything? It has meaning in pure js, but the haxe compiler probably just deletes it unless I'm missing something?

src/openfl/text/_internal/ShapeCache.hx
```diff
@@ -0,0 +1,109 @@
...
+import haxe.ds.StringMap;
+
+#if !openfl_debug
+@:fileXml('tags="haxe,release"')
+@:noDebug
+#end
+
+@:access(openfl.text.TextFormat)
+@SuppressWarnings("checkstyle:FieldDocComment")
+class ShapeCache
```
MSGhero, line 12: Probably appropriate to give some credit to chromium or blink or whoever

MSGhero (changes requested):
JS compile errors need to be fixed.

src/openfl/text/_internal/ShapeCache.hx
MSGhero: You can leave it, it's just annoying to type every time I imagine

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: Oh yeah I messed up the conditional compilation everywhere! I'll fix it right up.

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: There's a few things that would make this more simple. This was the hardest problem I faced nearing completion....how to avoid having a api separation between html5 and native. Instead of doing that to accommodate for the separation of getPositions in TextEngine, I tried to architect it so that we could unify some of the things. GlyphPositions would be more expensive to create, but would make code more readable. I'm more interested in consolidating the getPositions code in TE.

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: I knew I was forgetting some dynamics. Thanks.

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: If I gave credit to every code author I ever learned from after studying their open source code, it would be a pretty long list. 😝 There's nothing in here that warrants accreditation.

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: Compiler cleans it up?

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: Yea, I dont know, the conditional compilation gets messy and confusing sometimes am I right?

dimensionscape: Maybe there is a better way to unify it so I don't end up with separate methods for native and js (or a lot more code that will increase complexity)...I think it might be ok for now. If we could rework getPositions for js && html5, it would eliminate the need.

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: Should be fixed now. Something weird happened when I formatted before pushing.

src/openfl/text/_internal/ShapeCache.hx
dimensionscape: fixed

src/openfl/text/_internal/ShapeCache.hx
MSGhero: I guess. I'll just remove it at some point, as haxe automatically adds that in js export if it's needed

PR #2473, opened 2021-03-14
fix: Shader compiler is giving warnings only when there are errors present

### LINKED ISSUE #2472: Shader compiler is only giving warnings when there are errors present
**Describe the bug**
The [shader compiler in its function `__createGLShader()`](https://github.com/openfl/openfl/blob/775c0ed1a3a1e0c3d146afb4fda3237098bfd7d4/src/openfl/display/Shader.hx#L318)` is using two GL functions to report on errors and warnings during compilation:
- `getShaderParameter(shader, COMPILE_STATUS)`, that returns 0 (GL_FALSE) on errors and 1 (GL_TRUE) on success; if this returns zero, then
- `getShaderInfoLog(shader)` is executed, that returns a string with the message from the compiler.

The problem is: when there are only warnings and no errors, `getShaderParameter(shader, COMPILE_STATUS)` will return GL_TRUE and so the reporting code is not executed.
There is an [identical issue](https://github.com/haxelime/lime/issues/1451) in Lime. Sorry for the duplicated issue, the problem is the same because the code was originally the same.

I recently had a case where an implicit cast, in a shader, would be treated as a warning on my machine (and OpenFL wasn't reporting anything), while the program instantly crashed on another machine that treated it as an error. 

The OpenGL Reference is noting that [we should get the info log even when there aren't errors, if we're debugging a shader](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetShaderInfoLog.xhtml):

> The information log for a shader object is the OpenGL implementer's primary mechanism for conveying information about the compilation process. Therefore, the information log can be helpful to application developers during the development process, even when compilation is successful. 

So, if we want warnings while in debug mode, we should instead check if the `getShaderInfoLog(shader)` is not null, and only then call `getShaderParameter(shader, COMPILE_STATUS)` to separate errors from warnings.

I could be very difficult to debug shaders without this information, as there is too much variance in how opengl compilers are implemented. 

I can make a PR, if needed.

**Expected behavior**
When in debug mode, to show whatever the video driver shader compiler has to comment about my shaders.

giuppe:
Fixes #2472.

The new code uses both `getShaderInfoLog()` and `getShaderParameter()` to recognize the following states:

- no errors, no warnings: (`shaderInfoLog = null`, `compileStatus = 1`) --> show nothing;
- no errors,  warnings present: (`shaderInfoLog != null`, `compileStatus = 1`) --> show "Info compiling shader..." if debug,  else show nothing;
- errors present: (`shaderInfoLog != null`, `compileStatus = 0`) --> show "Error compliling shader..."

### DIFF
--- a/src/openfl/display/Shader.hx
+++ b/src/openfl/display/Shader.hx
@@ -322,13 +322,17 @@ class Shader
 		var shader = gl.createShader(type);
 		gl.shaderSource(shader, source);
 		gl.compileShader(shader);
+		var shaderInfoLog = gl.getShaderInfoLog(shader);
+		var compileStatus = gl.getShaderParameter(shader, gl.COMPILE_STATUS);
 
-		if (gl.getShaderParameter(shader, gl.COMPILE_STATUS) == 0)
+		if (shaderInfoLog != null || compileStatus == 0)
 		{
-			var message = (type == gl.VERTEX_SHADER) ? "Error compiling vertex shader" : "Error compiling fragment shader";
-			message += "\n" + gl.getShaderInfoLog(shader);
+			var message = (compileStatus == 0) ? "Error" : "Info";
+			message += (type == gl.VERTEX_SHADER) ? " compiling vertex shader" : " compiling fragment shader";
+			message += "\n" + shaderInfoLog;
 			message += "\n" + source;
-			Log.error(message);
+			if (compileStatus == 0) Log.error(message);
+			else if (shaderInfoLog != null) Log.debug(message);
 		}
 
 		return shader;


loudoweb:
I confirm that this PR is a plus for windows target. **Tested and approved.**
For html5, I don't see any differences (warnings on cpp are errors on webgl). Anyway there is no regression neither. 
**So this can be merged safely.**

loudoweb:
@giuppe I'm currently testing the Hashlink target and it throws an error :/

I should have test this target in my first comment :S

Can you do something?

```
String.fromUTF8 (c:\HaxeToolkit\haxe\std\hl\_std\String.hx:240)
lime._internal.backend.native.NativeOpenGLRenderContext.getShaderInfoLog (c:\HaxeToolkit\haxe\lib\lime\7,9,0\src\lime\_internal\backend\native\NativeOpenGLRenderContext.hx:2043)
openfl.display.Shader.__createGLShader (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Shader.hx:325)
openfl.display.Shader.__createGLProgram (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Shader.hx:345)
openfl.display.Shader.__initGL (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Shader.hx:509)
openfl.display.Shader.__init (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Shader.hx:458)
openfl.display.OpenGLRenderer.__initShader (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\OpenGLRenderer.hx:538)
openfl.display.OpenGLRenderer.new (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\OpenGLRenderer.hx:166)
openfl.display.Stage.__createRenderer (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Stage.hx:1196)
openfl.display.Stage.__onLimeWindowCreate (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Stage.hx:2191)
openfl.display.Stage.__onLimeCreateWindow (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Stage.hx:1720)
lime.app._Event_lime_ui_Window_Void.dispatch (c:\HaxeToolkit\haxe\lib\lime\7,9,0\src\lime\_internal\macros\EventMacro.hx:91)
openfl.display.Application.createWindow (c:\HaxeToolkit\haxe\lib\openfl\git\src\openfl\display\Application.hx:90)
ApplicationMain.create (i:\WORK\Workspace\OPEN SOURCE\kadabra\bin\hl\haxe\ApplicationMain.hx:99)
ApplicationMain.main (i:\WORK\Workspace\OPEN SOURCE\kadabra\bin\hl\haxe\ApplicationMain.hx:25)
```

loudoweb:
A fast way to fix the issue is this:
`var shaderInfoLog = #if hl null #else gl.getShaderInfoLog(shader) #end;`
but it means that you don't have warning if you don't have errors like before.

giuppe:
It seems a bug in Lime: the function `lime_get_shader_info_log()` seems not implemented for the hl target and returns null; but then the crash seems related to the function `getShaderInfoLog()` [trying to convert a null into a String](https://github.com/haxelime/lime/blob/f110d52e5a17e17422183c7dd90e9dfdc1f85782/src/lime/_internal/backend/native/NativeOpenGLRenderContext.hx#L2043). 

This should be fixed in Lime, checking for the null result in `getShaderInfoLog()`, otherwise the problem would still be present when someone uses Lime without OpenFL; and also we would need to remember to revert the `#if hl` conditional in OpenFL when the function will be implemented in Lime...

Edit: I opened a [pull request on Lime](https://github.com/haxelime/lime/pull/1458) about the issue.

loudoweb:
Thank you!

PR #2489, opened 2021-04-15
Enumerate device fonts

loudoweb:
`Font.enumerateFonts(true) `should enumerate device fonts (like in C:\Windows\Fonts). It outputs only TTF fonts.
Windows, neko and Hashlink work! I have not tested Mac and Linux yet.

There is room for improvement for the future, for e.g: cache all the fonts for next time enumerateFonts(true) is called.


loudoweb:
It should be easy to review, test and merge this. How can I help?
I'm using this in my editor to make all fonts available to user like in Animate, Photoshop, etc.
 @Dimensionscape @joshtynjala

joshtynjala:
Looks good to me.

PR #2491, opened 2021-04-30
add totalMemory to hl as other targets

loudoweb:
`System.totalMemory` returns  **0** for hash link so I've added the ability to HL to return a value.


joshtynjala:
Closing because merged #2561 instead.

PR #2496, opened 2021-05-26
Add dirty check for Tilemap, improve the performance of Tilemap

rainyt:
If there is no change in the Tilemap, the old vertexBufferData data will be used.


rainyt:
I found that this RP has rendering problems, such as incorrect transparency.

PR #2497, opened 2021-05-27
Improve the performance of Tilemap

rainyt:
My idea is that if there is no change in the tilemap, there is no need to refresh the Buffer data. I need to know whether __worldAlpha has been changed, so I need to use __worldAlphaChanged, but I found that it is only supported in Dom, so I will provide it to the tilemap in a unified implementation, if it is not needed, do I need to only provide it to the tilemap? ?


rainyt:
It disrupts some things and no longer applies.

PR #2504, opened 2021-09-08
pass supplied width/height args on html5

Geokureli:
In html5, when instantiating bitmap classes defined using `@:bitmap` the width and height are 0x0 until the data is fetched asynchronously.  this is true even when a width and height are passed into the constructor. This change allows devs to pass in default dimension args if the code needs the image to have a non-zero size, immediately after construction.

Related to https://github.com/HaxeFlixel/flixel/pull/2402

### DIFF
--- a/src/openfl/utils/_internal/AssetsMacro.hx
+++ b/src/openfl/utils/_internal/AssetsMacro.hx
@@ -46,7 +46,7 @@ class AssetsMacro
 			var constructor = macro
 				{
 					#if html5
-					super(0, 0, transparent, fillRGBA);
+					super(width, height, transparent, fillRGBA);
 
 					if (preload != null)
 					{


dimensionscape:
What negative impact would this change cause, if any? @jgranick ?

dimensionscape:
I noticed that code on line 71, for non-html targets allowed this as early 4.8.1 and was changed to 0,0 at some point thereafter as well (possibly for parity in behavior?)

jgranick:
It would have the following drawbacks:

 1. An additional `ImageBuffer` the size of the `BitmapData` would be allocated immediately, then discarded upon load
 2. A `Bitmap`, `Tilemap` or other object that uses the `BitmapData` might not break its cache to force rendering (would require testing)
 3. It would be impossible to know when the image is loaded (for example, when using `bitmapData.draw`, or `bitmapData.getPixel`, etc)
 4. It would diverge from the behavior of Flash Player (which ignores the width/height values in the constructor when using the Haxe `@:bitmap` embed)

Arguably `@:bitmap` on HTML5 is already less-than-ideal. One approach would be to read the size of the image within the embed macro at compile-time, and hard-code the size into a private field. Then it would have an initial size, and would not require the user to specify the size, though it would still suffer from problems 1, 2 and 3.

dimensionscape:
Thank you, should we consider this closed?

jgranick:
I'm open to ideas. 

I suppose `@:bitmap` on HTML5 is generally discouraged now due to its behavior. Improving it may help make it more livable, though it may also lead to more people trusting it and being let down when using `bitmapData.getPixel32` and getting an unexpected value, etc.

I'm in favor mostly of consistency, though I'm a little unsure here as synchronous loads are not possible on HTML5

dimensionscape:
I personally can't think of anything that doesn't seem janky. The extra buffer allocations and the inability to know when the image is done loading kind of feels like a deal breaker imo. However, I can really see conflicts arising from trying to get measurements too soon. I suppose the best practice would be to wait. This is sort of a difficult issue considering we can only do asynchronous loads. 

I was thinking we could allow the values to be set without actually doing anything if the image is not loaded and facilitate another mechanism to check if it has been loaded aside from waiting for the height and width to have a value other than 0. This would prevent the allocation and satisfy the need to know when it is loaded, but I'm not sure if that seems appropriate.

@Geokureli do you have any other ideas?

Geokureli:
is there any way to give the bitmapData a width and height without creating an image buffer? Could we pass 0x0 into the constructor and just set the dimensions afterward?

does it matter that this is only the case when parameters are passed into a constructor that normally doesn't need any arguments? In my particular case the images are small and it's worth the extra memory compared to listening for onLoad, however if the added memory isn't an obvious consequence it could be a bigger deal to some unsuspecting people.

Edit: I'm also unclear on why there needs to be any asnyc load for html5, I thought these `@:bitmap` classes were embedded and therefore available after the preload just like `Asset.getBitmapData`

flashultra:
I'm not sure is this will be a good solution, but you can have a embedded transparent pixel ( 1x1px) and load first that pixel and setting width height . , after that for loadBitmapData  listen for onComplete event and replace the transparent pixel with the new image. Something like that, not sure how it will works.

Geokureli:
the workaround I used to overcome this in flixel was:
```hx
#if html5
@:keep @:bitmap("assets/images/tile/autotiles.png")
private class RawGraphicAuto extends BitmapData {}
class GraphicAuto extends RawGraphicAuto
{
	public function new (width = 128, height = 8, transparent = true, fillRGBA = 0xFFffffff, ?onLoad:Dynamic)
	{
		super(width, height, transparent, fillRGBA, onLoad);
		// Set properties because `@:bitmap` constructors ignore width/height
		this.width = width;
		this.height = height;
	}
}
#else
@:keep @:bitmap("assets/images/tile/autotiles.png")
class GraphicAuto extends BitmapData {}
#end
```
We only needed width/height, not any methods like `draw` or `getPixel`.

It's also worth noting that html5 image classes using `@:bitmap` have an onLoad callback, contrary to Joshuas 3rd point
> It would be impossible to know when the image is loaded

also on that point, even if there are people checking width/height to determine if bitmaps are loaded, how many of them are passing in width/height values to the constructor?

jgranick:
Generally, the renderer and the `lime.graphics.Image` class expect there to be an HTML5 `Image` or `Canvas` in addition to an `ImageBuffer`. It is only possible to create an `Image` asynchronously in the browser nowadays. Creating an `ImageBuffer` synchronously is possible, though this would not work on the canvas or DOM renderers, as well as `bitmapData.draw`.

I am not sure if setting the `width` and `height` without the image size matching will cause any crashes or problems or not. Historically I think it did, though it may be resolved now, though I could still see (potential?) cache misses.

So perhaps we have a few options:

 1. Leave it as-is
 2. Use the macro to read the image size, and hard-code it in, and fix any bugs resulting from width and height being set with no actual image data, though currently width/height of zero is a sign internally that it is not ready for rendering or most image get/set functions
 3. Use the macro to embed the image data, and try and see if we can have an `ImageBuffer` without an `Image` on HTML5, which could potentially work for OpenGL at least
 4. Use some broader workaround where we try and push the image into a queue to be loaded by a preloader so that it is available in the Assets.cache on instantiation. This would (in some ways) be consistent with how we treat other resources on HTML5, though this could cause some problems with inversion of control, where currently these classes have no concept of pushing things into the preload queue, which is handled right now by the command-line tools

Geokureli:
> though currently width/height of zero is a sign internally that it is not ready for rendering or most image get/set functions

do you have any examples this in the library?

I'm a big fan of 4. I think this would solve most people's issues and anyone using @:bitmap on preload screens is already aware of these caveats. Also, this is how I originally expected it to behave

personally, I think 2, 3 and 4 would all be great additions, and we should force people to use some other way to determine whether images are loaded. Of course I don't know what that all entails, how much work it is and who will have to deal with breaking changes

PR #2508, opened 2021-10-09
Fix incorrect macOS font encoding

DigiEggz:
macOS doesn't render extra characters correctly when set to UTF8, instead requiring UTF16. See more about the issue here: https://community.openfl.org/t/font-problem-on-mac-os-x/12919/7

Original fix found by Beckwahn


DigiEggz:
@jgranick This fix has been tested on multiple versions of macOS.

joshtynjala:
f8301ceacb72180e1faf0f9b743416b097871f31

Issue #2505, opened 2021-09-13
Arabic  ligatures not showing correctly.


Shallowmallow:
**Describe the bug**

The arabic ligature don't show correctly.

**To Reproduce**

```
var format = new TextFormat (openfl.Assets.getFont("assets/rudawregular2.ttf").name, 30, 0x7A0026);
		var textField = new TextField ();
		
		textField.defaultTextFormat = format;
		textField.embedFonts = true;
		textField.selectable = false;
		textField.x = 50;
		textField.y = 50;
		textField.width = 500;
		textField.text = "صباح الخير";
		var textLayout = new openfl.text._internal.TextLayout();
textLayout.direction = RIGHT_TO_LEFT;
textLayout.script = ARABIC;
textLayout.language = "ar";
Reflect.setField(Reflect.field(textField, "__textEngine"), "__textLayout", textLayout);

```

( the text don't appear exactly the same, as the words are in the opposite order). But  I did some correction in openfl  to correct this ( which I'll submit once the ligature problem is corrected )
**Expected behavior**

The arabic text  should not be disjointed

**Screenshots**

On html5

![Sélection_035](https://user-images.githubusercontent.com/12767394/133151129-1fb61f0b-9c2a-4694-adbc-f7f3e0c4c68c.png)

On openfl linux

![Sélection_036](https://user-images.githubusercontent.com/12767394/133151166-11cbdadd-b5b3-4d44-948a-99e4e7d10168.png)

**OpenFL Targets**

openfl linux

**Additional context**

There are few posts talking about this problem

https://github.com/openfl/openfl/issues/1853

https://github.com/openfl/openfl/issues/1327
 
But at the same time in this issue
https://github.com/openfl/openfl/issues/1905

you can see that there are ligatures.

So is there a bug or am I doing something wrong ?

Shallowmallow:
When I try different cluster levels with harfbuzz 

https://harfbuzz.github.io/working-with-harfbuzz-clusters.html

It doesn't change anything, combining letters don't change cluster value as they should .

Shallowmallow:
Resolved !
Basically
https://github.com/openfl/openfl/blob/0c48c5428f6cf7f484d53b62bd1be0d75adf61dd/src/openfl/text/_internal/TextLayout.hx#L481-L489

it doesn't send the script. 
I think you just need to replace the code by  
[...]

PR #2517, opened 2021-12-06
Basic rTL and arabic ligatures

Shallowmallow:
https://github.com/openfl/openfl/issues/2505

So here are some basic rTL  for the textField .How does it work ? ( I wanted to follow flash way of doin it, but flash doesn't work on my computer, and it seems people needed to use some special libraries)

I have added the rTL,language and script attribute to the textAttribute.
Which mean you can some normal text written left to right.
```
	var textNormal = new TextField ();
		textNormal.x = 50;
		textNormal.y = 150;
		textNormal.embedFonts = true;
		textNormal.width = 500;
		textNormal.selectable = true;
		textNormal.text = "1234 567 12";
		textNormal.background = true;
		textNormal.rTL = true;
textNormal.backgroundColor = 0x00FF00;
textNormal.type = openfl.text.TextFieldType.INPUT;
```

You can also have some arabic, for example

```
		var format = new TextFormat (openfl.Assets.getFont("assets/NotoArabicUI.ttf").name, 30, 0x7A0026);
		textField = new TextField ();

textField.script = ARABIC;
textField.language = "ar";
		
		textField.defaultTextFormat = format;
		textField.embedFonts = true;
		textField.x = 50;
		textField.y = 50;
		textField.width = 500;
		textField.wordWrap = false;
		//textField.autoSize = openfl.text.TextFieldAutoSize.RIGHT;
textField.rTL = true;
```

I haven't tested on HTML5
What works:
  - ligatures
  - showing textfields left to right
  - selections

What still needs to be done
  - paragraphs
  - cutting long words
  - accepting some special unicode character which means lefttorigt or righttoleft ( so that you can easily have some arabic with numbers written ltr)
  
Noticed bugs:
- when you click  on a  character,, the caret goes to the right,  it shouldn't be the case.

### DIFF
--- a/src/openfl/display/_internal/CairoTextField.hx
+++ b/src/openfl/display/_internal/CairoTextField.hx
@@ -279,7 +279,15 @@ class CairoTextField
 					for (position in group.positions)
 					{
 						if (position == null || position.glyph == 0) continue;
-						glyphs.push(new CairoGlyph(position.glyph, x + position.offset.x + 0.5, y - position.offset.y + 0.5));
+						if (textEngine.rTL)
+						{
+							glyphs.push(new CairoGlyph(position.glyph, x - position.offset.x + 0.5, y - position.offset.y + 0.5));
+						}
+						else
+						{
+							glyphs.push(new CairoGlyph(position.glyph, x + position.offset.x + 0.5, y - position.offset.y + 0.5));
+						}
+
 						x += position.advance.x;
 						y -= position.advance.y;
 					}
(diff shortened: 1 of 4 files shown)


ShaharMS:
I was brought here by the openFL discord server, and I have something to say, and a question.

I'm working on a haxelib that is mentioned to add support for flixel&openfl text input. I made a lot of progress on flixel's FlxInputText (actually, pretty much done), but didnt really know where to start with openFL.

Can I use that code (from the PR) to implement RTL in openFL's textField and credit you?

This can really help me, not only in the implementation side but also in the learning side.

About credits, I'll credit you in the main contributors on github&haxelib and on the actual TextField class and it's inheriting classes (meaning that you will be credited on classes extending your TextField implementation). 

Sorry if there's bad English, and thanks in advance :)

If you want to see the lib I'm working on, its called `texter` and is on my profile (don't think i can link my repository here, also feels kinda bad, like doing self promo)

Shallowmallow:
Hi Shahar !

Of course you use the code :)

If you see other bugs, you can tell me  and maybe we can improvve the PR together :)

PR #2527, opened 2022-01-19
Fix TextField behaviour wrt stopPropagation()

dpomier:
TextFields should listen to stage events at the capture phase to ensure correct behaviour of text inputs when user code calls stopPropagation() in the event's target phase.

This PR is ready to be merged if this makes sense to you. It fixes the common situation where the parent of a text input stops the propagation of mouse up events, leading the text to be continuously in "selection mode", whereas in Flash it would still catch mouse inputs.

### DIFF
--- a/src/openfl/text/TextField.hx
+++ b/src/openfl/text/TextField.hx
@@ -3123,9 +3123,9 @@ class TextField extends InteractiveObject
 	{
 		if (stage == null) return;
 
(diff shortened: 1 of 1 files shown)


dpomier:
@MSGhero could you have a look at this PR? This change makes the behaviour of text fields closer to what we have in flash.

PR #2528, opened 2022-01-19
adding the CapLockcheck to work when the capLock works

Cheemsandfriends:
When I was looking into keyboard stuff cos I had problems with adding fancy stuff like a ~ on top of a N or vowels with accents and stuff like that.

When I discovered that the CapLock thing doesn't work when inputting text in-game.

This makes a small fix.

(Small feature but it inputs well when inspecting and also good when adding textboxes lol)


Cheemsandfriends:
I'll set this to draft because, it works but it has some issues like dots and stuff like that

Cheemsandfriends:
it works good now with the cap locked thing correctly

PR #2529, opened 2022-01-24
Haxe 3 support

player-03:
These changes are required to compile DisplayingABitmap in Haxe 3.4.7.

At some point - not necessarily now - we could consider dropping Haxe 3 support, allowing for cleaner code. Looking through the forums, it seems there were [only three references to "Haxe 3" within 2021](https://community.openfl.org/search?q=%22haxe%203%22%20order%3Alatest). Not a whole lot, though admittedly still greater than zero.

Note: the latest released version of OpenFL works in Haxe 3, because everything broken was added more recently than that.

### DIFF
--- a/src/openfl/utils/Object.hx
+++ b/src/openfl/utils/Object.hx
@@ -5,6 +5,7 @@ import openfl.display.DisplayObjectContainer;
 
 @:transitive
 @:callable
+#if !haxe4 @:forward #end
 abstract Object(ObjectType) from ObjectType from Dynamic to Dynamic
 {
 	public inline function new()


dimensionscape:
Did changes in Object break dynamic access for haxe 3?

I think we should continue to support Haxe 3 until Haxe 5 is released, but that is just my opinion.

I'm opposed to merging the latter part of these fixes if it's due to an issue with Object. We should fix Object instead.

player-03:
I'm pretty sure it simply never worked in Haxe 3. In Haxe 3, the accepted way of supporting dot access was `implements Dynamic`, which isn't allowed on abstracts. I know it says `to Dynamic`, but that doesn't actually help with dot access.

We could coerce it to `Dynamic`, I guess. It'd make the code look cleaner, but we'd lose `Object`'s built-in null checks. So we'd probably want to coerce it only in Haxe 3, which makes the code messier again.

dimensionscape:
Thats pretty fair, but also kind of the norm in openfl. Maybe its time to have a discussion about ending support for haxe 3. All the same, I feel like we should either support it without half measures or not at all.

player-03:
In late 2019, Joshua dropped Haxe 3 support (73a0c5a) and then restored it (b8a176c). I'm guessing due to complaints, but he didn't leave an explanation. Looking at the forums, there haven't been as many mentions of Haxe 3 lately, so it really could be time to start discussing it again. I'll post this on Discord.

---

To expand a bit on the feature having never worked, let's go over the timeline.

- Dynamic field access was added last March (cf53228), so it clearly wasn't used before then. (Whether or not there's a way it could have worked in Haxe 3, it would not have worked in Haxe 4.) A few days later, Joshua realized this broke Haxe 3 builds, and simply disabled the setter (d645b34).
- In April, the most recent version of OpenFL was released.
- In October, StyleSheet.hx was added (a743830). I haven't done a thorough search, but I believe this was the first time OpenFL used `Object`'s dot access setter.

Up until now, this hasn't been a problem because `Object` simply didn't have this feature. You had to type out `__set()` or coerce to `Dynamic` or otherwise work around it. But now the feature exists, and OpenFL has finally gotten around to using it, and it'll break Haxe 3 builds if released.

Hope that paints a clearer picture.

joshtynjala:
Okay, I added Haxe 3.4.7 to the matrix for unit tests. It is now failing, obviously.

I committed the `new Map()` fix. I'll let someone else figure out if this is the right solution for `StyleSheet`.

PR #2531, opened 2022-01-29
Support GLSL version directive in uploadSources

6J7KZg2f:
Uploading GLSL shader sources to a Program3D via uploadSources currently fails if there is a version directive in the shader, as the function adds a prefix which violates the version-must-come-first rule.
This checks for a version directive in the first line, and places the prefix between it and the rest of the shader source if there is.

### DIFF
--- a/src/openfl/display3D/Program3D.hx
+++ b/src/openfl/display3D/Program3D.hx
@@ -430,8 +430,28 @@ import lime.utils.BytePointer;
 			#endif
 			";
 
-		var vertex = prefix + vertexSource;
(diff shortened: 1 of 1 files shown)


loudoweb:
Can you share a sample to test your PR please.

6J7KZg2f:
> Can you share a sample to test your PR please.

@loudoweb 
Sure, please see this gist:
https://gist.github.com/6J7KZg2f/a0277e992dbf0e3e2e0ccff1d12c6790

PR #2533, opened 2022-01-30
Implement setProgramConstant byteArray & vector for GLSL shaders

6J7KZg2f:
Currently, these functions for setting shader uniforms silently discard input when the shader program format is GLSL.

### DIFF
--- a/src/openfl/display3D/Context3D.hx
+++ b/src/openfl/display3D/Context3D.hx
@@ -1,5 +1,6 @@
 package openfl.display3D;
 
+import openfl.errors.RangeError;
 #if !flash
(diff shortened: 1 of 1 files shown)


6J7KZg2f:
Note: I don't know if that's how the ByteArray one should work, but I can't see an analogue for that kind of thing in GLSL.

ghost:
I was wondering if you can put this pull request into your respiratory just incase they don't accept

ghost:
and also how do you find out which line of shader code is causing the errors cause I work for fnf stuff

ghost:
and sometimes the shaders break @6J7KZg2f

PR #2538, opened 2022-03-17
Fix TextEngine segfault in HashLink.

player-03:
HashLink crashes if you try to access a negative array index, while other platforms are fine.

You can easily reproduce this using FeathersUI:

```haxe
var textArea = new feathers.controls.TextArea();
textArea.width = 50;
textArea.text = "Enough for word wrap";
textArea.validateNow();
```


joshtynjala:
I took out the conditional compilation, since it doesn't seem to be necessary, and committed it.

player-03:
Yeah, that's fine too. I was going for minimum changes, but in retrospect, code readability matters more.

PR #2543, opened 2022-03-29
Lazily initialize static bitmaps.

player-03:
The web workers coming up in Lime 8.2.0 create a new static context on the background thread, initializing all static variables. Problem is, web workers aren't allowed to do I/O, so creating a bitmap causes an error. One solution is to lazily initialize these particular bitmaps.

All of OpenFL's other static variables work fine, as far as I'm aware.


player-03:
I took another look and found a better place to handle this: openfl/lime@166ac06e3e97b3cb3b695452e935fd9c74cd92fb.

PR #2550, opened 2022-04-29
TextField: Don't call set selection every update

junsred:


junsred:
Because of https://github.com/openfl/openfl/commit/9482d4896b882e54136c42c619cd6b3043d2d0dd `__caretIndex` is set to 0 every time text updates. This causes scroll setting to be unavailable because every time scroll is set `__updateLayout` is called and  when `setSelection` is called it scrolls to selected text but because `__caretIndex` is 0 scroll is always 1.

junsred:
@Dimensionscape

junsred:
Hmm this solves the not being able to set scroll issue but I'm not sure about when to scroll to `__caretIndex `.  
Thoughts? @player-03

junsred:
@joshtynjala

dimensionscape:
> Because of [9482d48](https://github.com/openfl/openfl/commit/9482d4896b882e54136c42c619cd6b3043d2d0dd) `__caretIndex` is set to 0 every time text updates. This causes scroll setting to be unavailable because every time scroll is set `__updateLayout` is called and when `setSelection` is called it scrolls to selected text but because `__caretIndex` is 0 scroll is always 1.

My commit only affects the htmlText setter, so I'm not so sure about the characterization there. In any case, Ill take a look at this over the weekend after I refresh my memory a bit.

dimensionscape:
Ill check again just to be sure, but I'm pretty sure set scrollH and scrollV work fine.

junsred:
@Dimensionscape https://github.com/openfl/openfl/commit/b08f9691fab94be2bc20749b6f745843020ebceb this commit is the one to blame. Your commit wasn't a mistake but it exposed one.

barisyild:
I cannot see any difference in functionality.
Is there an example where I can reproduce the difference?

junsred:
> I cannot see any difference in functionality. Is there an example where I can reproduce the difference?

@barisyild 
Create a text field with multiple lines. Add 1 more line. Set `scrollV` to the end of the line.  `scrollv` won't change.

junsred:
https://github.com/openfl/openfl/commit/b08f9691fab94be2bc20749b6f745843020ebceb makes it that every time `__updateLayout` is called it scrolls to the `__caretIndex` and because setting scroll calls `__updateLayout` updated scroll updates back to `__caretIndex`'s line which would be first line because of https://github.com/openfl/openfl/commit/9482d4896b882e54136c42c619cd6b3043d2d0dd.

My pull request solves this problem by removing the scrolling to `__caretIndex` every update.

dimensionscape:
Ok, thanks for the PR! I need to really take a look at some things before merging. TextField is really fragile 😔.

joshtynjala:
I suspect that this pull request can probably be closed _without_ merging now (or perhaps with some tweaks). While working on something related to Feathers UI, I discovered that Flash actually tries to keep the exact same selection as the previous text when `text` or `htmlText` changes, as long as the selection fits within the new length. So instead of setting `__caretIndex` and `__selectionIndex` to `0` or `length`, as we did previously, they now get clamped within the new length.

Relevant commits that make OpenFL work more like Flash:

c4bfca735328e02f722bb72caba160bd5bbadad8
86cf3f1882c9f30a8e4dc410d79be39c58881fe9

@junsred Can you try the latest code on the develop branch and let me know if you still see issues? Thanks!

junsred:
@joshtynjala  Thanks for the update I'll try to test it as soon as possible. But I still think set scrollV will be unusable since what ever it is set to it will go back to `__caretIndex`.

junsred:
It appears to be fixed! I'm closing this for now.

PR #2551, opened 2022-04-29
Arabic ligature + Better RTL

Shallowmallow:
Here's a pull request to start a discussion on rtl/ltr . There still are a few things I would like to correct before an openfl release.

### Improvements

- **Arabic ligatures**
- **rTL alignment**

-  **Automatic rTL detection**
_The automatic detection is done by harfbuzz. It detects from the first few characters_
_For the detection to work, the script and direction must be invalid ones_
_I've enabled by default, by setting the language, the script, and direction to invalid ones_
_I created a new text script GUESS which is in fact just the harfbuzz invalid one_ 

- **I've expose language, script and text direction  to the textField**
_I've done this before coding the detection part. I'm not sure it has so many uses now the detection is coded_
_TextDirection uses openfl.text.\_internal.TextLayout.TextDirection. If we keep exposing the textdirection in textField, maybe it's better that TextDIrection has own hx file not in \_internal_

- **You can use TextFormat to control rtl /ltr**
- **You can use htmlText and use TextFormat tags to control rtl/ltr**
_For example you can do `textfield.htmlText = "1234 <textformat dir='rtl'> 567 12 34</textformat>"`_

### Weird Bugs
- `textfield.htmlText = "12345<textformat dir='rtl'>12345</textformat>"`
_If you have the exact same word in rtl and ltr problem, it's because of caches. It's not really important IRL, because you usually use  different scripts in rtl and ltr_
- when you do a selection on a word where there are characters that are not in the font, it crashes
_It may be just an openfl bug and not this patch bug,  just noticed it here_

### What is left to do
- more testing (long words, etc ., bidirectional)
- make selection and interactivity work better on bidirectional textfield
- make  html5 working
_ I've focused more on harfbuzz.  Alignment should work on html5, but selection, etc  still doesn't work _

### What I don't plan to do
- I don't plan to implement Bidi characters, now that the textformat tag is working. Because it's maybe better to implement an external library Fribidi etc .

### DIFF
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -516,24 +516,24 @@ import js.html.CanvasRenderingContext2D;
 		iy1 = anchorY;
 		iy2 = anchorY;
 
-		if (!(((controlY1 < anchorY && controlY1 > __positionX) || (controlY1 > anchorY && controlY1 < __positionX))
-			&& ((controlY2 < anchorY && controlY2 > __positionX) || (controlY2 > anchorY && controlY2 < __positionX))))
+		if (!(((controlY1 < anchorY && controlY1 > __positionY) || (controlY1 > anchorY && controlY1 < __positionY))
+			&& ((controlY2 < anchorY && controlY2 > __positionY) || (controlY2 > anchorY && controlY2 < __positionY))))
 		{
-			var u = (2 * __positionX - 4 * controlY1 + 2 * controlY2);
-			var v = (controlY1 - __positionX);
-			var w = (-__positionX + 3 * controlY1 + anchorY - 3 * controlY2);
+			var u = (2 * __positionY - 4 * controlY1 + 2 * controlY2);
+			var v = (controlY1 - __positionY);
+			var w = (-__positionY + 3 * controlY1 + anchorY - 3 * controlY2);
 
 			var t1 = (-u + Math.sqrt(u * u - 4 * v * w)) / (2 * w);
 			var t2 = (-u - Math.sqrt(u * u - 4 * v * w)) / (2 * w);
 
 			if (t1 > 0 && t1 < 1)
 			{
-				iy1 = __calculateBezierCubicPoint(t1, __positionX, controlY1, controlY2, anchorY);
+				iy1 = __calculateBezierCubicPoint(t1, __positionY, controlY1, controlY2, anchorY);
 			}
 
 			if (t2 > 0 && t2 < 1)
 			{
-				iy2 = __calculateBezierCubicPoint(t2, __positionX, controlY1, controlY2, anchorY);
+				iy2 = __calculateBezierCubicPoint(t2, __positionY, controlY1, controlY2, anchorY);
 			}
 		}
 


piratejet:
if I try to select the text with the cursor or press ctrl + a to select them all - it crashes

```
Called from openfl.display._internal.Context3DDisplayObjectContainer::renderDrawable openfl/display/_internal/Context3DDisplayObjectContainer.hx line 30
Called from openfl.display.OpenGLRenderer::__renderDrawable openfl/display/OpenGLRenderer.hx line 858
Called from openfl.display._internal.Context3DTextField::renderDrawable openfl/display/_internal/Context3DTextField.hx line 39
Called from openfl.display._internal.Context3DTextField::render openfl/display/_internal/Context3DTextField.hx line 24
Called from openfl.display._internal.CairoTextField::render openfl/display/_internal/CairoTextField.hx line 410
Called from lime.graphics.cairo.Cairo::showGlyphs lime/graphics/cairo/Cairo.hx line 370
Error : Invalid object
```

piratejet:
if you start typing with Latin letters and then type Arabic it doesn't work as it should.

and also if you start typing with Arabic letters and then type Latin the behavior is also not correct.

piratejet:
when i try to build html5 it shows this error

```
D:/work/haxe/root/_libraries/openfl/src/openfl/text/TextField.hx:1820: characters 29-40 : openfl.text._internal.TextLayoutGroup has no field glyphsOrder
D:/work/haxe/root/_libraries/openfl/src/openfl/text/TextField.hx:961: characters 29-40 : openfl.text._internal.TextLayoutGroup has no field glyphsOrder
D:/work/haxe/root/_libraries/openfl/src/openfl/text/TextField.hx:961: characters 23-40 : You can't iterate on a Dynamic value, please specify KeyValueIterator or KeyValueIterable
```


there is no glyphsOrder for html5
```
#if (js && html5)
public var glyphs:Array<Float>; // TODO: Make consistent with native?
#else
public var glyphs:Array<GlyphPosition>;
public var glyphsOrder:Array<Int>;
#end
```

Shallowmallow:
Hi ! I'm quite bad at git. But can you try without the latest commit "1st
draft addPositions which enable to get char index under point
<https://github.com/openfl/openfl/pull/2551/commits/46e309ee877b9ea8d558728b6ffa736337cc1976>"
. This is te onethat is causing problems I think for html5 ( and maybe the
other error too, but I'm not sure) I'll look at  in te next few days.

Le mar. 12 juil. 2022 à 11:49, piratejet ***@***.***> a
écrit :

ShaharMS:
Us there a reason to not keep the current html5 implementation?

It seems as if the html5 breaking change at line 961 is easily avoidable by using conditional compilation:
```haxe
#if (js && html5)
return group.startIndex + i;
#else
for (k => g in group.glyphsOrder)
{
	if (g == i) return group.startIndex + k;
}
return -1;  // Shouldn't happen
#end
```

There is another occurrence of this type of breaking change in line 1820: you access glyphesOrder in all platforms, while its defined only on sys targets. You can wrap the usage of glyphsOrder in a conditional block, while keeping the old implementation for the web.

Shallowmallow:
Check there instead https://github.com/openfl/openfl/pull/2614

PR #2552, opened 2022-04-29
TextEngine: better restrict

junsred:
In flash every character is excluded after '^'.


junsred:
In flash every character is excluded after ^.

joshtynjala:
Yeah, in Flash, more than one character or range can appear after `^`. We allowed only one, for some reason.

PR #2555, opened 2022-05-04
Lowercase style names

junsred:
Style names are not case-sensitive.


joshtynjala:
Confirmed that style names are not case-sensitive in Flash. There are a few more places where toLowerCase() is needed, but I'll take care of that.

barisyild:
I didn't detect any functional difference, it's perfect 👍

PR #2559, opened 2022-05-05
TextField: add max and min restrictions to set_scrollV too

junsred:
add restrictions to scrollV too

### DIFF
--- a/src/openfl/text/TextField.hx
+++ b/src/openfl/text/TextField.hx
@@ -2815,7 +2815,7 @@ class TextField extends InteractiveObject
 		if (value > __textEngine.maxScrollV) value = __textEngine.maxScrollV;
 		if (value < 1) value = 1;
 
-		if (value > 0 && value != __textEngine.scrollV || __textEngine.scrollV == 0)
+		if (value != __textEngine.scrollV || __textEngine.scrollV == 0)
 		{
 			__dirty = true;
 			__setRenderDirty();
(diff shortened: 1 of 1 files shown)


src/openfl/text/TextField.hx
```diff
@@ -2812,6 +2812,9 @@ class TextField extends InteractiveObject
 	{
 		__updateLayout();
 
+		if (value > __textEngine.maxScrollV) value = __textEngine.maxScrollV;
+		if (value < 1) value = 1;
```
dimensionscape: I think this(L#2816) might be handled somewhere else. I can't set scrollV to less than 1 in the latest git.

junsred: Yeah nothing happens in latest git. What needs to happen is it should be set to 1.

dimensionscape: Ah yeah that makes sense. Cool.

dimensionscape (changes requested):
Ah it looks like this is partially handled in TextEngine but the `value>0` check prevents it. Just revert the initial commit and we're good.

junsred:
@Dimensionscape after reverting my first commit, when i set the scrollV to a number that is higher than max scrollV it becomes max - 1 because of https://github.com/openfl/openfl/commit/f4fcd02408bfd4f0a9c2998bf507feadaddf1572 (Line: 1902) commit.

junsred:
I don't see my first commit doing any harm.

junsred:
Oh, I just realized that https://github.com/openfl/openfl/commit/f4fcd02408bfd4f0a9c2998bf507feadaddf1572 is fixed and the code is fine without first commit. I reverted it again. It should be ready for merge.

PR #2560, opened 2022-05-09
DOM: Improve DOM text renderer.

rainyt:
1. Fix the problem that tap focus cannot be input.

2. When repairing the input, the default color is incorrect.

3. When clicking a text box other than text, the cursor positioning is wrong.

4. Support for password segment is added, but this requires a TTF, which I haven't handled yet.

5. Fix the problem of IME input.

6. Fix the problem that textfield did not get the text update in time.

### DIFF
--- a/src/openfl/display/_internal/DOMTextField.hx
+++ b/src/openfl/display/_internal/DOMTextField.hx
@@ -1,5 +1,7 @@
 package openfl.display._internal;
 
+import haxe.Timer;
+import openfl.events.Event;
 import openfl.text._internal.HTMLParser;
 import openfl.text._internal.TextEngine;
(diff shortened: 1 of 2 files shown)


rainyt:
Password is not perfect, but others are good.

rainyt:
For the rendering of password, I want to use the input tag. https://github.com/openfl/openfl/pull/2468

ShaharMS:
Any reason why this isnt merged?

rainyt:
Now it seems that it still has some defects, under certain circumstances. The input cursor sometimes returns to the first position.

rainyt:
I can no longer guarantee that it will function properly

PR #2561, opened 2022-05-11
Adding support in System for hl

Cheemsandfriends:
While I was going through System, I noticed that you couldn't check the totalMemory, so I was searching and I found out that it is totally possible to check the memory, (I'm not sure about gc, but it seems like the same)

### DIFF
--- a/src/openfl/system/System.hx
+++ b/src/openfl/system/System.hx
@@ -267,7 +267,7 @@ import hl.Gc;
 		return
 			untyped #if haxe4 js.Syntax.code #else __js__ #end ("(window.performance && window.performance.memory) ? window.performance.memory.usedJSHeapSize : 0");
 		#elseif hl
-		return untyped Gc.stats().currentMemory;
+		return Std.int(Gc.stats().currentMemory);
 		#else
 		return 0;
 		#end
--- a/src/openfl/system/System.hx
+++ b/src/openfl/system/System.hx
@@ -9,6 +9,8 @@ import lime.system.System as LimeSystem;
 import neko.vm.Gc;
 #elseif cpp
 import cpp.vm.Gc;
+#elseif hl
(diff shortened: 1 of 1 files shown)


loudoweb:
Thanks for your PR. I've made a similar PR one year ago here https://github.com/openfl/openfl/pull/2491. 
But yours seems more complete because you have added gc call, I haven't tested though.
Why to chose `untyped` instead of `int` for memory?

Cheemsandfriends:
> Why to chose `untyped` instead of `int` for memory?

I was just following what the majority had, which was `untyped`.
But I think it would be better to remove that since `hl.Gc` seems to be pretty close to neko's?

src/openfl/system/System.hx
```diff
@@ -195,6 +197,8 @@ import cpp.vm.Gc;
 	{
 		#if (cpp || neko)
 		return Gc.run(true);
+		#elseif hl
+		return Gc.enable(true);
```
Apprentice-Alchemist: If I remember correctly `Gc.major` is the correct function.
```suggestion
		return Gc.major();
```

src/openfl/system/System.hx
```diff
@@ -262,6 +266,8 @@ import cpp.vm.Gc;
 		#elseif (js && html5)
 		return
 			untyped #if haxe4 js.Syntax.code #else __js__ #end ("(window.performance && window.performance.memory) ? window.performance.memory.usedJSHeapSize : 0");
+		#elseif hl
+		return untyped Gc.stats().currentMemory;
```
Apprentice-Alchemist:
```suggestion
		return Std.int(Gc.stats().currentMemory);
```

Apprentice-Alchemist (changes requested):

PR #2570, opened 2022-07-04
Added pan to lime AudioSources, so use that for pan instead of position

justin-espedal:
This PR is based on potential lime changes found here: https://github.com/openfl/lime/pull/1557

Since OpenFL, being based on the Flash API, only deals with sound panning and not full-on spatialized audio, this switches us to using Lime AudioSource's new "pan" property, which is more natural and also fixes the HTML5 sound playback issues mentioned in the accompanying lime PR.

### DIFF
--- a/src/openfl/media/Sound.hx
+++ b/src/openfl/media/Sound.hx
@@ -646,11 +646,7 @@ class Sound extends EventDispatcher
 		if (loops > 1) source.loops = loops - 1;
 
 		source.gain = volume;
-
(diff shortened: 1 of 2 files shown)


DigiEggz:
I've tested this with https://github.com/openfl/lime/pull/1557 and it's working well. It greatly improves the audio quality.
@justin-espedal The changes needed to fix the conflicts are luckily very simple and should only take a minute or two.

PR #2572, opened 2022-07-05
Enable crisp text rendering for pixel fonts

### LINKED ISSUE #2568: HTML5 text artifacts for pixel fonts
When targeting HTML5, there are artifacts around zoomed-in text, but only on some devices. I've updated to the latest files from this repository and the issue persists.

https://community.openfl.org/t/html5-text-artifacts/13514
https://community.openfl.org/t/html5-blurry-font-issue/12712

One one computer, the text renders normally:

On another computer, it renders with visual noise:

It may have to do with HiDPI settings (values > 1) and possibly for just pixel fonts, but I am uncertain. I'm willing to test any changes or suggestions!

DigiEggz:
Pixel fonts render with artifacts around the edges of text for devices with high pixel ratios and also for Chromium-based browsers. This compiler flag forces a 3x pixel ratio, eliminating the ghosting effect on both sides of text. Fixes #2568, which has more information on the issue.

### DIFF
--- a/src/openfl/display/_internal/CanvasTextField.hx
+++ b/src/openfl/display/_internal/CanvasTextField.hx
@@ -92,7 +92,9 @@ class CanvasTextField
 
 		if (textField.__dirty || graphics.__softwareDirty)
 		{
-			#if (openfl_disable_hdpi || openfl_disable_hdpi_textfield)
+            #if openfl_pixel_font
+            var pixelRatio = 3;
+			#elseif (openfl_disable_hdpi || openfl_disable_hdpi_textfield)
 			var pixelRatio = 1;
 			#else
 			var pixelRatio = renderer.__pixelRatio;
--- a/src/openfl/display/_internal/CanvasTextField.hx
+++ b/src/openfl/display/_internal/CanvasTextField.hx
@@ -92,8 +92,8 @@ class CanvasTextField
 
 		if (textField.__dirty || graphics.__softwareDirty)
 		{
-            #if openfl_pixel_font
-            var pixelRatio = 3;
+			#if openfl_pixel_font
+			var pixelRatio = 3;
 			#elseif (openfl_disable_hdpi || openfl_disable_hdpi_textfield)
 			var pixelRatio = 1;
 			#else


joshtynjala:
This doesn't feel like a good solution to me. It would be really helpful if you could provide some code/steps that will reliably reproduce this issue so that I (or one of the other contributors) can try to debug it too.

barisyild:
It's definitely not right to cover up instead of solving the problem.

barisyild:
@Metlmeta The fonts look pretty neat for me, if you send me a sample project, I can look for a way to solve the actual problem.

DigiEggz:
Sure thing! I've included a sample project. The problem stems from when trying to zoom on a low resolution project:

[TextDemo.zip](https://github.com/openfl/openfl/files/9049075/TextDemo.zip)

sample


**Note**: Make sure to use a Windows Chromium browser (Chrome, Brave, etc.) when testing this problem. Results also vary depending on your screen ppi. macOS experiences a similar problem, but only with Retina displays. Let me know if you can't get the issue to appear.

Thank you for your time!

joshtynjala:
Thanks! I can reproduce on macOS Safari, Chrome, and Firefox.

barisyild:
I think this is flixel problem

also isTextBlurry function name is quite interesting

image

barisyild:
I think the cause of the problem is:

flixel draws the textfield as it should be and converts it to a bitmapdata.

As FlxGame scales, it also scales the textfield bitmapdata with itself and degrades the quality.

barisyild:
Possible Solution: To prevent flxgame scaling, keeping the FlxGame size the same as the canvas size.

joshtynjala:
I see that `BitmapData.draw()` creates a new `CanvasRenderer`, and it currently does not change the renderer's `__pixelRatio` value. This could be a clue about what's going on.

DigiEggz:
> Possible Solution: To prevent flxgame scaling, keeping the FlxGame size the same as the canvas size.

There's still a problem when the resolution is set 1:1. It's smaller in this case, but you can still see a blurring problem going on:


Setting a CSS `image-rendering` property of `pixelated` shows the artifacts more clearly:




> I see that `BitmapData.draw()` creates a new `CanvasRenderer`, and it currently does not change the renderer's `__pixelRatio` value. This could be a clue about what's going on.

That's an interesting point. When I was sifting through that code, I noticed that the `FlxText` TextFields were never passed the Stage's pixelRatio, presumably because they never use code that passes the renderer.

What threw me through a loop is that even when I manually set a pixelRatio of 2, there is still a [blur on one side](https://github.com/openfl/openfl/issues/2568#issuecomment-1172913965). Only when I manually set a minimum value of 3 would the pixel fonts render without blurring on either side.

All of this all only applies to the HTML5 canvas renderer - all other rendering is flawless for pixel fonts for Cairo without any special code or changes. Please let me know if you'd like me to test anything; I'd be happy to give any suggestion a shot.

joshtynjala:
> I see that BitmapData.draw() creates a new CanvasRenderer, and it currently does not change the renderer's __pixelRatio value. This could be a clue about what's going on.

I'm leaning towards this not being a clue about anything — meaning that it's actually the correct behavior. I realized that BitmapData is _supposed_ to always draw display objects at pixelRatio of 1 because it is raw pixels. So that seems to be the correct behavior not to set __pixelRatio on the renderer that BitmapData.draw() creates.

I'm now leaning toward something else being rendered/drawn incorrectly inside the canvas renderer (but probably not the cairo renderer, since that doesn't show the same blurring). Maybe something is being drawn to BitmapData with smoothing enabled when it shouldn't be, or it's being drawn on a fractional pixel when it should be snapping to the nearest integer pixel. It could be that the TextField itself is drawn/rendered correctly, but an intermediate BitmapData or CanvasRenderingContext2D that the TextField is drawn to first, before being merged with other display objects, is what doesn't get drawn correctly.

It's still a bit tricky to debug because Flixel is involved, which adds complexity over OpenfL. Especially since I don't know the Flixel internals well (other than knowing that it ends up drawing pretty much everything to BitmapData).

DigiEggz:
I first noticed this problem on a macOS Retina display. The text looked slightly muddy, whereas it looked perfectly clean on a non-Retina display Mac (in both Firefox and Chrome).

A friend sent me a screenshot of his game client (on Windows) and I noticed the text looked even worse - even more so than on the Retina screen. It was especially confusing because he had the same pixel ratio as me. Instead of having a slight haze, there was heavy blurring on both sides of the text. When I tested on Windows, I realized I could reproduce the blurry text, but only on Chrome.

Is there a reason why the text looks perfectly clear on Windows Firefox with a pixel ratio of 1, but not in Chrome? I'm worried it's some sort of browser rendering engine disparity.

**Edit**: I noticed that in the "1 to 2 pixel ratio demo" of [this post](https://itecnote.com/tecnote/javascript-canvas-text-rendering-blurry/), the author does something similar where they multiply by 3 to get non-blurry text. Not sure if it applies to this issue or not, but I found the problem similar to the one faced here.

**Edit 2**: [This Stack Overflow question](https://stackoverflow.com/questions/68416291/how-to-draw-a-pixel-font-on-the-canvas-without-anti-aliasing) looks to be highly related.

joshtynjala:
> Edit 2: [This Stack Overflow question](https://stackoverflow.com/questions/68416291/how-to-draw-a-pixel-font-on-the-canvas-without-anti-aliasing) looks to be highly related.

It sure does. Unfortunately, the new canvas properties that address the issue don't seem to be fully supported yet.

DigiEggz:
At this point I don't think it's a fault with OpenFL. HTML5 canvas wants to anti-alias anything other than graphics and it depends on the browser to decide how to render the text: https://stackoverflow.com/questions/47139528/html5-canvas-text-anti-aliases-in-chrome-not-in-firefox

Users would either have to use a different rendering system for text or OpenFL could expose a new TextField variable (something like pixelFont?) that forces a pixelRatio of 3 for that specific TextField variable when set. Alternatively, all CanvasTextFields could have a default of 3 for their pixelRatio instead of the default of 1.

In all of the tests I ran, the only way I could get all browsers and high pixel density screens to render correctly was by scaling by 3. When I tested the "geometricPrecision" variable in Chrome, it _still_ didn't get rid of the aliasing.

Please let me know your thoughts and what we could do to circumvent the issue.

PR #2534, opened 2022-02-03
Add public functions for retrieving lists of cached assets.

EliteMasterEric:
This PR adds the following functions to the AssetCache:

- `getKeys(prefix:String = null):Array<String>`
- `getBitmapKeys(prefix:String = null):Array<String>`
- `getFontKeys(prefix:String = null):Array<String>`
- `getSoundKeys(prefix:String = null):Array<String>`
- `clearBitmapData(prefix:String = null):Void`
- `clearFonts(prefix:String = null):Void`
- `clearSounds(prefix:String = null):Void`

These functions should allow for applications to better manage and cleanup any cached assets they wish to explicitly unload.

PR #2515, opened 2021-11-27
SampleDataEvent

X-sam:
This is a freshened up version of #1723. 

Quoting the original PR:
> Implementation of Flash's SampleDataEvent using WebAudio on HTML5 and OpenAL on Windows/Neko.
> Though untested it should work on other targets using on OpenAL.

In the time since the original PR was proposed, Chrome has [banned autoplaying audio](https://developer.chrome.com/blog/autoplay/#webaudio), so here's a modified version of that test program supporting modern browser & desktop via neko/hl targets:
```haxe
package;

import openfl.display.BitmapData;
import openfl.events.MouseEvent;
import openfl.events.Event;
import openfl.display.InteractiveObject;
import openfl.display.SimpleButton;
import openfl.display.Bitmap;
import openfl.Assets;
import openfl.display.Sprite;
import openfl.media.Sound;
import openfl.events.SampleDataEvent;
import openfl.utils.ByteArray;

class Main extends Sprite {
	private var sampleRate:Int;
	private var bufferSize:Int = 8192;
	private var bpm:Int = 90;
	private var numberOfRows:Int = 48; // 48=3/4 time | 64=4/4 time
	private var currentRow:Int = 0;
	private var quarterNoteLength:Float;
	private var sixteenthNoteLength:Float;
	private var numOctaves:Int = 8;
	private var patterns:Array<Dynamic> = new Array();
	private var currentPattern:Int = 0;
	private var songOrder:Array<Int> = [0, 1];
	private var notes:Array<String> = ["c-", "c#", "d-", "d#", "e-", "f-", "f#", "g-", "g#", "a-", "a#", "b-"];
	private var frequencies:Array<Dynamic> = new Array();
	private var samplePosition:Int = 0;
	private var position:Int = 0;
	private var channel1:Dynamic = {
		volume: .05,
		waveForm: "",
		frequency: [],
		noteTriggered: false,
		envelopePos: 0
	};
	private var envelope:Array<Float> = new Array();
	private var echoBytes:ByteArray = new ByteArray();
	private var maxEchoBytes:UInt = 8192 * 256;
	private var echoing:Bool = true;
	private var echoDelay:Float;
	private var echoPosition:UInt = 0;

	public function new() {
		super();

		var bitmapData = new BitmapData(120, 80, false, 0x010203FF);
		var pressButton = new BitmapData(120, 80, false, 0x008800FF);
		var bitmap = new Bitmap(bitmapData);
		var pressbitmap = new Bitmap(pressButton);
[...]

PR #2577, opened 2022-07-29
Shader system improvements and fixes.

EliteMasterEric:

These changes were performed to support fixes for FlxRuntimeShader, as part of the pull request HaxeFlixel/flixel-addons#368. Making changes here prevents code duplication in FlxRuntimeShader, improving maintainability.

These changes have been tested on the Windows HXCPP platform.


JonnycatMeow (approved):
PLEASE MERGE THIS!!!

Raltyro (approved):

loudoweb:
Interesting. I was looking for a feature to change a CONST value at runtime. I think storing fragment body and header would be easier for me than parsing, diassembling an reassembling the whole shader. Some shaders need to use CONST in a for loop because non-constant values are forbidden, but different values can provide a better accuracy depending on what has been set in the uniform (e.g: Glow, Outline...). So it's looking great.

That being said, if you have only tested on Windows. This need to be tested in HL, MAC, Android, iOS and HTML5 before being merged.

Also can you provide details on your issue on "some platforms", which platforms are your talking about?

JonnycatMeow:
i tested it on mac and i can confirm it works on mac too

MAJigsaw77:
@MasterEric This pull request makes android to crash, I think is because of `glVersion` you added, gl version is different in android btw

JonnycatMeow:
> @MasterEric This pull request makes android to crash, I think is because of `glVersion` you added, gl version is different in android btw

what glversion should be put to use for android?

MAJigsaw77:
Idk the exact version

MAJigsaw77:
You have to find out, is your pull request anyway

JonnycatMeow:
the glsl version for android is #version 300 es how i found out is because of the flxFixedShader https://github.com/YoshiCrafter29/YoshiCrafterEngine/blob/45ddea338a211ff2d05b7d85a3a2c38bb8ab287f/source/FlxFixedShader.hx#L55

MAJigsaw77:
That version is wrong

EliteMasterEric:
Took a bit of time today to fast forward the branch, as well as implement a couple fixes.

Mainly, `glVersion` is now a string, and the value defaults to `300 es` on Android as suggested.

Hopefully this works for people, I've been having reports that AMD Radeon cards are having issues specifically.

MAJigsaw77:
Dude, the default version by openfl is 100

MAJigsaw77:
If you put 300 es it will broke everything

JonnycatMeow:
> If you put 300 es it will broke everything

*break

MAJigsaw77:
Btw, apply that for enscripted and ios too, they use gl es too

EliteMasterEric:
Marked this PR as draft pending further changes, including:

* Fixes for crashes when setting glVersion too high (built-in fragment headers need to be migrated).
* Thorough testing on desktop, web, and Android to ensure default values are widely compatible (since nobody can agree what the correct values are supposed to be).
* A `@:glExtensions` annotation to resolve issue #2601.

EliteMasterEric:
I've been doing some more investigation into what GLSL versions actually support what. 

I was previously under the assumption that `120` implemented the `uniform` keyword, and that not using this was the cause of the crashes some users have experienced while using shaders. 

However, I had actually misread the specification. `uniform` variables are available in all versions of GLSL, with `120` being the first version to support shader initialization of `uniform` variables, like so:

```
// This is valid in all versions of GLSL
uniform float iTime;
// This is only valid in GLSL 120 and above
uniform float iTime = 0.0;
```

`120` is NOT actually necessary to run most shaders, and many of the shaders which are not compatible can easily be converted by performing the initialization outside of GLSL (i.e. in Haxe code). `100` is sufficient to run most shaders, and furthermore appears to be widely supported. In the event I find a use case or platform in which `100` is not supported, compile-time macros can be used to modify the default shader version.

If a specific shader needs a feature introduced in a later language version, they can increase the language version, keeping in mind that most platforms have a very limited range of support (and your only solid choices are `100` and `300 es`). It is recommended that you rewrite the shader if you need a language feature unavailable in `100` or `300 es`.

MAJigsaw77:
`300 es` removes texture2D and some things that I forgot, 100 is the default version that openfl/lime uses because it still have these futures.

MAJigsaw77:
As i see in the commit, now it should work.

EliteMasterEric:
The feature was not removed. `300 es` removes `texture2D` but replaces it with `texture`. 

By modifying the macro that builds the shaders, I have successfully tested and confirmed support for all GLSL versions on Windows Desktop.

`100` is still a good default version to work with.

MAJigsaw77:
Btw, `300 es` isn't the only version that needs texture2D to be replaced there's `310 es` and `320 es`.

EliteMasterEric:
> Btw, `300 es` isn't the only version that needs texture2D to be replaced there's `310 es` and `320 es`.

All versions after `300 es` require this replacement.

JonnycatMeow:
i get an error trying to build an fnf game for mac and here is the error: 
Screen Shot 2022-12-02 at 10 12 12 PM

JonnycatMeow:
actually nvm i fixed it

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +436,256 @@ class ShaderMacro
 		return fields;
 	}
 
+	private static inline function getDefaultGLVersion():String
+	{
+		// Specify the default glVersion.
+		// We can use compile defines to guess the value that prevents crashes in the majority of cases.
+		return #if (android) "100" #elseif (web) "100" #else "100" #end;
```
JonnycatMeow: added another #elseif for mac and make it 120 because that the only version that is supported and then put else 100

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +436,256 @@ class ShaderMacro
...
+		var texture2DKeyword:EReg = ~/texture2D/g;
+		var glFragColorKeyword:EReg = ~/gl_FragColor/g;
+
+		switch (glVersion)
+		{
+			default:
+				// We don't know this GLSL version, so don't do anything.
+				return source;
+
+			case "100":
```
JonnycatMeow, line 475: add a #if(android||web) because if you don't it will give macos a crash error!!

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +436,256 @@ class ShaderMacro
...
+	{
+		if (StringTools.endsWith(glVersion, " compatibility")) return "";
+		if (StringTools.endsWith(glVersion, " core")) return buildGLSLHeaders(StringTools.replace(glVersion, " core", ""));
+
+		switch (glVersion)
+		{
+			default:
+				return "";
+
+			case "100":
```
JonnycatMeow, line 553: add a #if(android||web) because if you don't it will give macos a crash error!!

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +436,256 @@ class ShaderMacro
...
+			isFragment:Bool):Array<{name:String, behavior:String}>
+	{
+		if (StringTools.endsWith(glVersion, " compatibility")) return glExtensions;
+		if (StringTools.endsWith(glVersion, " core")) return buildGLSLExtensions(glExtensions, StringTools.replace(glVersion, " core", ""), isFragment);
+
+		switch (glVersion)
+		{
+			default:
+				return glExtensions;
+
```
JonnycatMeow, line 619: add a #if(android||web) because if you don't it will give macos a crash error!!

src/openfl/utils/_internal/ShaderMacro.hx
JonnycatMeow: make sure to do it for everyone of the code that says 100 even the extension

JonnycatMeow: it will look like this 
Screen Shot 2022-12-02 at 10 23 01 PM
Sc
![Screen Shot 2022-12-02 at 10 23 35 PM
reen Shot 2022-12-02 at 10 23 21 PM](https://user-images.githubusercontent.com/107599365/205427877-9db26101-4f12-4271-8151-41fc02717b68.png)
Screen Shot 2022-12-02 at 10 24 46 PM

Screen Shot 2022-12-02 at 10 23 50 PM

EliteMasterEric: Wrapping the 100 makes no sense, if you force `100` on a platform that doesn't support it you should obviously get an error.

Thanks for letting me know the correct default version for Mac though.

EliteMasterEric:
This PR is now ready and fully tested on a variety of platforms.

See more at my test workbench here: https://github.com/MasterEric/Flixel-GalaxyShaderWorkbench

EDIT: Make sure to use the build instructions to ensure the correct library versions are installed before testing.

EliteMasterEric:
With help from @JonnycatMeow I was able to test the workbench on their Mac platform. Notably, their version of OpenGL ONLY supports `110` and `120`. This may be significant for those looking to resolve compatibility issues.

JonnycatMeow:
bro can someone merge this please!!!!

Raltyro:
The Sound modification seems pretty handy for streaming
although is it on purpose that the samplerate is locked at 44100?

EliteMasterEric:
> The Sound modification seems pretty handy for streaming although is it on purpose that the samplerate is locked at 44100?

This pull request is not related to sound.

joshtynjala:
> This pull request is not related to sound.

Look on the Commits tab of this PR. You merged in a `SampleDataEvent` commit, and maybe some other stuff. I mentioned this to you on Discord, but you never responded.

EliteMasterEric:
> > This pull request is not related to sound.
> 
> Look on the Commits tab of this PR. You merged in a `SampleDataEvent` commit, and maybe some other stuff. I mentioned this to you on Discord, but you never responded.

Thank you for informing me of this. I will re-review my branch, make appropriate fixes, and merge them ASAP so that review can continue.

EliteMasterEric:
It appears that changes related to other pull requests (#2534 and #2515) accidentally got included in the working branch for this PR. Those changes have now been removed, and the PR has additionally been fast-forwarded to match the target branch.

Please inform me here of any additional review changes.

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +442,249 @@ class ShaderMacro
...
+			case "420":
+				return glExtensions;
+			case "430":
+				return glExtensions;
+			case "440":
+				return glExtensions;
+			case "450":
+				return glExtensions;
+			case "460":
+				return glExtensions;
```
Geokureli, line 646: I don't get why this switch seems so needlessly convoluted. 

- Why have a bunch of cases that behave exactly like the default case?
- why not group similar results using `case "400" | "410":`?
- why "310 es" pass "300 es" into `buildGLSLExtensions` and so on for "320 es" and "330"?
- why store a result var just to return it

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +442,249 @@ class ShaderMacro
...
+			case "440":
+				var result = buildGLSLHeaders("430");
+				return result;
+			case "450":
+				var result = buildGLSLHeaders("440");
+				return result;
+			case "460":
+				var result = buildGLSLHeaders("450");
+				return result;
+		}
```
Geokureli: this also seems needlessly overcomplicated, compared to something like:

```suggestion
		
		return switch (glVersion)
		{
			#if desktop
			case "300 es": "layout (location = 0) out vec4 fragColor;\n";
			#else
			case "300 es": "out vec4 fragColor;\n";
			#end
			case "310 es": buildGLSLHeaders("300 es");
			case "320 es": buildGLSLHeaders("310 es");
			case "330": buildGLSLHeaders("320 es");
			case "400": buildGLSLHeaders("330");
			case "410": buildGLSLHeaders("400");
			case "420": buildGLSLHeaders("410");
			case "430": buildGLSLHeaders("420");
			case "440": buildGLSLHeaders("430");
			case "450": buildGLSLHeaders("440");
			case "460": buildGLSLHeaders("450");
			default: "";
		}
```

EliteMasterEric: This is nice, I forgot about return switch.

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +442,211 @@ class ShaderMacro
...
+		var varyingKeyword:EReg = ~/varying ([A-Za-z0-9]+) ([A-Za-z0-9_]+)/g; // g to match all
+
+		var texture2DKeyword:EReg = ~/texture2D/g;
+		var glFragColorKeyword:EReg = ~/gl_FragColor/g;
+
+		switch (glVersion)
+		{
+			default:
+				// We don't know this GLSL version, so don't do anything.
+				return source;
```
T1mL3arn, line 479: Maybe show an explicit warning like "OpenFl doesn't know glsl version ${glVersion}, no conversion will happen" ?

JonnycatMeow: So then how do we use a certain glsl version in the project.xml or could it only be used in .hx code?

EliteMasterEric: You don't specify GLSL version in the project.xml you specify it when you construct the shader

JonnycatMeow: oh ok so then the default whould be 120 for desktops and for web it whould be 300?

EliteMasterEric: > oh ok so then the default whould be 120 for desktops and for web it whould be 300?

The default would be the result of `getDefaultGLSLVersion`

JonnycatMeow: ohhh ok

JonnycatMeow: then how whould i use a certain glsl version?

EliteMasterEric: > then how whould i use a certain glsl version?

You can use `@:glVersion("120")` the same way you use `@:glFragmentBody("...")`.

As a note there is an `FlxRuntimeShader` in `flixel-addons`. It was originally built by me and ran on this, but it was modified to not use this code so it could be merged. When this is merged it will get changed back and then it will support setting the `glVersion` among other things.

Raltyro:
> The Sound modification seems pretty handy for streaming
> although is it on purpose that the samplerate is locked at 44100?

When did i said this what

EliteMasterEric:
> > The Sound modification seems pretty handy for streaming
> > although is it on purpose that the samplerate is locked at 44100?
> 
> When did i said this what

I accidentally included another PR in this one by mistake.

JonnycatMeow:
there should be core and compatibility support for macOS 
https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.1.50.pdf 
this shader version only uses core for macOS because compatibility craps its self

JonnycatMeow:
Screen Shot 2024-02-21 at 12 45 41 AM
365/ae5dac53-00b8-4ffb-a752-6939d5622e37)

JonnycatMeow:
Screen Shot 2024-02-21 at 12 46 24 AM

JonnycatMeow:
Screen Shot 2024-02-21 at 12 46 36 AM

JonnycatMeow:
Screen Shot 2024-02-21 at 1 19 20 AM

JonnycatMeow:
My source https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.1.50.pdf

EliteMasterEric:
Closed in favor of #2715.

PR #2578, opened 2022-08-02
Make sure rotation is within -180 and +180

fourst4r:
As per flash docs for DisplayObject.rotation

> Indicates the rotation of the DisplayObject instance, in degrees, from its original orientation. Values from 0 to 180 represent clockwise rotation; values from 0 to -180 represent counterclockwise rotation. Values outside this range are added to or subtracted from 360 to obtain a value within the range. For example, the statement my_video.rotation = 450 is the same as my_video.rotation = 90.

### DIFF
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -2079,6 +2079,9 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if (open
 	{
 		if (value != __rotation)
 		{
+			while (value > 180) value -= 360;
+            		while (value < -180) value += 360;
+			
 			__rotation = value;
 			var radians = __rotation * (Math.PI / 180);
 			__rotationSine = Math.sin(radians);
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -2079,8 +2079,15 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if (open
 	{
 		if (value != __rotation)
 		{
-			while (value > 180) value -= 360;
-            		while (value < -180) value += 360;
+			value = value % 360.0;
+			if (value > 180.0) 
+			{
+				value -= 360.0;
+			}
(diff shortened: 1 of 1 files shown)


dimensionscape:
In Flash, if rotation == 180 and you increment by 1, what is the output if you trace rotation?

Vortelio:
What if the angle is 2000000000 degrees?
Maybe something like this:

```
if (Math.abs(value) > 180)
{
	value = value / Math.abs(value) * (Math.abs(value) - Math.round(Math.abs(value) / 360) * 360);
}
```

joshtynjala:
> In Flash, if rotation == 180 and you increment by 1, what is the output if you trace rotation?

I did some tests:

```hx
target.rotation = 180;
trace(target.rotation); // 180
target.rotation++;
trace(target.rotation); // -179
```

```hx
button.rotation = -180;
trace(button.rotation); // -180
button.rotation--;
trace(button.rotation); // 179
```

```hx
button.rotation = -181;
trace(button.rotation); // 179
```

```hx
button.rotation = 181;
trace(button.rotation); // -179
```

joshtynjala:
> What if the angle is 2000000000 degrees?
> Maybe something like this:

Good point. I think that it would be better for performance to use the modulo operator, though. Your suggestion has a lot of function calls.

I think this will work:

```haxe
value = value % 360.0;
if (value > 180.0) {
	value -= 360.0;
}
else if (value < -180.0) {
	value += 360.0;
}
```

dimensionscape:
> > What if the angle is 2000000000 degrees?
> > Maybe something like this:
> 
> Good point. I think that it would be better for performance to use the modulo operator, though. Your suggestion has a lot of function calls.
> 
> I think this will work:
> 
> ```hx
> value = value % 360.0;
> if (value > 180.0) {
> 	value -= 360.0;
> }
> else if (value < -180.0) {
> 	value += 360.0;
> }
> ```

Yeah this isn't quite as heavy. 👍

If it's supposed to be clamped, lets do it.

PR #2580, opened 2022-08-08
Implement word selection for textfield, start word selection by double clicking and dragging

ShaharMS:
Heres a PR that implements word selection.

To summarize, i made two external, non-completion variables - `__wordSelection` and `__wordSelectionInitialIndex`.

 - `__wordSelection` is used to check if the textfield is currently "double clicked". if so, regular dragging is modified to only account for word starts/endings as indices
  - `__wordSelectionInitialIndex` is used to know where to start the selection from, and to which direction (for example, if `__caretIndex > __wordSelectionInitialIndex`, we need to select from the start of the first selected word to the end of the last. othewise, select from the end of the first word selected, to the start of the last)
  
  I also made two non-completeion functions to calculate the selection indices:
  
This one is used to set `__caretIndex`
https://github.com/ShaharMS/openfl/blob/f8d1f60c2391d6aaf0ebdb1be6e1a08b7aad443a/src/openfl/text/TextField.hx#L1964-L1987

And this one is used to set `__selectionIndex`
https://github.com/ShaharMS/openfl/blob/f8d1f60c2391d6aaf0ebdb1be6e1a08b7aad443a/src/openfl/text/TextField.hx#L1989-L2011


joshtynjala:
> regular dragging is modified to only account for word starts/endings as indices

Interesting! I knew that double-clicking a word would select the whole word, but I didn't know that if you dragged after double-clicking, it would expand selection to whole words instead of character by character.

ShaharMS:
another thing i didnt know (which i plan to implement too) is that a triple click selects an entire pragraph, and dragging selects from the initial paragraph to the one under the mouse. i thought it selected lines, but thats probably because i use vscode too much😆

ShaharMS:
Another thing, you might have noticed i removed the event listener for `DOUBLE_CLICK`. i still kept the function just in-case we want to revert some things, but the event listener can be removed

ShaharMS:
would this be merged right after 9.2? if so, should i also add the triple click, paragraph-selection capabilities?

ShaharMS:
I know what to fix to make CI working, I'll do the fix after 9.2 is out

dimensionscape:
It just your use of the final keyword. This is only supported since haxe 4. Since we still support backwards compatibility with older versions, you should avoid using syntax only available in new releases of Haxe.

ShaharMS:
can someone trigger CI to check the last commit? i wan to see if there's anything i missed

ShaharMS:
Can this be merged now? i want to work on the other selection changes, but I'm wating for this to get merged

ShaharMS:
Bumping this, the selection changes, or at least further modification of the way openfl does text selection is _really_ needed, especially for web platforms, on which OpenFl's textfields stick out like a sore thumb...
#2587 may still be relevant. If not, ill merge from develop and make progress on this PR

And another question - if this PR is still both relevant & applicable, can it be merged into 9.4.0-Dev instead? seems more fitting imo

joshtynjala:
I merged #2587 the other day, and I'd like to merge this one too (into 9.4.0-Dev). I think that this one should be updated to use `clickCount`, now that it is available (instead of the `decideDoubleClick` extra listener workaround), but I'm guessing that's probably the only change necessary.

ShaharMS:
I'm quite busy today, and there's also some work to do with paragraph selection iirc (triple clilck)
Will probably get to fix things tomorrow

ShaharMS:
All is pretty much done! some things have changed since the start of this pr:

 - There are now two "special" selection modes: `__wordSelection` and `__lineSelection`. Double click selects text by word, while triple click selects text by line. Both support dragging, and maintain the selection mode throughout the drag
 - `__wordSelectionInitialIndex` has been renamed to `__specialSelectionInitialIndex`
 - `__getPositionByWord` and `__getOppositeWordBound` has been slightly changed to support line selection too, and are now called `__getPositionByIdentifier` and `__getOppositeIdentifierBound`.
 - `onDoubleClick` callback has been completely removed, as it has no use now - it was used for word selection only, and it's behavior wasn't consistent with most other text inputs/editors, as you had to release the mouse to select a word. It is now inside the `MOUSE_DOWN`, `MOUSE_UP` and `MOUSE_MOVE` callbacks
 
Things i haven't tested:
 - phones. They should behave the same in theory, though

Things i haven't implemented/done:
 - word selection using `ctrl`+`shift`+`→ ←`. works a little differently, and I didn't want to clutter this PR too much
 - fixing line selection ignoring the original line's length when dropping/rising to another, shorter line. Not too related to this PR, so i decided to do this in another PR if needed

joshtynjala:
@ShaharMS I may have some more feedback after I get into testing the code, but just reading through it, can you remove the `doubleClickEnabled = true` line in the constructor too? It appears to me that is no longer needed with the removal of `this_onDoubleClick`.

ShaharMS:
> @ShaharMS I may have some more feedback after I get into testing the code, but just reading through it, can you remove the `doubleClickEnabled = true` line in the constructor too? It appears to me that is no longer needed with the removal of `this_onDoubleClick`.

Or maybe go all the way and only allow these types of selection when doubleClickEnabled is true? Which route should I go?

joshtynjala:
I checked the other day, and TextField's doubleClickEnabled is false in Flash. So just remove that line, and it's all good.

ShaharMS:
@joshtynjala line removed. Can this be merged?

ShaharMS:
@joshtynjala line removed. Can this be merged?

joshtynjala:
Thanks! When I have time again, I'll do some testing, and merge it if everything looks good.

joshtynjala:
FYI — I found some regressions related to double-click selections and fixed them in the commits below.

f50feb44887f37887808b7f6b59deacbd6240ef8
440f28a007c1b0af2422c3b31f8e6dba0d5b0265
9f0b2d66712b58c5cd28523c3e05ce8026ea9642

PR #2585, opened 2022-08-17
fix #2584 unused extension variables in ShaderFilter.hx

### LINKED ISSUE #2584: ShaderFilter unused public variables
**Describe the bug**
ShaderFilter has unused public variables : topExtension, leftExtension, etc.

**To Reproduce**
1. Use your own Shader inside a ShaderFilter that outbounds your bounds Sprite (you can use this one https://github.com/loudoweb/openfl-shaders/blob/master/openfl/shaders/Outline.hx)
2. Set topExtension, leftExtension, etc. in the ShaderFilter
3. Compile and see that bounds haven't change

**Expected behavior**
It needs to change the bounds of the texture in the shader. Private variables __topExtension, __leftExtension work. But public variables don't.

loudoweb:
When creating a shaderfilter, we can now set extensions variables (to use a bigger area in the shader) using public variables. The public variables weren't used at all.

### DIFF
--- a/src/openfl/filters/ShaderFilter.hx
+++ b/src/openfl/filters/ShaderFilter.hx
@@ -68,7 +68,7 @@ class ShaderFilter extends BitmapFilter
 
 		@default 0
 	**/
-	public var bottomExtension:Int;
+	public var bottomExtension(default, set):Int;
 
 	/**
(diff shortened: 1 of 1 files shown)


JonnycatMeow (approved):
PLEASE MERGE THIS!!!

PR #2588, opened 2022-09-22
prevent logging unneeded shader info on js

Geokureli:
On JS, every shader was logged, by default. `gl.getShaderInfoLog` always returns a non-null string, as per [the doc](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getProgramInfoLog).

> When a [WebGLProgram](https://developer.mozilla.org/en-US/docs/Web/API/WebGLProgram) object is initially created, its information log will be a string of length 0.


joshtynjala:
Good fix. Merged. Thanks!

Issue #2574, opened 2022-07-12
UncaughtErrorEvents seems not correctly initialized in ApplicationMain


giuppe:
It seems that UncaughtErrorEvents is never enabled when checking for it in ApplicationMain.hx: https://github.com/openfl/openfl/blob/c8f994c66ef0755d6a75c7a125e9d2a672e4f080/assets/templates/haxe/ApplicationMain.hx#L146

```hx
if (stage.__uncaughtErrorEvents.__enabled)
{
	try
	{
		ApplicationMain.getEntryPoint();
		// ...
	}
	catch (e:Dynamic)
	{
		stage.__handleError(e);
	}
}
else
{
	ApplicationMain.getEntryPoint();
	// ...
}
```
At this point `__enabled` is always null, because every user-mandated `addEventListener(UncaughtErrorEvent)` (where `__enabled` could be set to true) is happening only after the call to `ApplicationMain.getEntryPoint()`, and thus the try/catch code path is never run. I could not find a way to enable UncaughtErrorEvents before this point.

This affects the error management, as some uncaught error slips through. I made some tests and I could still catch a `throw` happening in an ENTER_FRAME listener, but not in a MouseEvent.CLICK. The behavior also seems to change in an undefined way depending on the architecture, as in neko-release everything seems to work without any problem.

I could reproduce the issue in windows, both debug and release; and neko debug.

I could think of two ways to handle this:  either use `#if openfl_disable_handle_error` instead of checking for `__enabled` in ApplicationMain, like in `UncaughtErrorEvents.hx`: https://github.com/openfl/openfl/blob/c876e837d6dad29cf7a1acb559602eaec4d55bb5/src/openfl/events/UncaughtErrorEvents.hx#L63

... or initialize `__enabled` to true if openfl_disable_handle_error == false.

I can produce a PR if needed.

ShaharMS:
I'm also encountering the same issue, a PR would be greatly appreciated :D

giuppe:
I made a PR.

Geokureli:
Boosting Giuppe's PR, would love to see this merged

PR #2595, opened 2022-10-12
fixed UncaughtErrorEvents not working (#2574)

giuppe:
I found that there are two separate causes of issue #2574:
- commit cde0386f6f70728854084b19b37fadf380f86aef made it impossible to enable uncaught error handling, because the try/catch in ApplicationMain is run only if the user had already added an event listener - but that is not possible, as ApplicationMain is the first code run by the application;

- commit 6b58d04cecadde99f46e847d610d0dd3640f7faa made it impossible to enable uncaught error handling in debug mode. This line in `include.xml`:
```xml
<define name="openfl-disable-handle-error" unless="openfl-enable-handle-error" if="debug" />
```
... seems to only work when `openfl_enable_handle_error` is defined explicitly via the build command line (e.g. `-D  openfl_enable_handle_error`), but it doesn't have any effect if is defined in the local `Project.xml`. Probably because definitions in `Project.xml` are executed later than `include.xml`. That means that there is no way to configure a project to activate uncaught error handling in debug mode.

This PR moves the "unless enable" logic from `include.xml` to the class `UncaughtErrorEvents`, by initializing the static var `__enabled` to true if  `openfl_enable_handle_error || !openfl_disable_handle_error` is defined. This fixes both issues.

After the PR, the logic becomes:
- in debug mode: error handling disabled by default, enabled by defining `openfl-enable-handle-error`;
- in non-debug mode: enabled by default, disabled by defining `openfl-disable-handle-error`.

The logic of "disable the error handling if there are no listeners" remains untouched.

This seems to me to be the intention behind the two original commits.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -56,7 +56,7 @@
 
 	<section unless="custom-backend">
 
-		<define name="openfl-disable-handle-error" unless="openfl-enable-handle-error" if="debug" />
+		<define name="openfl-disable-handle-error" if="debug" />
 
 		<section if="native">
 
--- a/src/openfl/events/UncaughtErrorEvents.hx
+++ b/src/openfl/events/UncaughtErrorEvents.hx
@@ -39,7 +39,7 @@ import openfl.events.UncaughtErrorEvent;
 #end
 class UncaughtErrorEvents extends EventDispatcher
 {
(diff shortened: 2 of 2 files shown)


joshtynjala:
Closed with #2656

PR #2597, opened 2022-10-25
Load external shader file at compile time via @:glFragmentSourceFile and @:glVertexSourceFile

giuppe:
This PR adds a way to load an external file to be used as a shader (fragment or vertex).

Usage:
```hx
class MyShader extends Shader
{
// ...
@:glFragmentSourceFile("./shaders/myShader.frag")
// ...
}
```

This could help separate the actual GLSL code from its OpenFL management. Most IDEs can correctly highlight/format the syntax of an external shader file, while long strings (as the `@:glFragmentSource()` argument) are usually rendered as a uniform color and must be formatted by hand; so this could be a nice quality of life change.

This PR is changing the `ShaderMacro.build()` function so that:
- it would check for the metas `glFragmentSourceFile`, `:glFragmentSourceFile`, `glVertexSourceFile`, and `:glVertexSourceFile`;
- if one of the metas is found and the argument file exists, its content will be set as the `glFragmentSource`/`glVertexSource`, overriding the `@:glFragmentSource`/`@:glVertexSource` argument (if defined).
- if the file does not exist or cannot be read, an error is returned at compile time.

### DIFF
--- a/src/openfl/utils/_internal/ShaderMacro.hx
+++ b/src/openfl/utils/_internal/ShaderMacro.hx
@@ -27,6 +27,9 @@ class ShaderMacro
 		var glFragmentSource = null;
 		var glVertexSource = null;
 
+		var glFragmentSourceFile = null;
+		var glVertexSourceFile = null;
+
 		for (field in fields)
 		{
 			for (meta in field.meta)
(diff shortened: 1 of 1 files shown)


T1mL3arn:
Completely and totally upvote this. Why is it still not merged?

PR #2605, opened 2022-11-16
Delete Context3DAlphaMaskShader.hx

mgood7123:
not used anywhere and interferes with `--macro include('openfl') --macro keep('openfl')`


joshtynjala:
Fixed in e08ab2d8f8d53544caf3c1db49692b6bac06ec29

PR #2609, opened 2023-01-10
Fix 3D matrix setter on `openfl.geom.Transform`

Cheemsandfriends:
While trying to do some 3D matrix stuff, I noticed that my Sprite didn't appear for some reason, but it would appear when using a normal one!

and I mean cases like `sprite.transform.matrix3D = new Matrix3D()` would make the sprite disappear for some reason.


joshtynjala:
Sorry it took a bit of time to merge. Thanks!

Issue #2358, opened 2020-06-29
Clipping problem with Graphics.cubicCurveTo on HTML5


stg:
As illustrated by the attached example, there is a clipping problem when drawing curves using cubicCurveTo on HTML5 targets.
While not highly pronounced in the example, for more extreme curves it can be.

[ProblemDemo2.zip](https://github.com/openfl/openfl/files/4844490/ProblemDemo2.zip)

![screenshot](https://user-images.githubusercontent.com/304625/85988575-89e48a80-b9ef-11ea-9db9-77103ef0b0d0.png)

m8mike:
This issue should be closed. It was resolved in commit 66cad37. The problem was in the calculation of bounding boxes of curves.

PR #2610, opened 2023-01-12
Fix: calculation of bounding boxes of cubicCurveTo (#2358)

m8mike:
1. There was a typo - __positionX was used insead of __positionY (lines 519 - 536).
2. I added an if statement to check whether w == 0 to prevent division by zero.

### DIFF
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -516,24 +516,24 @@ import js.html.CanvasRenderingContext2D;
 		iy1 = anchorY;
 		iy2 = anchorY;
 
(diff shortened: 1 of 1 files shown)


joshtynjala:
Commit 93ddcf237e3982eee8cb3121e2171613d23dc23c is perfectly straightforward, and I'm ready to merge it.

For the other commits, do you think you could write a few basic functions where you draw a curve, and the bounds are wrong if those commits are missing. I'd just like to do a few quick tests to see the difference before and and after, and that would a big help for me to get started.

Thanks!

m8mike:
Wait, I've made a mistake on the line 507. Also, there should be a check whether the value which we are taking square root of is negative. I better refactor it to reduce the size of the function and to avoid more mistakes. Brb with another pull request.

PR #2612, opened 2023-01-14
fix setTimeout memory leak

barisyild:
setInterval should be cleared manually, but setTimeout should be cleared automatically.

My pull request fixes leak problem.

Reproduce:

```haxe
class LeakObject {
    private var timeout:UInt;

    public function new() {
        timeout = openfl.Lib.setTimeout(leak, 1000);
    }

    private function leak():Void
    {
        //If clearTimeout not work, a memory leak occurs.
        //openfl.Lib.clearTimeout(timeout);
        trace("LeakObject never released from memory!");
    }
}
```

### DIFF
--- a/src/openfl/Lib.hx
+++ b/src/openfl/Lib.hx
@@ -559,7 +559,7 @@ import js.Browser;
 		__timers[id] = Timer.delay(function()
 		{
 			Reflect.callMethod(closure, closure, args == null ? [] : args);
-			clearTimeout(id);
+			__timers.remove(id);
 		}, delay);
 		return id;
 	}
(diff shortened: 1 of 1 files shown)


player-03:
Maybe just call `__timers.remove(id)` instead of `clearTimeout(id)`. Not a big deal, but `clearTimeout()` also calls `timer.stop()`, which is redundant.

barisyild:
> Maybe just call `__timers.remove(id)` instead of `clearTimeout(id)`. Not a big deal, but `clearTimeout()` also calls `timer.stop()`, which is redundant.

Yes, you are right

barisyild:
> Maybe just call `__timers.remove(id)` instead of `clearTimeout(id)`. Not a big deal, but `clearTimeout()` also calls `timer.stop()`, which is redundant.

Thanks i improved

joshtynjala:
Can you move `__timers.remove(id)` before the call to `Reflect.callMethod()`? While it's probably unlikely to happen, the called method could have side effects that modify `__timers`. Best to be safe.

After that change, I'll be happy to merge.

PR #2614, opened 2023-01-22
Arabic ligature + Better RTL

Shallowmallow:
Here's a pull request to start a discussion on rtl/ltr . There still are a few things I would like to correct before an openfl release.

### Improvements ( only on non js targets)

- [x] **Arabic ligatures**
- [x] **rTL alignment**

- [x] **Automatic rTL detection**
_The automatic detection is done by harfbuzz. It detects from the first few characters_
_For the detection to work, the script and direction must be invalid ones_
_I've enabled by default, by setting the language, the script, and direction to invalid ones_
_I created a new text script GUESS which is in fact just the harfbuzz invalid one_ 

- [x] **I've expose language, script and text direction  to the textField**
_I've done this before coding the detection part. I'm not sure it has so many uses now the detection is coded_
_TextDirection uses openfl.text.\_internal.TextLayout.TextDirection. If we keep exposing the textdirection in textField, maybe it's better that TextDIrection has own hx file not in \_internal_

- [x] **You can use TextFormat to control rtl /ltr**
- [x] **You can use htmlText and use TextFormat tags to control rtl/ltr**
_For example you can do `textfield.htmlText = "1234 <textformat dir='rtl'> 567 12 34</textformat>"`_

### Weird Bugs
- `textfield.htmlText = "12345<textformat dir='rtl'>12345</textformat>"`
_If you have the exact same word in rtl and ltr problem, it's because of caches. It's not really important IRL, because you usually use  different scripts in rtl and ltr_
- when you do a selection on a word where there are characters that are not in the font, it crashes
_It may be just an openfl bug and not this patch bug,  just noticed it here_
- The is bug where two carets appear only on windows
- There is bug if you use <textformat> in normal ltr words, navigating in the word is difficult, it's the only negative agains openfl official but will be soon corrected.

### What is left to do
- [ ] right margin ( but not sure in normal ltr words openfl)
- [ ] more testing (long words, etc ., bidirectional, selection, pos)
- [ ] make selection and interactivity work better on bidirectional textfield
- [ ] make  html5 working
_ [ ] I've focused more on harfbuzz. 

### What I don't plan to do
- I don't plan to implement Bidi characters, now that the textformat tag is working. Because it's maybe better to implement an external library Fribidi etc .

### DIFF
--- a/src/openfl/text/TextField.hx
+++ b/src/openfl/text/TextField.hx
@@ -1,5 +1,6 @@
 package openfl.text;
 
+import openfl.globalization.LocaleID;
 #if !flash
 import haxe.Timer;
 import openfl.text._internal.HTMLParser;
@@ -7,6 +8,8 @@ import openfl.text._internal.TextEngine;
 import openfl.text._internal.TextFormatRange;
 import openfl.text._internal.TextLayoutGroup;
 import openfl.text._internal.UTF8String;
+import openfl.text._internal.TextLayout;
+import openfl.text._internal.TextLayout.TextDirection;
 import openfl.utils._internal.Log;
 import openfl.display.DisplayObject;
 import openfl.display.Graphics;
@@ -664,7 +667,11 @@ class TextField extends InteractiveObject
 		have word wrap. The default value is `false`.
 	**/
 	public var wordWrap(get, set):Bool;
+	public var textDirection(get, set):TextDirection;
+	public var script(get, set):TextScript;
+	public var language(get, set):String;
 
+	@:noCompletion private var __textDirection:TextDirection;
 	@:noCompletion private var __bounds:Rectangle;
 	@:noCompletion private var __caretIndex:Int;
 	@:noCompletion private var __cursorTimer:Timer;
@@ -796,6 +803,18 @@ class TextField extends InteractiveObject
 				get: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function () { return this.get_wordWrap (); }"),
 				set: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function (v) { return this.set_wordWrap (v); }")
 			},
+			"textDirection": {
+				get: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function () { return this.get_rTL (); }"),
+				set: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function (v) { return this.set_rTL (v); }")
+			},
+			"script": {
+				get: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function () { return this.get_script(); }"),
+				set: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function (v) { return this.set_script (v); }")
+			},
+			"language": {
+				get: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function () { return this.get_language (); }"),
+				set: untyped #if haxe4 js.Syntax.code #else __js__ #end ("function (v) { return this.set_language (v); }")
+			},
 		});
 	}
 	#end
@@ -818,6 +837,8 @@ class TextField extends InteractiveObject
 		__displayAsPassword = false;
 		__graphics = new Graphics(this);
 		__textEngine = new TextEngine(this);
+		__textEngine.language = language;
+		__textEngine.script = script;
 		__layoutDirty = true;
 		__offsetX = 0;
 		__offsetY = 0;
@@ -914,7 +935,9 @@ class TextField extends InteractiveObject
 
 		__updateLayout();
 
-		x += scrollH;
+		if (textDirection.backward) x -= scrollH;
+		else
+			x += scrollH;
 
 		for (i in 0...scrollV - 1)
 		{
@@ -1808,14 +1831,21 @@ class TextField extends InteractiveObject
 				try
 				{
 					var x = group.offsetX;
+					if (group.textDirection().backward) x += group.width;
 
 					for (i in 0...(charIndex - group.startIndex))
 					{
-						x += group.getAdvance(i);
+						if (group.textDirection() == LEFT_TO_RIGHT) x += group.getAdvance(i);
+						else
+							x -= group.getAdvance(group.positions.length - 1 - i);
 					}
 
 					// TODO: Is this actually right for combining characters?
 					var lastPosition = group.getAdvance(charIndex - group.startIndex);
+					if (group.textDirection().backward)
+					{
+						lastPosition = group.offsetX - group.width;
+					}
 
 					rect.setTo(x, group.offsetY, lastPosition, group.ascent + group.descent);
 					return true;
@@ -1884,7 +1914,9 @@ class TextField extends InteractiveObject
 	{
 		__updateLayout();
 
-		x += scrollH;
+		if (textDirection.backward) x -= scrollH;
+		else
+			x += scrollH;
 
 		for (i in 0...scrollV - 1)
 		{
@@ -1912,14 +1944,19 @@ class TextField extends InteractiveObject
 			if (firstGroup)
 			{
 				if (y < group.offsetY) y = group.offsetY;
-				if (x < group.offsetX) x = group.offsetX;
+				if (x < group.offsetX && (group.textDirection() != TextDirection.RIGHT_TO_LEFT)) x = group.offsetX;
+				if (x > group.offsetX && (group.textDirection() == TextDirection.RIGHT_TO_LEFT)) x = group.offsetX;
 				firstGroup = false;
 			}
 
 			if ((y >= group.offsetY && y <= group.offsetY + group.height) || (!precise && nextGroup == null))
 			{
-				if ((x >= group.offsetX && x <= group.offsetX + group.width)
-					|| (!precise && (nextGroup == null || nextGroup.lineIndex != group.lineIndex)))
+				if ((x >= group.offsetX && x <= group.offsetX + group.width) || (!precise && (nextGroup == null || nextGroup.lineIndex != group.lineIndex)))
+				{
+					return group;
+				}
+				else if (((group.textDirection() == TextDirection.RIGHT_TO_LEFT) && x <= group.offsetX + group
+					.width) || (!precise && (nextGroup == null || nextGroup.lineIndex != group.lineIndex)))
 				{
 					return group;
 				}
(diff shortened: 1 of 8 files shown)


Shallowmallow:
@piratejet @ShaharMS

I'm using this new PR to make it cleaner.

Compared to the old one  https://github.com/openfl/openfl/pull/2551
- TextAlignment works better 
- Html 5 compiles
- It doesn't have the last commit which is was a draft tp make better bidirectionality rtl and ltr. This will surely be done in a different PR

ShaharMS:
Niceee!

Might be nice to display what works and what's not in this pr with tasks, for example:
```
- [ ] something to do
- [x] something done
 ```
Should display like:

- [ ] something to do
- [x] something done

**Comment by barisyild** [comment]
html5 target already has partial rtl support.

**Comment by ShaharMS** [comment]
> when you do a selection on a word where there are characters that are not in the font, it crashes

Indeed an openfl bug, I remember encountering this from time to time...

**Comment by tobil4sk** [comment]
Came across a small bug with this on desktop targets, when the text of a textfield starts with a new line (e.g. `"\nאוטובוס"`) the direction of the text is deduced incorrectly as LtR. Here some sample code:

```haxe
package;

import openfl.events.MouseEvent;
import openfl.text.TextField;
import openfl.display.Sprite;

class Main extends Sprite
{
	var textField:TextField;

	public function new()
	{
		super();
		textField = new TextField();
		textField.text = "\nאוטובוס";

		final format = new TextFormat("arialuni.ttf");
		textField.defaultTextFormat = format;

		addChild(textField);

		textField.addEventListener(MouseEvent.MOUSE_OVER, mouseOver);
	}

	function mouseOver(_) {
		trace(textField.textDirection); // leftToRight
	}
}
```

ShaharMS:
Just interested - does this problem still accur when you replace the newline with every other weakly-directionaled chars like - or +? Or is it purely newline related?

tobil4sk:
I've just tested it now with - and +, those work as intended:
```
Source/Main.hx:38: rightToLeft
Source/Main.hx:38: rightToLeft
```

ShaharMS:
> right margin ( but not sure in normal ltr words openfl)

Actually curious, what's the problem with that one?
Right margin already works in openfl, and as long as lines of text get aligned correctly (i.e. lines that have an rtl char as their first strong char should be aligned to the right).

Is the problem here a missing implementation of line-based auto alignment? If not, where's the issue?

Shallowmallow:
If I remember well it's not right aligning,  it's right margins that didn't seem to work correctly.  Like if you put the right margin to 20for example. I was doing some tests with textfields with background colour, to see if left margins worked, and I noticed the right margins didn't seem to work, like I thought it should work.

flashultra:
@joshtynjala Is there any chance this pull request to be merged in the next release ?
For the tests I did everything worked great .

src/openfl/text/TextField.hx
```diff
@@ -3109,6 +3184,68 @@ class TextField extends InteractiveObject
...
+	{
+		if (value != __textEngine.language)
+		{
+			__setTransformDirty();
+			__dirty = true;
+			__layoutDirty = true;
+			__setRenderDirty();
+		}
+
+		if (LocaleID.RTL_LANGUAGES.contains(value))
```
tobil4sk, line 3241: This "contains" method call is causing compilation failures on older versions of Haxe.

tobil4sk:
Here is a fork of the branch with the current merge conflicts resolved: https://github.com/tobil4sk/openfl/tree/TempTextField

src/openfl/text/_internal/TextLayout.hx
```diff
@@ -153,6 +154,31 @@ class TextLayout
...
+					direction = cast direction.toHBDirection();
+				}
+				else
+				{
+					direction = cast __hbBuffer.direction;
+				}
+
+				// This has to do with a weird bg in Guess mode where the first textLayout repeats itself
+				// instead of "ال صبا صباح"  it writes "ال ال صبا صباح "
+				if (positions.length > 0) return;
```
tobil4sk, line 178: Is it possible for `positions.length > 0` to be true here? `positions` is set to an empty array on line 103.

tobil4sk: Looks like the answer is yes, this is because of the `direction` setter which changes `dirty` to true. Therefore, everywhere else in this function where `positions` is accessed, its getter does another call to `__position`. This is a bit messy.

tobil4sk: If `__direction` is set directly instead of via the setter, this issue can be avoided, then maybe this check is no longer needed?

Shallowmallow:
If you're talking about

'positions.length > 0`

It's been ages so i don't remember. I think it was a kind "weird bug" I was
never able to understand, that's why I put the comment with the exact
sentence. I'm not sure I even had this problem in other non arabic text.

If you're talking about the _direction. I think it was necessary, but I
don't remember why. Maybe it's for when you  have commas of stuff like this
where you can go both directions. As the script would be invalid in this
case,  the buffer direction would be ltr, when in reality it should be the
direction of the text.

src/openfl/text/_internal/TextLayout.hx
Shallowmallow: For 'positions.length > 0`

I only put it in case of the "weird bug" I'm not sure I even had this problem in other non arabic text.

For _direction. I think it was necessary  Maybe it's for when you have only special characters like commas  where you can go both directions As the script would be invalid in this case,  the buffer direction would be ltr, when in reality it should be the direction of the text.

tobil4sk: > I only put it in case of the "weird bug" I'm not sure I even had this problem in other non arabic text.

I think this "weird bug" was happening because `direction =` accesses the `set_direction` setter, which has a side effect of setting `dirty = true`:
https://github.com/openfl/openfl/blob/88a12d2072fa4922a0cdd2ac161620e67b65936f/src/openfl/text/_internal/TextLayout.hx#L237

This means that all further accesses of the `positions` member variable (e.g. on [line 206](https://github.com/openfl/openfl/blob/88a12d2072fa4922a0cdd2ac161620e67b65936f/src/openfl/text/_internal/TextLayout.hx#L206)) in the `__position()` method, which pass through the `get_positions` getter, now call `__position()` again:

https://github.com/openfl/openfl/blob/88a12d2072fa4922a0cdd2ac161620e67b65936f/src/openfl/text/_internal/TextLayout.hx#L218-L222

So you end up with a `__position()` call in the middle of another `__position()` call, which is how it is possible for the `positions` array to no longer be empty even though it was emptied at the start of the function:
https://github.com/openfl/openfl/blob/88a12d2072fa4922a0cdd2ac161620e67b65936f/src/openfl/text/_internal/TextLayout.hx#L103


By using `__direction =` directly instead of `direction =` equals, this side effect can be avoided, and we don't end up with nested `__position()` calls:
https://github.com/tobil4sk/openfl/commit/e4bf39554c7671d5c7a550032bc3120c208b8818

ShaharMS: Shouldn't testing this be as easy as just adding two underlines before where we set the direction?

I might be able to test with Hebrew, just not in the next 12 hours (not at home rn)

ShaharMS:
> I don't plan to implement Bidi characters, now that the textformat tag is working. Because it's maybe better to implement an external library Fribidi etc .

Shouldn't that be done automatically though? If we use harfbuzz for deciding text direction automatically, it should automatically assign the current line direction for punctuation/signs.

There's also no problem with numbers, as they're generally written ltr (even Arabic numerals, and Arabic is written rtl)

ShaharMS:
> Came across a small bug with this on desktop targets, when the text of a textfield starts with a new line (e.g. `"\nאוטובוס"`) the direction of the text is deduced incorrectly as LtR. Here some sample code:
> 
> ```hx
> package;
> 
> import openfl.text.TextFormat;
> import openfl.events.MouseEvent;
> import openfl.text.TextField;
> import openfl.display.Sprite;
> 
> class Main extends Sprite
> {
> 	var textField:TextField;
> 
> 	public function new()
> 	{
> 		super();
> 		textField = new TextField();
> 		textField.text = "\nאוטובוס";
> 
> 		final format = new TextFormat("arialuni.ttf");
> 		textField.defaultTextFormat = format;
> 
> 		addChild(textField);
> 
> 		textField.addEventListener(MouseEvent.MOUSE_OVER, mouseOver);
> 	}
> 
> 	function mouseOver(_) {
> 		trace(textField.textDirection); // leftToRight
> 	}
> }
> ```

I know I'm _very_ late to the party, but after revisiting this i remembered that most text fields decide text direction and alignment in a line-by-line basis. Can that be the case here, that we overlook and assume that the direction of the first line is the direction of the text, and therefore it defaults to LTR? genuinely curious here though, I'm not sure how harfbuzz/this PR does RTL support, im just used to working with both LTR and RTL text

barisyild:
is pull request dead?

AxGord:
Hi @Shallowmallow and OpenFL maintainers,

Wanted to add a data point: we've been carrying this PR's `TempTextField` branch in our production fork (https://github.com/soccertutor/openfl) since ~late 2024, used in a live native (hxcpp) OpenFL app handling Arabic-script text alongside LTR content. We recently rebased the 3 RTL commits (`eb78cbb1c`, `241eb3e65`, `88a12d207`) onto current `develop` (as of `9d376331a`) — the rebase was clean and the squashed combined diff comes out to ~7 files / +478 / -105 against current develop, all in `openfl/text/**`.

Tested working in our app across all native hxcpp targets we ship to:

- macOS (Retina)
- Windows
- Android
- iOS (iPad)

What we exercised on each:

- Arabic ligature shaping ✓
- Mixed LTR/RTL paragraphs via `<textformat dir="rtl">` in `htmlText` ✓
- `TextField.textDirection` / `script` / `language` setters ✓
- Caret movement and selection in RTL text ✓

The known caveats from the PR body (selection corner cases, the duplicate caret on Windows, mixed-direction selection edge cases) are still there — we work around them at app level by not exposing complex bidirectional editing surfaces to users — but the basic display & read-only path is solid in production. We haven't exercised the html5 path that was on this PR's TODO list, so I can't speak to its current state.

We'd love to see this land. A few questions:

1. @Shallowmallow — are you still around and willing to drive this to merge? Happy to help test specific changes or contribute fixes for any of the open items if useful.
2. @joshtynjala / @Dimensionscape — is there anything blocking review apart from author availability? If the PR is in "nobody to drive it" limbo, would the project be open to a fresh PR rebased onto current `develop` (with @Shallowmallow credited as the original author via commit metadata)?

If the consensus is option (2), we can prepare and push the rebased branch publicly as a starting point. Don't want to step on toes by opening a duplicate PR without checking first — preferring to revive this thread instead.

Thanks!

Shallowmallow:
Hi @AxGord ..
Oh so nice to see it used :)
I'm not sure I want to drive the merge ... 
I'm actually doing an universal text library ... which could work with openfl ( and lime's harbuzz and freetype) ... started to working on it again after seeing your post ( so thanks for the motivation! ), but could also work outside of openfl (using own harbuzz externs etc). I have a bit too much ambition for it so I'm not sure when it will be ready ... I'll tag/invite you when I'll create a repo.

barisyild:
@Shallowmallow Are you proficient in Arabic?

Shallowmallow:
No absolutely not, don't know how to read it at all  ^

Shallowmallow:
@barisyild @AxGord 
image
(image of alternative openfl textfied)

Lots of lovely features
- supports direct write on windows
- rtl/ltr , double carets
- compositor
- multiple fonts
- exclusion zones
- writing on multiple zones
I've put everything in https://github.com/temp-showcase/
I'll push stuff over these next few days/weeks and reread reorganise code
things I want to add but I'll take some pause
- speed / it's a little bit slower than openfl I think
- Top to bottom bottom to top hardening
- library to support atlas fonts
- making it work maybe with some other backends ( haxui backends ? )
- more subtitles /effects
- android stuff
- compositor without needing lime/sdl

Issue #2535, opened 2022-02-28
TypeError: this.text.indexOf is not a function


Sylveondev:
Hello, I'm on using openfl 9.1.0 in haxe 4.2.4 and I keep getting this error in my game. I thought it was probably a problem with haxe version so I tried it on 4.1.5 and the same error happens.
![image](https://user-images.githubusercontent.com/71215126/155985686-c06c2b7a-9848-4c08-b39b-0361bda67728.png)

barisyild:
The probable cause of the problem is that the text value is null.

Please use the value "" instead of the null value if you want to empty the text field.


this.text = null;
Instead of
this.text = "";

Borman2000:
@barisyild note that this is not the "null object reference" exception.
This error could occurs if value consists of numbers only while targets js/html5 (text value becomes Number which doesn't have "indexOf" function).

Workaround:
```
private function setValue(strNumber:String):Void {
	strNumber += "";	// cast to string
	textField.text = strNumber;
OR
	textField.text = strNumber.toString();	// cast to string
}
```

barisyild:
> @barisyild note that this is not the "null object reference" exception.
> 
> This error could occurs if value consists of numbers only while targets js/html5 (text value becomes Number which doesn't have "indexOf" function).
> 
> 
> 
> Workaround:
> 
> ```
> 
> private function setValue(strNumber:String):Void {
> 
> 	strNumber += "";	// cast to string
> 
> 	textField.text = strNumber;
> 
> OR
> 
> 	textField.text = strNumber.toString();	// cast to string
> 
> }
> 
> ```
> 
> 

It is correct to say null pointer exception because;

As in the example you mentioned, indexOf cannot access because its data type is a number, not a string. (In my example above, an exception occurs because it cannot access indexOf, so it is same)

This is a null pointer exception as the software cannot execute the function because indexOf is string function.

by the way, trying to fix this situation with a pull request is not correct because that is the behavior that should be.

PR #2616, opened 2023-02-06
Fix for issue #2535 "TypeError: this.text.indexOf is not a function"

Borman2000:
"TypeError: this.text.indexOf is not a function" exception occurs for js/html5 targets if value consists of numbers only - so clearly cast to string.

Haxe code to reproduce:
smth. like this
```
private function setValue(strNumber:String):Void {
	textField.text = strNumber;
}
```

### DIFF
--- a/src/openfl/text/TextField.hx
+++ b/src/openfl/text/TextField.hx
@@ -2962,6 +2962,10 @@ class TextField extends InteractiveObject
 
 	@:noCompletion private function set_text(value:String):String
 	{
+#if (js && html5)
+		value += "";		// cast to string (fix for #2535 'TypeError: this.text.indexOf is not a function')
+#end
+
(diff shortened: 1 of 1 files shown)


joshtynjala:
The type of the setter is already `String`, so if you are managing to pass in a different type, it should be considered a bug in _your_ code. Here's the function signature. The compiler will not allow any other value here, unless it's typed as `Dynamic`.

```haxe
private function set_text(value:String):String
```

It's "dangerous" to use `Dynamic`, and you need to be extra careful when you use it. In this case, the fix you need somewhere in your code is to convert the value to a string. `Std.string()` is a good way to do that.

```haxe
var strNumber:Dynamic = 123.4;
textField.text = Std.string(strNumber);
```

PR #2635, opened 2023-05-03
Fix: Track blendMode changes at the Context3D level instead of OpenGLRenderer level.

andresa88:
Fixes behavior when a blendMode is assigned to a child at the top of the layer stack in a container and the container's cacheAsBitmap is set to true. The child OpenGLRenderer will change the Context3D blend settings and so the parent renderer needs to be able to update the blend settings if the supplied blendMode differs from the one last used in the child renderer.

### DIFF
--- a/src/openfl/display/OpenGLRenderer.hx
+++ b/src/openfl/display/OpenGLRenderer.hx
@@ -753,6 +753,9 @@ class OpenGLRenderer extends DisplayObjectRenderer
 		__context3D.setScissorRectangle(null);
 
 		__blendMode = null;
+		if (__defaultRenderTarget == null) {
+			OpenGLRenderer.__currentBlendMode.remove(__context3D);
+		}
 		__setBlendMode(NORMAL);
 
 		if (__defaultRenderTarget == null)
--- a/src/openfl/display/OpenGLRenderer.hx
+++ b/src/openfl/display/OpenGLRenderer.hx
@@ -753,7 +753,8 @@ class OpenGLRenderer extends DisplayObjectRenderer
 		__context3D.setScissorRectangle(null);
 
 		__blendMode = null;
-		if (__defaultRenderTarget == null) {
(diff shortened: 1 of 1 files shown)


joshtynjala:
Can you write some simple code that demonstrates how to reproduce the issue? Thanks!

andresa88:
Sure here you go! The green square should be opaque, but is instead blended with BlendMode.ADD and so it appears transparent. But once the scene is re-rendered, the green squared is blended correctly.

OTOH if you set container.cacheAsBitmap to false, the green square is opaque from the start as expected.

```
package;

import haxe.Timer;
import openfl.display.Sprite;
import openfl.display.BlendMode;

class Main extends Sprite
{
	public function new ()
	{
		super ();

		stage.color = 0;

		var container = new Sprite();
		var red = createSquare(0xFFFF0000);
		var green = createSquare(0xFF00FF00);
		var blue = createSquare(0xFF0000FF);

		stage.addChild(container);
		container.addChild(red);
		container.addChild(blue);
		stage.addChild(green);

		blue.x = 100;
		blue.y = 100;
		green.x = 200;
		green.y = 200;

		// green should be opaque from the start
		container.cacheAsBitmap = true;
		blue.blendMode = BlendMode.ADD;

		var tick = new Timer(1000);
		tick.run = function()
		{
			// Re-render. This fixes the issue... at least until the next time 
			// the cached bitmap needs to be re-rendered
			green.x += 50;
			tick.stop();
		}
	}

	function createSquare(color: Int)
	{
		var square = new Sprite();
		square.graphics.beginFill(color);
		square.graphics.drawRect(0, 0, 400, 400);
		square.graphics.endFill();
		return square;
	}
}
```

joshtynjala:
I can reproduce the issue using the supplied code. I checked in Flash too, and I confirm that green should be opaque from the start.

PR #2636, opened 2023-05-05
[TextField] Fix commented prop breaks doc for the prop below

T1mL3arn:
The doc is broken in VSCode and also on api.openfl.org (see [wordWrap](https://api.openfl.org/openfl/text/TextField.html#wordWrap))

It should be `A Boolean value that indicates ...` but it is `Specifies whether to copy...`

image

Maybe this fix is not the proper way and commented not used properties (`useRichTextClipboard` and `alwaysShowSelection`) should be completely deleted.


joshtynjala:
Thanks! It looks like we normally use `#if false` to hide unimplemented properties/methods so that their docs don't affect anything else. I made that change before I merged.

PR #2639, opened 2023-05-09
Fix. Allow resizing of the Video object via the width and height props

andresa88:
Changing the width and height of the Video object does nothing when rendering in OpenGL. This change fixes that.


joshtynjala:
Thanks!

joshtynjala:
Adding steps reproduce this issue, in case it needs to be revisited in the future.

In the PlayingVideo sample (`openfl create PlayingVideo`), change the `netConnection_onNetStatus` method:

```haxe
private function netConnection_onNetStatus (event:NetStatusEvent):Void {
	
	if (event.info.code == "NetStream.Play.Complete") {
		
		Actuate.tween (overlay, 1, { alpha: 1 });
		video.width = 100;
		video.height = 50;
		
	}
}
```
Then compile for html5 and play the video until the end. At the end, the video should resize smaller. Without this commit, it does not resize.

PR #2643, opened 2023-05-20
Work around "instance constructor not found" error.

player-03:
Occasionally, the Haxe compiler gets confused and crashes because `ByteArrayData` doesn't have a constructor. Specifically, it prints "Instance constructor not found: openfl.utils.ByteArrayData."

I think this is usually caused by some other compile error, but the compiler won't tell you about that error until you declare a constructor. Then once you fix the other error(s) you can safely delete the constructor again.

I don't know how to reproduce the error reliably. It just happens once every couple months. But this change will permanently fix it and shouldn't have any side effects.


player-03:
Huh. The constructor was definitely already there. I'm going to have to take a closer look the next time this error comes up.

PR #2650, opened 2023-07-02
Include Moonshine IDE under Code Editors

B-Interactive:
Moonshine IDE claims support for Haxe, is maintained, and is cross platform.


dimensionscape (approved):
Very well.

PR #2653, opened 2023-07-31
fix offset math in ColorTransform.concat

Geokureli:
in [this thread](https://community.openfl.org/t/openfl-geom-colortransform-concat/6647/2?u=geokureli) Thomas outlined the expected offset formula:
```hx
concat_offset = multiplier2*offset1 + offset2;
```
Joshua replied that it was implemented, however the formula used was:
```hx
concat_offset = multiplier1*offset2 + offset1;
```

the doc states
> the effect is the same as applying the `second` color transformation after the _original_ color transformation.

but the previous code has this backwards

Came across this when I was implementing flixel tools for adobe animate's tint and brightness color transforms, I noticed that concatenating a full brightness or full tint on top of another transform did not completely override the initial transforms values.

### DIFF
--- a/src/openfl/geom/ColorTransform.hx
+++ b/src/openfl/geom/ColorTransform.hx
@@ -185,10 +185,10 @@ class ColorTransform
 	**/
 	public function concat(second:ColorTransform):Void
 	{
-		redOffset = redOffset * redMultiplier + second.redOffset;
-		greenOffset = greenOffset * greenMultiplier + second.greenOffset;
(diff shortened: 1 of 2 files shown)


Geokureli:
I'm not very well versed in unit testing visual utilities like this. For test_concat1, I clearly just took the new values and put then as the expected value, I feel like this is not great practice, and I'm open to suggestions

Edit: I see, so the test's purpose is to ensure that openfl behaves identical to air/flash, and it appears that my changes do not, perhaps concat's doc is incorrect?

Is my interpretation of the doc, wrong?

PR #2656, opened 2023-08-17
Restore uncaught error handling.

EliteMasterEric:
Essentially a rebase/fast-forward of #2595.

Ensures that `openfl-enable-handle-error` can be used, regardless if a build is debug or release.

### DIFF
--- a/include.xml
+++ b/include.xml
@@ -63,7 +63,7 @@
 
 	<section unless="custom-backend">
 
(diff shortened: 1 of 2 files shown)


ninjamuffin99:
vouch for this one, we've been using this internally on FNF and it has worked as expected and very nicely as a fix to handle uncaught errors!

Issue #2007, opened 2018-09-13
[HTML5] Context3D.drawToBitmapData produces transparent BitmapData


mayakwd:
During developing Starling app I've faced with the problem of drawing starling sprites to `BitmapData. DisplayObject.drawToBitmapData` produces  transparent `BitmapData`. By the help of @peteshand we've discovered that `context3D.drawToBitmapData` producing same result on HTML5 target. 

@peteshand  created sample to reproduce this issue: [drawToBitmapdata.zip](https://github.com/openfl/openfl/files/2378373/drawToBitmapdata.zip)

Issue in starling project: https://github.com/openfl/starling/issues/110

peteshand:
ok, down grading to Lime 6.4.0 and OpenFL 8.3.0 resolves the issue, so it must be a regression in the latest versions.

peteshand:
One difference I've noticed within the drawToBitmapData function is the following line:

OpenFL 8.3.0
`var image = window.renderer.readPixels ();`

OpenFL 8.4.1
`var image = #if (lime >= "7.0.0") window.readPixels () #else window.renderer.readPixels () #end;`

peteshand:
This issue has been resolved via this PR:
https://github.com/openfl/lime/pull/1240

mayakwd:
Thank you, Pete! I'll try to compile with develop branch to check if it's resolved :-)

mayakwd:
I done testing with latest develop + current starling version. I got total mess with masks and canvas rendering: 
1. Quad masks that are not added to displaylist stopped working, their positions attached to stage instead of local position of target, most content with masks is not displaying at all.
2. Rendering of canvas blinks like it's being erased every second frame.

Drawing into bitmapData still produces transparent pixels.

jgranick:
I just built your sample, it appears to look the same on Flash, Neko and HTML5, using current latest dev of Lime, OpenFL and Starling.

It's tan and white and first, and then clicking any of the other buttons makes it tan instead, and it stays tan over a black background no matter what

peteshand:
@mayakwd I'm guessing the Masking / Canvas problems you're seeing are a different issue to the drawToBitmapdata issue right? if so it might make sense to create a separate issue.

mayakwd:
@peteshand, yup, I just need more time, to extract some samples, to illustrate new issues. Closing this.

PR #2660, opened 2023-09-11
Fixed a crash in BlurFilter

moxie-coder:
apparently making the destPoint null crashes, so this fixes that

### DIFF
--- a/src/openfl/filters/BlurFilter.hx
+++ b/src/openfl/filters/BlurFilter.hx
@@ -185,6 +185,9 @@ import lime._internal.graphics.ImageDataUtil; // TODO
 	{
 		#if lime
 		var time = Timer.stamp();
+		if (destPoint == null)
+			destPoint = new Point();
+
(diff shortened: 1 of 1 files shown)


joshtynjala:
When calling `BitmapData.applyFilter()`, Flash throws the following error when `destPoint` is null:

```
TypeError: Error #2007: Parameter destPoint must be non-null.
	at flash.display::BitmapData/applyFilter()
```

With that in mind, OpenFL allowing `destPoint` to be null doesn't seem right to me.

moxie-coder:
> When calling `BitmapData.applyFilter()`, Flash throws the following error when `destPoint` is null:
> 
> ```
> TypeError: Error #2007: Parameter destPoint must be non-null.
> 	at flash.display::BitmapData/applyFilter()
> ```
> 
> With that in mind, OpenFL allowing `destPoint` to be null doesn't seem right to me.

yeah, apparently I found out by testing that making it null crashed and I made my own little fix for it, and if someone else had done the same thing, I thought fixing it would help but I feel like this just seems wrong, so I’m just gonna close this

moxie-coder:
I’ll keep the branch just in case

PR #2667, opened 2023-10-10
Sound Length Fix Part 1

### LINKED ISSUE #2666: Sounds over a certain length do not play properly.
I originally posted this on the HaxeFlixel repository, but it turned out to be an issue with OpenFL and lime.

Sounds over a certain length, seemingly around 25 minutes, do not play properly. An acquaintance discovered the details regarding the reason for this, and how to fix it:

image

Read below for additional details:
https://github.com/HaxeFlixel/flixel/issues/2899

Jaduplex:
Fixes https://github.com/openfl/openfl/issues/2666. Must be pulled along with the pull request to lime, as the fix only works when made in both places.

The lime pull request: https://github.com/openfl/lime/pull/1716

### DIFF
--- a/src/openfl/media/Sound.hx
+++ b/src/openfl/media/Sound.hx
@@ -867,8 +867,8 @@ class Sound extends EventDispatcher
 			#else
 			if (__buffer.data != null)
 			{
-				var samples = (__buffer.data.length * 8) / (__buffer.channels * __buffer.bitsPerSample);
-				return Std.int(samples / __buffer.sampleRate * 1000);
+				var samples = __buffer.data.length / ((__buffer.channels * __buffer.bitsPerSample)/8);
+				return Std.int(Math.min(Math.max((samples / __buffer.sampleRate * 1000), 0), 12173936));
 			}
 			else if (__buffer.__srcVorbisFile != null)
 			{


joshtynjala:
> Must be pulled along with the pull request to lime, as the fix only works when made in both places.

What happens if the change is made in OpenFL only? Or the change in Lime only? Will the app crash? Will audio sound wrong, such as being silent, sped up, slowed down, or distorted in some other way?

I ask because new OpenFL releases should generally remain compatible with a certain range of older releases of Lime (and similarly new Lime releases should be compatible witth a range of older releases of OpenFL). If this change breaks older releases of one or the other, then this PR may need to be expanded to include some conditional compilation.

player-03:
Looks like you're trying to avoid integer overflow? (Multiplying first generates a large number, whereas dividing first keeps it to a more reasonable range.)

But that doesn't explain why you're limiting the output to 12173936. That isn't even close to the integer limit.

player-03:
Might as well use multiplication rather than division. It's slightly faster and cuts down on the number of parentheses.

```diff
-__buffer.data.length / ((__buffer.channels * __buffer.bitsPerSample)/8)
+__buffer.data.length / (__buffer.channels * __buffer.bitsPerSample) * 8
```

Jaduplex:
I'm just doing what someone else told me to do, I don't really have much experience with math stuff like this

Jaduplex:
> > Must be pulled along with the pull request to lime, as the fix only works when made in both places.
> 
> What happens if the change is made in OpenFL only? Or the change in Lime only? Will the app crash? Will audio sound wrong, such as being silent, sped up, slowed down, or distorted in some other way?
> 
> I ask because new OpenFL releases should generally remain compatible with a certain range of older releases of Lime (and similarly new Lime releases should be compatible witth a range of older releases of OpenFL). If this change breaks older releases of one or the other, then this PR may need to be expanded to include some conditional compilation.

I haven't actually tested, but given how the final change works (sound length produces the same variable on non-too-long sounds both with and without the change), I'd assume it's fine. I'll give it a try, though.

Raltyro:
Pretty sure that doesn't change anything

Jaduplex:
> Pretty sure that doesn't change anything

It... does? It makes audios longer than ~25 minutes actually work.

Raltyro:
I mean you could probably just use int64 in this case instead of clamping

player-03:
We don't know why they chose to clamp the value, but it seems unlikely to be related to the integer limit, since 12173936 is less than a percent of that.

While we're on the subject, I vote against accepting this pull request until it can be properly explained.

Jaduplex:
> We don't know why they chose to clamp the value, but it seems unlikely to be related to the integer limit, since 12173936 is less than a percent of that.
> 
> While we're on the subject, I vote against accepting this pull request until it can be properly explained.

I, quite honestly, have no idea. It's the code I got from an acquaintance who told me how to fix the problem. I'll ask them.

Jaduplex:
I don't really wanna deal with this. If there's any objections or changes to this, I'm sure there's some other way to fix this somewhere. I just wanted to throw this out randomly because it's a fix I found useful, but apparently it's not as helpful as I first thought.

player-03:
Thanks for bringing it to our attention, in any case. If your friend does get back to you at some point, I'd be curious to hear their explanation.

Until then, I think we can go ahead with the rearranged equation.

Raltyro:
I'm actually curious to see whats up :eyes:

player-03:
I've [implemented a workaround](https://github.com/openfl/lime/commit/1a3a9bd5c54bfd52cd9a690296c243ccd044f940) on Lime's side of things. Not the same workaround proposed here, but one that I think keeps the equation more readable.

@joshtynjala, could you make the same change in OpenFL? Just the `* 8.0` change for now, and we'll look at clamping later.

Edit: here's a [quick demo](https://try.haxe.org/#a51950E8) to prove the solution works.

joshtynjala:
> @joshtynjala, could you make the same change in OpenFL? Just the * 8.0 change for now, and we'll look at clamping later.

Back from a quick vacation, so I may not have time this week, but yes, I can do that once I have some time.

joshtynjala:
Oh, I just saw your PR. Thanks!

PR #2674, opened 2023-11-09
fix scrollV 

barisyild:
Since the maximum value is not taken into account, an infinite loop occurs.

#### Reproduce:
```haxe
var textField:TextField = new TextField();
textField.addEventListener(Event.SCROLL, function(e:Event)
{
    trace("scroll");
    textField.scrollV = 100000000;
});
textField.scrollV = 100000000;
trace("PASS");
```


joshtynjala:
In Flash, it traces "PASS!" only. OpenFL gets a stack overflow, which is fixed with your change.

PR #2678, opened 2023-12-28
CanvasTilemap compressed texture crash fixed

barisyild:
When I create a tilemap with a compressed texture to draw using bitmapData, the application crashes.
The main source of the problem is that textures do not have the image variable.
As it should, it will skip textures and only draw what can be drawn.

This commit will solve the problem.

### DIFF
--- a/src/openfl/display/_internal/CanvasTilemap.hx
+++ b/src/openfl/display/_internal/CanvasTilemap.hx
@@ -138,7 +138,7 @@ class CanvasTilemap
 				}
 
 				bitmapData = tileset.__bitmapData;
-				if (bitmapData == null) continue;
+				if (bitmapData == null || bitmapData.image == null) continue;
 
 				if (bitmapData != cacheBitmapData)
 				{


joshtynjala:
@barisyild Looks fine to me. Merged.

Can you share some code that reproduces the issue, though? I'd like to test whether the same issue exists on native targets, and apply a similar fix.

barisyild:
You can reproduce the problem with this example.
With this code you can duplicate the behavior of the compressed texture (i.e. texture bitmapData only) 1:1.

```haxe
package;

import openfl.display.Sprite;
import openfl.display.BitmapData;
import openfl.display.Tilemap;
import openfl.display.Tile;
import openfl.display.Tileset;
import openfl.Lib;

class Main extends Sprite
{
	public function new()
	{
		super();
		
		var bitmapData:BitmapData = new BitmapData(200, 200, false, 0xAAAAAAAA);
		var textureOnlyBitmapData:BitmapData = toTexture(bitmapData);
		
		var tile:Tile = new Tile();
		tile.tileset = new Tileset(textureOnlyBitmapData);
		tile.tileset.addRect(textureOnlyBitmapData.rect);
		var tilemap:Tilemap = new Tilemap(1920, 1080);
		tilemap.addTile(tile);
		
		this.addChild(tilemap);
		
		var tilemapBitmapData:BitmapData = new BitmapData(Math.floor(tilemap.width), Math.floor(tilemap.height));
		tilemapBitmapData.draw(tilemap);
	}
	
    public static function toTexture(source:BitmapData):BitmapData
    {
        if(source.readable)
        {
            var context = Lib.current.stage.context3D;
            var texture = source.getTexture(context);
            return BitmapData.fromTexture(texture);
        }

        return source;
    }
}
```

joshtynjala:
Thanks!

PR #2680, opened 2024-01-03
Add mipmapping support in the OpenGL rendering

Cheemsandfriends:
MIP maps (also known as mipmaps), are bitmaps grouped together and associated with a texture to increase runtime rendering quality and performance. Each bitmap image in the MIP map is a version of the main bitmap image, but at a reduced level of detail from the main image. (taken from the [Actionscript developer guide](https://help.adobe.com/en_US/as3/dev/WS5b3ccc516d4fbf351e63e3d118a9b90204-7d5b.html))

This PR generates Mipmaps from a texture, plus sets a specific lod bias specified in the `SamplerState`

### DIFF
--- a/src/openfl/display3D/textures/TextureBase.hx
+++ b/src/openfl/display3D/textures/TextureBase.hx
@@ -313,7 +313,14 @@ class TextureBase extends EventDispatcher
 
 			if (__textureTarget == __context.gl.TEXTURE_CUBE_MAP) __context.__bindGLTextureCubeMap(__textureID);
 			else
+			{
 				__context.__bindGLTexture2D(__textureID);
+				if (state.mipfilter != MIPNONE)
+				{
+					gl.generateMipmap(__textureTarget);
+					state.mipmapGenerated = true;
+				}
+			}
 
 			var wrapModeS = 0, wrapModeT = 0;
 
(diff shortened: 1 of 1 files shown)


rainyt:
image
Hello, I have merged this PR and it appears that the rendering is working properly; I would like to ask, if I change all the renderings to MIPNEAREST, will it all take effect? How can I check if it is functioning properly?

rainyt:
Does it have a significant improvement in rendering performance? Will the memory decrease?

SomeGuyWhoLovesCoding:
Why wasn't this even merged yet? It's important for rendering performance

rainyt:
I think there may be other issues, such as' generateMipmap 'being constantly called, which could cause GPU memory leaks.

SomeGuyWhoLovesCoding:
Oh yeah

Cheemsandfriends:
> Hello, I have merged this PR and it appears that the rendering is working properly; I would like to ask, if I change all the renderings to MIPNEAREST, will it all take effect? How can I check if it is functioning properly?

I use this guide to see the differences between the mipmapping options

image

Cheemsandfriends:
> I think there may be other issues, such as' generateMipmap 'being constantly called, which could cause GPU memory leaks.

https://stackoverflow.com/questions/73908487/why-i-have-to-generate-mipmap-every-frame

rainyt:
> > I think there may be other issues, such as' generateMipmap 'being constantly called, which could cause GPU memory leaks.
> 
> https://stackoverflow.com/questions/73908487/why-i-have-to-generate-mipmap-every-frame

I see, that's great!

SomeGuyWhoLovesCoding:
Cool! That means this definitely needs to be merged.

Raltyro:
Still waiting for this so you can use texture2D(sampler2D, vec2, lodMapBias) in shaders

PR #2686, opened 2024-02-09
Refer to OpenAL constants instead of instance variables.

player-03:
Most likely, the intent was to allow users to override these values, but there isn't a clear use case for doing so. Audio libraries besides OpenAL might use other constants, but in that case `#if lime_openal` would fail and this code wouldn't even run.

If this is merged, we could then clean up the 74 variables from [`OpenALAudioContext`](https://github.com/openfl/lime/blob/develop/src/lime/media/OpenALAudioContext.hx). (This is only a fraction of the constants listed in `AL`, and I'm not sure why those 74 were picked. They aren't even all used.)

Speaking of, I also don't think `**OpenAL**AudioContext` would function correctly if the backend library isn't OpenAL. I guess we could rename it to just `AudioContext` and then add some backend-switching code... Oh wait, [we already have `lime.media.AudioContext`](https://github.com/openfl/lime/blob/develop/src/lime/media/AudioContext.hx).

Is there a use case I'm missing here? It just seems pointlessly risky to be using variables.

### DIFF
--- a/src/openfl/media/SoundChannel.hx
+++ b/src/openfl/media/SoundChannel.hx
@@ -14,6 +14,7 @@ import js.html.audio.ScriptProcessorNode;
 #if lime_openal
 import openfl.events.SampleDataEvent;
 import openfl.utils.ByteArray;
+import lime.media.openal.AL;
 import lime.media.openal.ALBuffer;
 import lime.media.openal.ALSource;
 import lime.utils.ArrayBufferView;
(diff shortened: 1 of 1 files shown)


SomeGuyWhoLovesCoding:
Looks good! It's pretty much ready to merge now.

SomeGuyWhoLovesCoding:
> Looks good! It's pretty much ready to merge now.

Also you're not missing anything

PR #2700, opened 2024-03-19
Adjust rendered DisplayObjects' dimensions with filters

NexIsDumb:
This is has been an old bug;
For example if you put one or more filters on a Flixel's FlxCamera (that for rendering Shaders, and so Filters, uses an OpenFL's Sprite) on certain screens you can easily notice that the camera gains +1 pixel on the right and bottom side (and so it's oblivious that this doesn't happen on FlxCameras only but on every OpenFL's Rendered DisplayObjects).
Me and @NeeEoo found an easy way to fix this on certain resolutions which comprehends a lot of screen sizes actually (still testing on FlxCameras), but we feel like there could be a more dynamic fix so this is going to stay a draft until we find out a better fix.
Here's an example of two overlapped cameras with the bottom one using filters (filters on that camera'a OpenFL Sprite) and the top one not using them but having a black sprite that uses the camera's dimensions (better visible on larger screens or zooming): 
Screenshot_20240319-135118_Gallery
As you can see the bottom camera with filters has one more pixel on bottom and right side than the one without filters.
This disappears with this Pull Request's fix.

### DIFF
--- a/src/openfl/display/DisplayObjectRenderer.hx
+++ b/src/openfl/display/DisplayObjectRenderer.hx
@@ -370,8 +370,8 @@ class DisplayObjectRenderer extends EventDispatcher
 
 				displayObject.__getFilterBounds(rect, displayObject.__cacheBitmapMatrix);
 
-				filterWidth = rect.width > 0 ? Math.ceil(rect.width * pixelRatio) : 0;
-				filterHeight = rect.height > 0 ? Math.ceil(rect.height * pixelRatio) : 0;
+				filterWidth = rect.width > 0 ? Math.floor(rect.width * pixelRatio) : 0;
+				filterHeight = rect.height > 0 ? Math.floor(rect.height * pixelRatio) : 0;
 
 				offsetX = rect.x > 0 ? Math.ceil(rect.x) : Math.floor(rect.x);
 				offsetY = rect.y > 0 ? Math.ceil(rect.y) : Math.floor(rect.y);
--- a/src/openfl/display/DisplayObjectRenderer.hx
+++ b/src/openfl/display/DisplayObjectRenderer.hx
@@ -370,8 +370,8 @@ class DisplayObjectRenderer extends EventDispatcher
 
 				displayObject.__getFilterBounds(rect, displayObject.__cacheBitmapMatrix);
 
-				filterWidth = rect.width > 0 ? Math.ceil((rect.width + 1) * pixelRatio) : 0;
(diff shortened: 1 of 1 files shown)


joshtynjala:
> Here's an example of two overlapped cameras with the bottom one using filters

Please post the code for this example. Thanks!

NexIsDumb:
> > Here's an example of two overlapped cameras with the bottom one using filters
> 
> Please post the code for this example. Thanks!

Sure thing, here's WITHOUT the fix (the original code was tested on [CodenameEngine](https://github.com/FNF-CNE-Devs/CodenameEngine), so sorry for the delay I had to rewrite it so you would be able to use it) (Also one last reminder: I'm using FlxCamera since for filters it uses an OpenFL normal Sprite):
```haxe
package;

import flixel.FlxCamera;
import flixel.FlxG;
import flixel.FlxGame;
import flixel.FlxState;
import flixel.system.FlxAssets.FlxShader;
import openfl.display.Sprite;
import openfl.filters.ShaderFilter;

class Main extends Sprite
{
    public function new()
    {
        super();
        addChild(new FlxGame(1280, 720, PlayState));
    }
}

class PlayState extends FlxState
{
    override public function create()
    {
        super.create();

        FlxG.camera.bgColor = 0xFF0000FF;
        FlxG.camera.filters = [new ShaderFilter(new FlxShader())];

        var cam2:FlxCamera = new FlxCamera();
        cam2.bgColor = 0xFFFF0000;
        FlxG.cameras.add(cam2, false);
    }
}
```

And this is what I noticed when resizing to big window mode:
image

If instead I include this Pull Request's fix that +1 pixel is gone.

NexIsDumb:
Oops, had to edit, attached the wrong image

joshtynjala:
Thanks!

NexIsDumb:
With this the problem shouldn't be happening again because it's going to round for defect

NexIsDumb:
just curiosity, any updates on this?

joshtynjala:
It's unfortunate that the commit that added the `+ 1` doesn't include a reference to an issue or forum thread to indicate what it was trying to fix.

70c804dc0722bb741c6c928c0f99f968a502e6ad

I'd personally be okay removing the `+ 1`, but I think that `Math.ceil()` is still a better choice compared to `Math.floor()` here.

NexIsDumb:
hmm.. alrightt, i'll run more tests to see

and besides i dont think any issue was made about this in the repository, i'll create one if i have to

Issue #1724, opened 2017-09-25
Add support for `stage.mouseLock`


jgranick:
Lime has support for native platform mouse locking already (using `lime.ui.Mouse.lock`), so it would be great to wire this up with `stage.mouseLock` support, and relative `MouseEvent` `movementX/movementY` events

PR #2702, opened 2024-04-15
Implement `stage.mouseLock` and `MouseEvent.movement` properties

ShaharMS:
Mouse locking is already implemented in lime. This this pull requests implements mouse locking similarly to flash, as in:

 - Locking is only available in desktop targets and only when in full screen
   - On non-desktop targets, the `mouseLock` property is not available
   - When not in full-screen, setting `mouseLock` to `true` throws `IllegalOperationError("Property can not be set in non full screen mode.");`
 - When the mouse is locked `MouseEvent`'s `stageX/Y` and `localX/Y` are equal to 0, and `movementX/Y` reports differences between previous and current mouse position.
 - Locking is temporarily disabled when the window loses focus, or exits fullscreen
   - When re-entering both of the states above, mouse locking is re-enabled
 
### Implementation details:

under the hood, there's a less sensitive mouseLock property named `__mouseLock`. It is not changed when the user forces mouse unlocking, and is used for when the user temporarily disables mouse locking using one of the methods above. It is not used when retrieving `stage.mouseLock`, so it is possible that `stage.mouseLock` gets changed with user interaction, when the developer explicitly set it to `true`. The only property that the developer has full control of is `__mouseLock`, and it is the property deciding when to re-enable `window.mouseLock`.

Mouse events switch to reporting movement based coordinates only when the mouse is visually locked, so setting `stage.mouseLock` to `true` and receiving non-0 `stageX` property is possible, for example, but, this only happens when mouse locking is temporarily off. We know mouse locking is temporarily off when `__mouseLock && !window.mouseLock`

mouse movement details are retrieved from `__onLimeMouseMoveRelative` event listener

### DIFF
--- a/src/openfl/display/Stage.hx
+++ b/src/openfl/display/Stage.hx
@@ -946,8 +946,8 @@ class Stage extends DisplayObjectContainer #if lime implements IModule #end
 	#end
 	#if desktop
 	@:noCompletion private var __mouseLock:Bool;
-	@:noCompletion private var movementMouseX:Float;
-	@:noCompletion private var movementMouseY:Float;
+	@:noCompletion private var __movementX:Float;
+	@:noCompletion private var __movementY:Float;
 	#end
 
 	#if openfljs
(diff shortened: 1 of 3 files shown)


ShaharMS:
Closes #1724

PR #2703, opened 2024-04-22
Dispose of textures cleaned by the garbage collector

barisyild:
This commit will make reduce the possibility of VRAM memory leak.


barisyild:
It seems that there are possibilities that some OpenGL textures are shared, in which case it is necessary to directly dispose of native textures when they are collected by the garbage collector.

barisyild:
It seems that assets that are no longer referenced but still binded textures in OpenGL cause corruption.

barisyild:
Looks like lime already manages opengl objects with finalizer so this pull request is dead.

https://github.com/openfl/lime/blob/bb3b31b08d75c0b9aa2dc1f3ab5f4539737f4d57/project/src/graphics/opengl/OpenGLBindings.cpp#L155

PR #2709, opened 2024-05-26
 Improve the rendering performance of Tilemap

rainyt:
Improve the rendering performance of Tilemap to avoid reconstructing VertexBufferData per frame from unmodified Tilemap objects.


rainyt:
200 tilemaps with no change:
image

200 tilemaps with change:
image

dimensionscape:
can you resolve the conflicts and pr to 9.5?

rainyt:
Yes, But it seems a bit incompatible with multiple textures, I need time to take a look again.

PR #2711, opened 2024-06-03
Make ``DrawCommandType`` an abstract Int enum

MaybeMaru:
By making it an integer it makes sure the compiler doesnt turn it into a dynamic enum, improving the performance of switch statements like the ones in ``DrawCommandReader``


SomeGuyWhoLovesCoding:
Now do that for all enums

joshtynjala:
I'd lean toward not merging this. However, if someone can create a benchmark that shows that this change makes an impactful difference in performance, I'd reconsider. My gut feeling is that this change won't make a measurable difference, but I'd be very happy to be proven wrong.

nanjizal:
Currently enums in Haxe are normally heavy and slow, so this is a good move.
abstract enums are much faster, as there is lots of AST magic where results are resolved to values rather than using more complex structures. But I would forget the less than 4 as that is largely irrelevant being too much in the past.
I have some support for similar.
https://github.com/nanjizal/justPath/blob/master/justPath/PathCommand.hx
enums are theortically cool but in real use they often suck for performance, abstract enums very limited compared,  but they almost disappear with ast optimization.

nanjizal:
an abstract enum to the compiler is normally just a basic value so can be replaced in place rather than requiring a look up of a structure? Or my understanding atleast.

joshtynjala:
> enums in Haxe are normally heavy and slow, so this is a good move

Maybe. I suspect that the difference is insignificant the vast majority of the time. I still want to see a benchmark that shows this change actually makes things measurably faster before I even consider merging.

nanjizal:
Well largely the point is haxe can replace the abstract enum with a number at runtime, and for low level drawing speed is important, rather than game logic where it may not alway be as important. 
Take a simple case for JS since that is easy to look at generated source.
https://try.haxe.org/#9582b2c5
```haxe
enum ABC {
  A;
  B;
  C;
}

enum abstract ABC_( Int ) {
  var A;
  var B;
  var C;
}
class Test {
  static function main() {
    trace("Haxe is great!");
    //trace( ABC.A );
    trace( ABC_.A );
  }
}
```
so for abstract enum the trace is just an integer with no generation overhead.
```JS
// Generated by Haxe 4.3.6
(function ($global) { "use strict";
class Test {
	static main() {
		console.log("Test.hx:15:","Haxe is great!");
		console.log("Test.hx:17:",0);
	}
}
class haxe_iterators_ArrayIterator {
	constructor(array) {
		this.current = 0;
		this.array = array;
	}
	hasNext() {
		return this.current < this.array.length;
	}
	next() {
		return this.array[this.current++];
	}
}
{
}
Test.main();
})({});
```
But for the current setup..
```JS
// Generated by Haxe 4.3.6
(function ($global) { "use strict";
var $estr = function() { return js_Boot.__string_rec(this,''); },$hxEnums = $hxEnums || {},$_;
Math.__name__ = true;
var ABC = $hxEnums["ABC"] = { __ename__:true,__constructs__:null
	,A: {_hx_name:"A",_hx_index:0,__enum__:"ABC",toString:$estr}
	,B: {_hx_name:"B",_hx_index:1,__enum__:"ABC",toString:$estr}
	,C: {_hx_name:"C",_hx_index:2,__enum__:"ABC",toString:$estr}
};
ABC.__constructs__ = [ABC.A,ABC.B,ABC.C];
class Test {
	static main() {
		console.log("Test.hx:15:","Haxe is great!");
		console.log("Test.hx:16:",ABC.A);
	}
}
Test.__name__ = true;
class haxe_iterators_ArrayIterator {
	constructor(array) {
		this.current = 0;
		this.array = array;
	}
	hasNext() {
		return this.current < this.array.length;
	}
	next() {
		return this.array[this.current++];
	}
}
```
You could try adding some switch statements... I would be inclined to just change it and see who complains! 
However although frowned on I would put 'to' and 'from' to save compile errors you can always consider removal in the future but for non breaking changes in areas like switch statements probably worth it.
```Haxe

enum ABC {
  A;
  B;
  C;
}

enum abstract ABC_( Int ) from Int to Int {
  var A;
  var B;
  var C;
}
class Test {
  static function main() {
    trace("Haxe is great!");
    //trace( ABC.A );
    trace( ABC_.A );
  }
}
```

nanjizal:
I guess with an array / for loop you could test timings with a timing macro. But it is obviously less overhead and a small worthwhile win.

joshtynjala:
It may not be worthwhile, which is why I’m continuing to ask for a performance benchmark showing an improvement in OpenFL drawing performance. It’s quite possible that this PR will make no significant difference without billions of drawing commands (assuming other factors don’t affect performance long before this would, which is very likely). If it’s not measurablely better, I’m inclined to make no changes because they’ll feel like making changes only for the sake of some kind of unnecessary purity rather than for practical reasons.

On the other hand, if there’s a nice little boost, then absolutely, I’m happy to merge. 

I think I’ve made my thoughts on the matter clear. Any further discussion of theoretical performance is a waste of time. I want to see some real code that can demonstrate a performance win.

PR #2715, opened 2024-06-14
Shader system improvements (redo)

EliteMasterEric:
An update of #2577 which removes extraneous changes and updates the target branch to `9.4.0-Dev`.

These changes were performed to support fixes for FlxRuntimeShader, as part of pull requests HaxeFlixel/flixel-addons#368 and HaxeFlixel/flixel-addons#399. Making changes here prevents code duplication in FlxRuntimeShader, improving maintainability.

These changes have been tested with Friday Night Funkin', where it is used on Windows, MacOS, Linux, and HTML5 platforms.

### DIFF
--- a/src/openfl/utils/_internal/ShaderMacro.hx
+++ b/src/openfl/utils/_internal/ShaderMacro.hx
@@ -469,20 +469,10 @@ class ShaderMacro
 		switch (glVersion)
 		{
 			default:
-				// We don't know this GLSL version, so don't do anything.
+				// Don't make any changes to undefined versions.
 				return source;
 
-			case "100":
-				return source;
-			case "110":
-				return source;
-			case "120":
-				return source;
-			case "130":
-				return source;
-			case "140":
-				return source;
-			case "150":
+			case "100", "110", "120", "130", "140", "150":
 				return source;
 
 			case "300 es":
@@ -501,41 +491,20 @@ class ShaderMacro
 				result = glFragColorKeyword.replace(result, "fragColor");
 				return result;
 
-			case "310 es":
+			case "310 es", "320 es":
 				var result = processGLSLText(source, "300 es", isFragment);
-				return result;
-
-			case "320 es":
-				var result = processGLSLText(source, "310 es", isFragment);
 				return result;
 
 			case "330":
 				#if desktop
-				var result = processGLSLText(source, "320 es", isFragment);
+				var result = processGLSLText(source, "300 es", isFragment);
 				#else
 				var result = source;
 				#end
 				return result;
-			case "400":
+
+			case "400", "410", "420", "430", "440", "450", "460":
 				var result = processGLSLText(source, "330", isFragment);
-				return result;
-			case "410":
-				var result = processGLSLText(source, "400", isFragment);
-				return result;
-			case "420":
-				var result = processGLSLText(source, "410", isFragment);
-				return result;
-			case "430":
-				var result = processGLSLText(source, "420", isFragment);
-				return result;
-			case "440":
-				var result = processGLSLText(source, "430", isFragment);
-				return result;
-			case "450":
-				var result = processGLSLText(source, "440", isFragment);
-				return result;
-			case "460":
-				var result = processGLSLText(source, "450", isFragment);
 				return result;
 		}
 	}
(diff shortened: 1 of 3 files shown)


src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +436,211 @@ class ShaderMacro
...
+				return glExtensions;
+			case "430":
+				return glExtensions;
+			case "440":
+				return glExtensions;
+			case "450":
+				return glExtensions;
+			case "460":
+				return glExtensions;
+		}
```
Geokureli: Pet-peeve, but I still don't get why this needs to be so overly complicated
```suggestion
		switch (glVersion)
		{
			case "100", "110", "120", "130", "140", "150", "400", "410", "420", "430", "440", "450", "460":
				return glExtensions;
			default: 
				return glExtensions;

			case "300 es":
				var hasSeparateShaderObjects = false;
				for (extension in glExtensions)
				{
					if (extension.name == "GL_ARB_separate_shader_objects") hasSeparateShaderObjects = true;
					if (extension.name == "GL_EXT_separate_shader_objects") hasSeparateShaderObjects = true;
				}

				#if desktop
				if (!hasSeparateShaderObjects)
				{
					#if linux
					return glExtensions.concat([{name: "GL_EXT_separate_shader_objects", behavior: "require"}]);
					#else
					return glExtensions.concat([{name: "GL_ARB_separate_shader_objects", behavior: "require"}]);
					#end
				}
				#end
				return glExtensions;

			case "310 es":
				return buildGLSLExtensions(glExtensions, "300 es", isFragment);
			case "320 es":
				return buildGLSLExtensions(glExtensions, "310 es", isFragment);
			case "330":
				return buildGLSLExtensions(glExtensions, "320 es", isFragment);
		}
```

src/openfl/utils/_internal/ShaderMacro.hx
```diff
@@ -199,6 +436,211 @@ class ShaderMacro
...
+			case "440":
+				var result = processGLSLText(source, "430", isFragment);
+				return result;
+			case "450":
+				var result = processGLSLText(source, "440", isFragment);
+				return result;
+			case "460":
+				var result = processGLSLText(source, "450", isFragment);
+				return result;
+		}
```
Geokureli: Same here:
```suggestion
return switch (glVersion)
		{
			case "100", "110", "120", "130", "140", "150": source;
			default: source;

			case "300 es":
				var result = source;
				// Migrate, replacing "attribute" with "in" and "varying" with "out".
				if (isFragment)
				{
					result = varyingKeyword.replace(result, "in $1 $2");
				}
				else
				{
					result = attributeKeyword.replace(result, "in $1 $2");
					result = varyingKeyword.replace(result, "out $1 $2");
				}
				result = texture2DKeyword.replace(result, "texture");
				result = glFragColorKeyword.replace(result, "fragColor");
				return result;

			case "310 es": processGLSLText(source, "300 es", isFragment);
			case "320 es": processGLSLText(source, "310 es", isFragment);
			#if desktop
			case "330": processGLSLText(source, "320 es", isFragment);
			#else
			case "400": processGLSLText(source, "330", isFragment);
			case "410": processGLSLText(source, "400", isFragment);
			case "420": processGLSLText(source, "410", isFragment);
			case "430": processGLSLText(source, "420", isFragment);
			case "440": processGLSLText(source, "430", isFragment);
			case "450": processGLSLText(source, "440", isFragment);
			case "460": processGLSLText(source, "450", isFragment);
		}
```

Geokureli: can probably simplify further with 
```hx
case "400", "410", "420", "430", "440", "450", "460":
	processGLSLText(source, "330", isFragment);
```

src/openfl/utils/_internal/ShaderMacro.hx
Geokureli: this also seems functionally identical to 
```hx
switch (glVersion)
{
	default: 
		return glExtensions;

	#if desktop
	case "300 es", "310 es", "320 es", "330":
		var hasSeparateShaderObjects = false;
		for (extension in glExtensions)
		{
			if (extension.name == "GL_ARB_separate_shader_objects") hasSeparateShaderObjects = true;
			if (extension.name == "GL_EXT_separate_shader_objects") hasSeparateShaderObjects = true;
		}

		if (!hasSeparateShaderObjects)
		{
			#if linux
			return glExtensions.concat([{name: "GL_EXT_separate_shader_objects", behavior: "require"}]);
			#else
			return glExtensions.concat([{name: "GL_ARB_separate_shader_objects", behavior: "require"}]);
			#end
		}
		return glExtensions;
	#end
}
```

EliteMasterEric: I think it was only worded like this as, while testing, I anticipated requiring more version-specific changes, but ended up only needing a couple. I will definitely rewrite this to match your proposed syntax.

EliteMasterEric:
@Geokureli Let me know if you have any other syntax changes to make, I implemented the ones you suggested.

PR #2722, opened 2024-07-22
Shader system improvements (redo)

EliteMasterEric:
An update of #2715 which updates the target branch to `develop`.

These changes were performed to support fixes for FlxRuntimeShader, as part of pull requests HaxeFlixel/flixel-addons#368 and HaxeFlixel/flixel-addons#399. Making changes here prevents code duplication in FlxRuntimeShader, improving maintainability.

These changes have been tested with Friday Night Funkin', where it is used on Windows, MacOS, Linux, and HTML5 platforms.

### DIFF
--- a/src/openfl/display/GraphicsShader.hx
+++ b/src/openfl/display/GraphicsShader.hx
@@ -8,7 +8,6 @@ import openfl.utils.ByteArray;
 #end
 class GraphicsShader extends Shader
 {
-	@:glVersion("120")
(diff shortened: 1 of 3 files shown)


ninjamuffin99:
Can confirm that our usage in it for FNF has gone well and it's been pretty battle tested!

EliteMasterEric:
We should probably re-review this soon, we've been using this in Funkin for like a year now with no issues.

EDIT: Aside from the merge conflict which I can try to resolve locally.

PR #2730, opened 2024-08-11
OpenGLRenderer: Avoid duplicate construction of shaders

rainyt:
When creating BitmapData.disposeImage and using draw, repeatedly constructing OpenGLrenderer will result in the creation of a new shader, which incurs performance costs for multiple shader compilations.


joshtynjala:
To be included in OpenFL 9.5 with commit bd45fc0235474bdd6f3d36adcc64e4829d421cec

PR #2731, opened 2024-08-13
BitmapData: Dispose BaseTexture if not shared

dimensionscape:
Ideally we dispose a texture if its not being shared.

We can reasonably predict when a BitmapData's Textured is shared or not and dispose of it. Typically this means we called getTexture(), clone() or fromTexture().

With this change, we track when the Texture is being shared and then proceed to call dispose on the texture when the BitmapData is disposed.

I'm not 100% fond of this solution but we need to prevent the memory leak caused by the texture failing to be disposed.

Open to other ideas or solutions.

### DIFF
--- a/src/openfl/display/BitmapData.hx
+++ b/src/openfl/display/BitmapData.hx
@@ -406,7 +406,12 @@ class BitmapData implements IBitmapDrawable
 
 	/**
 		Returns a new BitmapData object that is a clone of the original instance with an exact copy of the contained bitmap.
-		@return		A new BitmapData object that is identical to the original.
+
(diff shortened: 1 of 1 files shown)


joshtynjala:
We should also add some documentation to `getTexture()` to indicate that calling it marks the texture as shared, and that the developer is required to dispose the texture manually when finished.

I would suggest putting this in 9.5.0-dev instead of develop, since we are very close to release for 9.4, and it won't get a ton of testing before then. Even if I don't think this change is super risky, I always try to be a little extra careful near release time.

dimensionscape:
Oh yeah, no rush, just wanted to get the conversation going.

PR #2732, opened 2024-08-13
Bugfix/dispose texture

dimensionscape:
Continuing this discussion from https://github.com/openfl/openfl/pull/2731
> 
> Ideally we dispose a texture if its not being shared.
> 
> We can reasonably predict when a BitmapData's Textured is shared or not and dispose of it. Typically this means we called getTexture(), clone() or fromTexture().
> 
> With this change, we track when the Texture is being shared and then proceed to call dispose on the texture when the BitmapData is disposed.
> 
> I'm not 100% fond of this solution but we need to prevent the memory leak caused by the texture failing to be disposed.
> 
> Open to other ideas or solutions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -171,8 +171,7 @@ jobs:
   unit-test-air:
     strategy:
       matrix:
-        # until Lime 8.0.3 is release, we can't test Haxe 3.4.7 with AIR
-        haxe-version: [4.0.5, 4.1.5, 4.2.5, 4.3.5]
+        haxe-version: [3.4.7, 4.0.5, 4.1.5, 4.2.5, 4.3.5]
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


LeonGamerPS1:
fuck this! DONT MERGE.

joshtynjala:
I'm okay if we close this. I originally thought (incorrectly) that the Haxe object would be GCed, but the texture would remain in GPU memory indefinitely. However, since we (re-)discovered that Lime's C++ code smartly disposes the GPU texture automatically when the Haxe object is GCed, it's not quite as vital that this gets fixed.

To be clear, I still strongly believe that calling `dispose()` on `BitmapData` should clear all memory, both the software pixel data and the GPU texture, to better align with how Flash developers have long understood the purpose of that method. However, but I'm not going to push hard for it at this time. That being said, I think we should aim to improve the design of the APIs used for sharing bitmap data textures in a future major version of OpenFL so that we can eventually make `dispose()` clear both software and hardware memory.

PR #2741, opened 2024-10-04
Make `curveTo()` and `cubicCurveTo()` update drawing position.

### LINKED ISSUE #2740: Graphics: calling `lineStyle()` after `curveTo()` reverts drawing position
**Describe the bug**
If you call `curveTo()`, the next line you draw starts from the end of the curve. However, if you call `curveTo()` followed by `lineStyle()`, the next line you draw starts from the _beginning_ of the curve.

**To Reproduce**
Create a new project and insert this code:

```haxe
graphics.lineStyle(1, 0x000000);
graphics.curveTo(100, 0, 100, 100);
graphics.lineStyle(1, 0x000000); //<-- problem
graphics.lineTo(200, 100);
```

Both the curve and the line will be drawn starting from the origin:

Actual result

**Expected behavior**
The line should start from the end of the curve, as happens when you don't call `lineStyle()` in between.

```haxe
graphics.lineStyle(1, 0x000000);
graphics.curveTo(100, 0, 100, 100);
graphics.lineTo(200, 100);
```

Expected behavior

**OpenFL Targets**
Reproduced on C++, HL, Neko, and HTML5, all on Linux.

player-03:
Fixes #2740.

```haxe
graphics.lineStyle(1, 0x000000);

graphics.curveTo(100, 0, 100, 100);
graphics.lineStyle(1, 0x000000);
graphics.lineTo(200, 100);

graphics.cubicCurveTo(250, 100, 300, 150, 300, 200);
graphics.lineStyle(1, 0x000000);
graphics.lineTo(400, 200);
```

This code is supposed to draw alternating curves and horizontal lines, like these:

Expected

But without this fix, `lineStyle()` reset the drawing position to where it was before the previous curve. This means the first `lineTo()` starts from (0, 0) and the second starts from the end of the first:

Actual

### DIFF
--- a/src/openfl/display/_internal/CairoGraphics.hx
+++ b/src/openfl/display/_internal/CairoGraphics.hx
@@ -533,10 +533,16 @@ class CairoGraphics
 						c.anchorY
 						- offsetY);
 
+					positionX = c.anchorX;
+					positionY = c.anchorY;
+
 				case CURVE_TO:
 					var c = data.readCurveTo();
 					hasPath = true;
 					quadraticCurveTo(c.controlX - offsetX, c.controlY - offsetY, c.anchorX - offsetX, c.anchorY - offsetY);
+
+					positionX = c.anchorX;
+					positionY = c.anchorY;
 
 				case DRAW_CIRCLE:
 					var c = data.readDrawCircle();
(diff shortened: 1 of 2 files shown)


joshtynjala:
Thanks! I'll keep this in mind for merging after 9.4 is released, to be included in 9.4.1, since I don't want to risk introducing any unexpected bugs right before the upcoming release.

PR #2744, opened 2024-10-14
Add System.totalMemoryNumber

ACrazyTown:
This PR adds the `openfl.system.System.totalMemoryNumber` field to allow getting a 64bit memory usage value. If the target does not support getting a higher precision memory usage value, it will simply return `totalMemory`

### DIFF
--- a/src/openfl/system/System.hx
+++ b/src/openfl/system/System.hx
@@ -80,7 +80,20 @@ import hl.Gc;
 	**/
 	public static var totalMemory(get, never):Int;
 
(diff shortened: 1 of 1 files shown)


ACrazyTown:
BTW I just realized I forgot to strip the path here: https://github.com/openfl/openfl/pull/2744/commits/e94cd631005fa27445e498a21e401803b44bf932#diff-4abe1af825f9cd4fd6505b3bbb60038b037029cd96b121c572473ab3af1fa63aR296, it can be removed since the cpp Gc is already imported

PR #2745, opened 2024-10-16
Tracy profiler support added

barisyild:
https://github.com/HaxeFoundation/hxcpp/pull/1153


EliteMasterEric (approved):

EliteMasterEric:
This should be stable to add, since it's locked behind a compile define.

PR #2748, opened 2024-11-20
Remove textfield cache immediately.

barisyild:
This pull request specifically affects the html5 target.

Since the html5 target is constantly creating textures, the memory starts to swell and forces the application to major gc.

Reproduce Steps
- Run Code Below
```haxe
var textField:TextField = new TextField();
addChild(textField);

openfl.Lib.setInterval(() -> {
	var str:String = "";
	for(i in 0...10000)
	{
		str += Std.random(9);

		if(i % 100 == 0)
		{
			str += "\n";
		}
	}
	textField.text = str;
	textField.width = textField.textWidth;
	textField.height = textField.textHeight;
}, 0);
```
- Open Task Manager in a Chromium-based browser (SHIFT + ESC)
- Observe the memory footprint used by the GPU Process.

Animation

### DIFF
--- a/src/openfl/display/_internal/CairoTextField.hx
+++ b/src/openfl/display/_internal/CairoTextField.hx
@@ -158,8 +158,8 @@ class CairoTextField
 			// Clear old cache immediately.
 			if(graphics.__bitmap != null)
 			{
-				if(graphics.__bitmap._texture != null)
-					graphics.__bitmap._texture.dispose();
+				if(graphics.__bitmap.__texture != null)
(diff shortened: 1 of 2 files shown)


joshtynjala:
The disposal of the texture will be included in OpenFL 9.5. I've added you as a co-author of commit 35151f3c5d70accb5a46dc2e46e49d74f709abdf.

I also added an additional optimization here that you may be interested in: 5f09231abc9b3721c390398e80193b5664a5324d

PR #2755, opened 2024-12-11
Fix WebKit Bug 195325: Clear and reset canvas/context to prevent memo…

NipponSunrise:
This PR fixes the issue described in [WebKit Bug 195325](https://bugs.webkit.org/show_bug.cgi?id=195325). While the bug was originally reported on iOS 12, we have observed the same behavior on iOS 15.8 on the `html5` target.

The issue involves properly clearing and resetting canvas and context references to prevent memory leaks. The solution ensures that both the canvas and its context are explicitly reset and nullified.

Key changes:
- Explicitly set the canvas width and height to `0` before nullifying the canvas reference.
- Call `clearRect(0, 0, 0, 0)` on the context to ensure any lingering state is cleared before nullifying the context reference.

Tested on iOS 15.8 and confirmed that the issue is resolved.

If there's a need for further clarification or adjustments, let me know!


rainyt:
Is there any documentation regarding this repair? Does setting the width and height help with this?

NipponSunrise:
> Is there any documentation regarding this repair? Does setting the width and height help with this?

As outlined in the PR description, the approach is based on observations from WebKit Bug 195325 and has been tested on iOS 
15.8. If you have any specific concerns beyond what has already been explained, please clarify.

joshtynjala:
I need to finish up some work for my clients before the end of the month, but I should have time to review this PR next week.

joshtynjala:
I am able to reproduce this issue using the following code on an iPhone 6 running iOS 12.5.7 and an iPhone SE running iOS 18.0:

```haxe
var sprites:Array<Sprite> = [];

function createSprite():Void {
  while (sprites.length > 0) {
    var sprite = sprites.shift();
    removeChild(sprite);
  }
  sprites = [];
  for (i in 0...20) {
    var sprite = new Sprite();
    sprite.graphics.beginFill(0xff0000);
    sprite.graphics.drawCircle(400, 400, 400);
    sprite.graphics.endFill();
    sprite.x = 50 + i;
    sprite.y = 50 + i;
    addChild(sprite);
    sprites.push(sprite);
  }
}

var tf = new TextField();
tf.autoSize = LEFT;
tf.border = true;
tf.backgroundColor = 0xeeeeee;
tf.background = true;
tf.selectable = false;
tf.text = "Click Me";
tf.x = 20;
tf.y = 20;
addChild(tf);
tf.addEventListener(MouseEvent.MOUSE_DOWN, event -> {
  tf.backgroundColor = 0xcccccc;
  tf.borderColor = 0x333333;
  function onStageMouseUp(event:MouseEvent):Void {
    cast(event.currentTarget, Stage).removeEventListener(MouseEvent.MOUSE_UP, onStageMouseUp);
    tf.backgroundColor = 0xeeeeee;
    tf.borderColor = 0x000000;
  }
  tf.stage.addEventListener(MouseEvent.MOUSE_UP, onStageMouseUp);
});
tf.addEventListener(MouseEvent.CLICK, event -> {
  createSprite();
});
```

Click the TextField repeatedly until the tab crashes. For me, it was only twice on both devices.

After applying the changes, the tab stops crashing.

rainyt:
image
Regarding the modification of the BitmapData class, this PR may cause some abnormal drawing issues. In my project, it may result in blank images that cannot be drawn.
wrong:
image

NipponSunrise:
> Regarding the modification of the BitmapData class, this PR may cause some abnormal drawing issues. In my project, it may result in blank images that cannot be drawn. wrong:

Hi @rainyt,

Thanks for reporting this. It's unexpected, since the change only touches what happens inside `dispose()`. After a `BitmapData` **is disposed**, it typically **shouldn't be used anymore**, so I’m curious what kind of usage might be affected here.

Is it possible that something is referencing or reusing the bitmap after disposal? Any logs or minimal code sample would be really helpful to understand the issue.

Looking forward to your input.

joshtynjala:
@rainyt Is there any chance that you're using #2748? I suspect that some of the changes in #2748 may conflict with this one. In particular, the `graphics.__bitmap.dispose()` calls. I think that `graphics.__bitmap.__texture.dispose()` is still safe to call, though.

rainyt:
> @rainyt Is there any chance that you're using #2748? I suspect that some of the changes in #2748 may conflict with this one. In particular, the `graphics.__bitmap.dispose()` calls. I think that `graphics.__bitmap.__texture.dispose()` is still safe to call, though.

I tested this and it is indeed related to  #2748.I can comment out the code related to #2748, and the implementation of BitmapData can work properly. So does the conflict between them need to be optimized? Will the data generated in the text texture be optimized due to this PR? Is the modification of #2748 unnecessary.

rainyt:
wrong:
image

none:
image

But it only solves one of the problems. Here is a graphic drawing implemented as a mask, but now the rendering is not working properly.

rainyt:
I can't create a small case to reproduce it, it's a strange problem.

joshtynjala:
> So does the conflict between them need to be optimized? Will the data generated in the text texture be optimized due to this PR? Is the modification of https://github.com/openfl/openfl/pull/2748 unnecessary.

You must remove `graphics.__bitmap.dispose()` from the changes in #2748. You should not dispose the BitmapData because it will clear/reset the Canvas that is shared by both the original BitmapData and the new BitmapData. However, you should be able to keep `graphics.__bitmap.__texture.dispose()` because the new BitmapData will create a new texture, and there is still a benefit to disposing the old texture.

rainyt:
> You must remove `graphics.__bitmap.dispose()` from the changes in #2748. You should not dispose the BitmapData because it will clear/reset the Canvas that is shared by both the original BitmapData and the new BitmapData. However, you should be able to keep `graphics.__bitmap.__texture.dispose()` because the new BitmapData will create a new texture, and there is still a benefit to disposing the old texture.

Yes, fixed.

PR #2756, opened 2024-12-15
Fix access violation on HL when using hardware only bitmaps and shader fills

ACrazyTown:
When using hardware only bitmaps and shader fills, previously it tried to pass it to `createImagePattern`, which would end up passing `null` to the native binding code and cause an access violation

Repro:
```hx
package;

{
	public function new() 
	{
		super();

		var shader = new openfl.display.GraphicsShader();
		
		var bmd = new openfl.display.BitmapData(100, 100, true, 0xFF3F64CC);

		bmd.disposeImage();

		shader.bitmap.input = bmd;

		graphics.beginShaderFill(shader);
		graphics.drawRect(0, 0, stage.stageWidth, stage.stageHeight);
		graphics.endFill();
	}
}

```

### DIFF
--- a/src/openfl/display/_internal/CairoGraphics.hx
+++ b/src/openfl/display/_internal/CairoGraphics.hx
@@ -781,10 +781,21 @@ class CairoGraphics
 
 					if (shaderBuffer.inputCount > 0)
 					{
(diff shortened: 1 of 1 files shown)


joshtynjala:
Thanks!

PR #2761, opened 2024-12-19
Instead of constantly allocating the buffer, just update it if the byteLength is the same.

barisyild:
If the buffer is allocated and the size of the new buffer to be allocated is the same, update the buffer in memory instead of allocating a new one.

### DIFF
--- a/src/openfl/display3D/VertexBuffer3D.hx
+++ b/src/openfl/display3D/VertexBuffer3D.hx
@@ -60,6 +60,7 @@ class VertexBuffer3D
 	@:noCompletion private var __tempFloat32Array:Float32Array;
 	@:noCompletion private var __usage:Int;
 	@:noCompletion private var __vertexSize:Int;
+	@:noCompletion private var __byteLength:Int;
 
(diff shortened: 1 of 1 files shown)


joshtynjala:
I made a little benchmark to test this PR (similar to BunnyMark, but with some tweaks to ensure that this code gets run frequently every frame). In my benchmark, this change results in consistently lower FPS on macOS (I saw lower FPS in html5 using Firefox, and also in hxcpp). I didn't try on Windows or Linux, though.

Admittedly, it's kind of an artificial benchmark that doesn't necessarily replicate real-world use of OpenFL. If you can post your own sample project where this change helps rather than hurts, I'd like to see it. Thanks!

barisyild:
> I made a little benchmark to test this PR (similar to BunnyMark, but with some tweaks to ensure that this code gets run frequently every frame). In my benchmark, this change results in consistently lower FPS on macOS (I saw lower FPS in html5 using Firefox, and also in hxcpp). I didn't try on Windows or Linux, though.
> 
> Admittedly, it's kind of an artificial benchmark that doesn't necessarily replicate real-world use of OpenFL. If you can post your own sample project where this change helps rather than hurts, I'd like to see it. Thanks!

I remember it giving better results in the Bunnymark test.

PR #2764, opened 2025-01-08
Add pretty printing to shader errors and info

### LINKED ISSUE #2763: Improve shader error logging
Take an example shader error log:
```
Uncaught exception: [openfl.display.Shader] ERROR: Error compiling fragment shader
ERROR: 0:58: '>' does not operate on 'float' and 'int'

#ifdef GL_ES
#ifdef GL_FRAGMENT_PRECISION_HIGH
precision highp float;
#else
precision mediump float;
#endif
#endif

// <large section removed for brevity>

        uniform bool _enabled;
        uniform float inSize;
        uniform float outSize;
        uniform vec4 inColor;
        uniform vec4 outColor;

        bool outline(vec2 coord, float size)
        {
            float w = size / openfl_TextureSize.x;
            float h = size / openfl_TextureSize.y;
            
            return size > 0
            &&  ( flixel_texture2D(bitmap, vec2(coord.x + w, coord.y)).a != 0.0
                || flixel_texture2D(bitmap, vec2(coord.x - w, coord.y)).a != 0.0
                || flixel_texture2D(bitmap, vec2(coord.x, coord.y + h)).a != 0.0
                || flixel_texture2D(bitmap, vec2(coord.x, coord.y - h)).a != 0.0
                );
        }
        
        void main()
        {
            vec4 color = flixel_texture2D(bitmap, openfl_TextureCoordv);
            if (color.a == 0.0 && _enabled)
            {
                if (outline(openfl_TextureCoordv, inSize))
                    color = inColor;
                else if (outline(openfl_TextureCoordv, outSize))
                    color = outColor;
            }
            gl_FragColor = color;
        }
```

My goal is to reduce this to:

```
Uncaught exception: [openfl.display.Shader] ERROR: Error compiling fragment shader: MyShader
LINE:            return size > 0
ERROR: 0:58: '>' does not operate on 'float' and 'int'
```

Geokureli:
Fixes #2763

Example shader that causes errors:
```hx
package vfx;

class TestShader extends flixel.system.FlxAssets.FlxShader
{
    @:glFragmentSource('
#pragma header

void main()
{
    vec4 color = flixel_texture2D(bitmap, openfl_TextureCoordv);
    
    // 0 should be 0.0
    if (color.a > 0)
    {
        // brg should be vec4
        color = color.brg;
    }
    // color_ should be color
    gl_FragColor = color_;
}
    ')
    public function new()
    {
        super();
    }
}
```

The error message:
```
Uncaught exception: [openfl.display.Shader] ERROR: Error compiling fragment shader: source/vfx/TestShader.hx:3

Line 52: '>' does not operate on 'float' and 'int'
        if (color.a > 0)

Line 55: Incompatible types (vec4 and vec3) in assignment (and no available implicit conversion)
        color = color.brg;

Line 58: Use of undeclared identifier 'color_'
        gl_FragColor = color_;
```

### DIFF
--- a/src/openfl/display/Shader.hx
+++ b/src/openfl/display/Shader.hx
@@ -349,12 +349,8 @@ class Shader
 	@:noCompletion private var __isEmptyLine = ~/^\W*$/;
 	@:noCompletion private function __logGLShaderInfo(isError:Bool, type:Int, infoLog:String, source:String):Void
 	{
-		var message = isError ? "Error" : "Info";
-		var typeName = (type == __context.gl.VERTEX_SHADER) ? "vertex" : "fragment";
-		message += ' compiling $typeName shader';
-		
+		var message = "";
 		var lines = source.split("\n");
-		var prettyLog = "";
 		var failingLine:String = null;
 		for (log in infoLog.split("\n"))
 		{
@@ -376,24 +372,24 @@ class Shader
 			if (lineNumber >= lines.length)
 			{
 				// EOF errors will not have a valid line
-				prettyLog += '\n\n $lineNumber | $info';
+				message += '\n\n $lineNumber | $info';
 			}
 			else
 			{
 				// Add the relevant line to each log
 				var line = lines[lineNumber - 1];
 				var indent = StringTools.lpad("|", " ", lineNumberStr.length + 3);
-				prettyLog += '\n\n $lineNumber | $line\n$indent ${info}';
+				message += '\n\n $lineNumber | $line\n$indent ${info}';
 			}
 		}
 
+		// If we couldn't parse the logs, output the old, verbose format
 		if (failingLine != null)
-			message += '\nFailed to simplify log:"$failingLine"\n$infoLog\n$source';
-		else
-			message += prettyLog;
-
-		if (isError) Log.error(message);
-		else Log.debug(message);
+			message = '\nFailed to simplify log:"$failingLine"\n$infoLog\n$source';
+		
+		var typeName = (type == __context.gl.VERTEX_SHADER) ? "vertex" : "fragment";
+		if (isError) Log.error('Error compiling $typeName shader $message');
+		else Log.debug('Info compiling $typeName shader $message');
 	}
 
 	@:noCompletion private function __createGLProgram(vertexSource:String, fragmentSource:String):GLProgram
--- a/src/openfl/display/Shader.hx
+++ b/src/openfl/display/Shader.hx
@@ -413,13 +413,13 @@ class Shader
 		var linesBack = 0;
 		while (i-- > 0)
 		{
-			if (StringTools.contains(lines[i], "openfl_endregion"))
+			if (lines[i].indexOf("openfl_endregion") != -1)
 				nests++;
 			
 			if (nests == 0)
 				linesBack++;
 			
-			if (StringTools.contains(lines[i], "openfl_region"))
+			if (lines[i].indexOf("openfl_region") != -1)
 			{
 				if (nests == 0)
 					break;
--- a/src/openfl/display/Shader.hx
+++ b/src/openfl/display/Shader.hx
@@ -336,7 +336,17 @@ class Shader
 		return shader;
 	}
 
-	private var __lineExtractor = ~/^\w+?: \d+:(\d+):(.+$)/;
+	/**
+	 * Retrieves the line number from a shader log line.
+	 */
+	@:noCompletion private var __lineExtractor = ~/^\w+?: \d+:(\d+):(.+$)/;
+	/**
+	 * Searches for strings that have only whitespace.
+	 * 
+	 * **Note:** Searching for all whitespace via `~/^\s*$/` caused false-negatives,
+	 * notably: `String.fromCharCode(0)` is `false` but `\W` is `true`.
+	 */
+	@:noCompletion private var __isEmptyLine = ~/^\W*$/;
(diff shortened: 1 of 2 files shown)


Geokureli:
I should change this to print similarly to Haxe's new pretty message reporting, for example :

```
52 |    if (color.a > 0)
   | '>' does not operate on 'float' and 'int'
```

Geokureli:
Does anyone know what the zero in "ERROR: 0:22: '==' does not operate on 'float' and 'int'" refers to?

Geokureli:
I think it might be possible to include the path to the shader's .hx file and give the actual line numbers of each error. Gonna look into that

Edit: never mind

ninjamuffin99:
> Does anyone know what the zero in "ERROR: 0:22: '==' does not operate on 'float' and 'int'" refers to?

Talking to someone who knows a lot more about WebGL / glsl stuff, he just told me it's there all the time and doesn't have much meaning.

> _in opengl, glShaderSource() can set multiple sources by passing an array of strings:_ https://registry.khronos.org/OpenGL-Refpages/gl4/html/glShaderSource.xhtml
> 
> _webgl can just do one at a time:_ https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/shaderSource
> _so maybe its a holdover from an OpenGL feature that just never got ported to WebGL?_
> _wouldnt be the only thing that is parked for a potential future where more opengl features are brought in haha_

small theory crafting from him

Geokureli:
@ninjamuffin99 has convinced me to try to get the filename included in the error

dimensionscape:
> Does anyone know what the zero in "ERROR: 0:22: '==' does not operate on 'float' and 'int'" refers to?

This usually refers to the shader source index, and it's most often 0 if you’re only compiling a single shader source (e.g., just one vertex or fragment shader).

This might seem confusing at first since common sense suggests you cannot have multiple fragment or vertex shaders attached at once doing separate things in the same program, but, when you create a shader with `glCreateShader` you can actually feed it multiple strings and then all those source strings are concatenated in order into one big shader source before compilation. 

.. And that is where the "0" in the error message comes in. It's effectively the index in the source list, and of course it usually 0 because its common to use a single source. In practice, it's quite useful for separation and organization of complex glsl code, especially when you're working on big engines.

Vortex2Oblivion:
Any updates on this?

EliteMasterEric:
I personally found that it doesn't work properly for me on windows. I'm doing more thorough research into it right now.

EDIT: Looks like errors are formatted differently by platform, might be able to just use a `#if` to fix it.
image

src/openfl/display/Shader.hx
```diff
@@ -326,21 +326,72 @@ class Shader
...
+			__logGLShaderInfo(isError, type, shaderInfoLog, source);
 		}
 
 		return shader;
 	}
 
+	/**
+	 * Retrieves the line number from a shader log line.
+	 */
+	@:noCompletion private var __lineExtractor = ~/^\w+?: \d+:(\d+):(.+$)/;
```
EliteMasterEric, line 347: Change this to the following to make it work on Windows:

```haxe
	#if windows
	// 0(5) : whatever
	@:noCompletion private var __lineExtractor = ~/^\d+\((\d+)\) : (.+$)/;
	#else // (html5 || macos || linux)
	// ERROR: 0:5:whatever
	@:noCompletion private var __lineExtractor = ~/^\w+?: \d+:(\d+):(.+$)/;
	#end
```

EliteMasterEric (changes requested):
Managed to fix it on my fork

src/openfl/display/Shader.hx
Geokureli:
```suggestion
	#if windows
	// 0(5) : whatever
	@:noCompletion private var __lineExtractor = ~/^\d+\((\d+)\) : (.+$)/;
	#else // (html5 || macos || linux)
	// ERROR: 0:5:whatever
	@:noCompletion private var __lineExtractor = ~/^\w+?: \d+:(\d+):(.+$)/;
	#end
```
Screenshot 2025-06-10 at 8 56 34 AM

Thanks @EliteMasterEric, btw you can create suggestions via replies

Geokureli:
Added the shader class's filepath to the error. `pos.toLocation()` did not seem to work on the haxe 3.4.7 neko-sample tests, so I simply wrapped the macro lines in `#if haxe4` and ignore `__shaderLocation` if it is null

We could take this a step further and store the lines of each `@:gl` tag as well as info about where each block starts in the source, but I think this is fine, for now

Here's how it looks in vscode:
Screenshot 2025-06-26 at 5 22 44 PM

Geokureli:
I took it a step further and made it show source file and line number on every error 

Screenshot 2025-06-28 at 12 57 43 AM

Geokureli:
My goal with GlMetas was to store all meta info and hopefully generate both the shaders and the source maps for those shaders, but every way i tried to generate source maps looked pretty ugly, and I was already spitting out shader regions, for clarity, so I decided to use those at runtime to to determine the file paths and line line numbers. It would be ideal to generate all that at compile time, so let me know if thats preferred

Additionally, this shows the full filepath for all errors on lines from openfl and flixel classes, i.e.:
`/Users/booboo/Documents/haxe/lib/openfl/openfl/src/openfl/display/GraphicsShader.hx` rather than the preffered `openfl/display/GraphicsShader.hx`

PR #2765, opened 2025-02-03
Update README.md

MrScottyPieey:
Added new consoles


dimensionscape:
We do not currently support any new consoles, but that may change in the future.

PR #2766, opened 2025-02-05
This only gets compiled if on haxe 4 anyways so why check twice

Vortex2Oblivion:
This is only going to get compiled if on haxe 4 anyways so why check twice.

also,
 ```hx
 untyped #if haxe4 js.Syntax.code #else __js__ #end
 ```
 is present a lot in the code.
I feel like it would probably be a good idea to make a function that just returns the proper value depending on the haxe version to make things look a lot neater.

### DIFF
--- a/src/openfl/display/Stage.hx
+++ b/src/openfl/display/Stage.hx
@@ -920,7 +920,7 @@ class Stage extends DisplayObjectContainer #if lime implements IModule #end
 		// when using Haxe 4, this value doesn't seem to get initialized,
 		// so it is undefined, which breaks ObjectMap. this line changes it
 		// from undefined to 0, but won't mess with numeric value > 0.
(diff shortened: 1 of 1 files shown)


joshtynjala:
> This is only going to get compiled if on haxe 4 anyways so why check twice.

Thanks! However, I left that there on purpose for consistency with the rest of the codebase. Even though we have the outer conditional, I felt it was still better to have the inner code look the same here as it does everywhere else throughout OpenFL. It should have virtually no impact on compiler performance.

>  I feel like it would probably be a good idea to make a function that just returns the proper value depending on the haxe version to make things look a lot neater.

I may be remembering incorrectly, but I seem to recall that __js__ isn't actually a function in Haxe 3, so this might be be possible.

Regardless, it's likely that we'll drop Haxe 3 after Haxe 5 is released, so we'll completely remove the `__js__` parts at that point. With that in mind, I'd rather not spend time coming up with something cleaner.

PR #2769, opened 2025-03-10
Merge Develop into 9.5.0

dimensionscape:


joshtynjala:
Heads up about a minor issue.

This method of merging seems to create new hashes for each commit added to the target branch. For instance, 17e961ed6e775b51ba411497755947c2d7e1e65e in develop became 0f8d7e2debb75e105e9f637ebd87ebaff1654699 in 9.5.0-dev.

However, when I ran a simple `git merge develop` on 9.5.0-dev recently, I can see that 9f0b2d66712b58c5cd28523c3e05ce8026ea9642 has the same commit hash on both branches.

It's not a huge deal, but I recall that creating the CHANGELOG for one of our recent feature releases (probably Lime 8.2.0) was a bit more time-consuming than usual because there had been one or more PRs similar to this. I discovered quite a few commits that git reported as supposedly new in that release that weren't in the previous tag, but they actually had been included in previous releases with the original pre-merge hash.

dimensionscape:
Ok. I didn't consider that! Thanks for the heads up. Sorry to create extra work, I'll be sure to use the git command going forward instead. If it's any consolation, I'll do the busy work for the next release to make up for it.

PR #2770, opened 2025-03-10
Fix documentation for Rectangle.setTo()

ACrazyTown:
This was partially undocumented


joshtynjala:
Interestingly, it was partially documented the same way in Flash's docs too.

PR #2771, opened 2025-03-17
Handle Shader Errors Properly.

MAJigsaw77:
- Previously, uncaught shader errors couldn’t be caught because they weren’t handled where they occurred. Now, these errors are caught directly at the point where they happen.
- Previously, if the shader field couldn’t be set to `this` Shader instance, it could crash the program. Now, this issue is resolved, ensuring that runtime generated shaders would work properly without causing crashes.


MRK2009:
Wtf

PR #2778, opened 2025-04-19
CairoShape: check if graphics.__cairo is null before trying to draw

### LINKED ISSUE #2777: Crash when trying to draw cleared graphics with lineStyle
**Describe the bug**
A crash occurs when trying to draw cleared graphics with a line style to a bitmap

**To Reproduce**
```haxe
class Main extends Sprite
{
    var shape:Shape;
    var bmd:BitmapData;

    public function new()
    {
        super();
        shape = new Shape();
        addChild(shape);

        bmd = new BitmapData(100, 100);
        addChild(new Bitmap(bmd));

        shape.graphics.lineStyle(3, 0xFF00FF);
        shape.graphics.moveTo(10, 10);
        shape.graphics.lineTo(90, 90);
        bmd.draw(shape);

        stage.addEventListener(MouseEvent.MOUSE_DOWN, (_) ->
        {
            shape.graphics.clear();
            shape.graphics.lineStyle(3, 0xFF00FF);
            bmd.draw(shape);
        });
    }
}
```

Clicking on the window will cause a null access exception here:
https://github.com/openfl/openfl/blob/782627b89de76ff6ca85c267f8e36e2e09e1dc92/src/openfl/display/_internal/CairoShape.hx#L170

**Expected behavior**
Nothing should be drawn instead of causing a crash

**OpenFL Targets**
I have only tested this with targetting Hashlink (1.15.0)

ACrazyTown:
Fixes #2777 

CairoShape never checked if `graphics.__cairo` was null which would result in a crash when trying to access `graphics.__cairo.target`

### DIFF
--- a/src/openfl/display/_internal/CairoShape.hx
+++ b/src/openfl/display/_internal/CairoShape.hx
@@ -42,7 +42,7 @@ class CairoShape
 			var height = graphics.__height;
 			var cairo = renderer.cairo;
 
-			if (cairo != null && graphics.__visible && width >= 1 && height >= 1)
+			if (cairo != null && graphics.__cairo != null && graphics.__visible && width >= 1 && height >= 1)
 			{
 				var transform = graphics.__worldTransform;
 				var scale9Grid = shape.__worldScale9Grid;


joshtynjala:
I see that `CanvasShape` has a similar null check for `graphics.__canvas`. With that in mind, this looks like a good (partial) fix to me. I say partial because I also noticed that `graphics.__width` and `graphics.__height` are > 0, which they should not be if the graphics were cleared. I can see how to fix that, so I'll take care of it too.

PR #2779, opened 2025-04-24
Keyboard: Fix `__getCharCode` to respect `Caps Lock` state for letter keys.

MAJigsaw77:
This PR updates the internal `__getCharCode` method to correctly interpret `Caps Lock` when determining the char code for letter keys `(A–Z)`.


dimensionscape:
This seems like the correct behavior to me,

PR #2786, opened 2025-07-08
Fixing Null access for SimpleButton

m0rkeulv:
Fixing issue where a SimpleButton would cause a Null access if clicked when its parent had cacheAsBitmap enabled and the buttons down state was null.


joshtynjala:
Thanks!

PR #2790, opened 2025-07-30
add Uniform Arrays to Shaders

Blossomical:
basically adds support to uniform float, int, vec, ivec, mat and bool arrays
and an extra variable so the user can specify what shader version they would like to use


Geokureli:
Note: @EliteMasterEric tried to add the version directive, here, as well: https://github.com/openfl/openfl/pull/2722

JonnycatMeow:
I whould say get the compatibility working and core to work too

Raltyro:
This is really needed

PR #2791, opened 2025-08-01
replace deprecated `apple-mobile-web-app-capable` with `mobile-web-app-capable`

ninjamuffin99:
chrome tosses a deprecation warning about this
image

and from some reading from [flutter](https://github.com/flutter/flutter/issues/154596) and [next.js](https://github.com/vercel/next.js/issues/70272) it seems like `mobile-web-app-capable` works just fine across browsers. 

So this shouldn't have drastic repercussions just merely removing a deprecation warning in the template!


joshtynjala:
Already fixed by #2782, but thanks!

PR #2792, opened 2025-08-06
Add DatagramSocket broadcast capabilities for supported targets (cpp and neko).

B-Interactive:
Currently, OpenFL's `DatagramSocket` does not support sending UDP broadcast packets. However, the underlying Haxe `sys.net.UdpSocket`, in my testing, is able to support broadcast for limited number of targets, such as cpp and neko.  HashLink currently does not support it.

This patch adds OpenFL DatagramSocket broadcast capabilities for cpp and neko targets.  Exposing this functionality will enable devs to send broadcast packets (e.g., for LAN discovery, multiplayer games, etc.) with a standard OpenFL API.  I'm needing it for Art-Net node discovery, and Art-Net DMX broadcast purposes.

## Usage ##

In this patch, broadcast is not enabled by default.  A user can enable broadcasts for supported targets by:
```haxe
socket.enableBroadcast = true;
```

A broadcast is made using the existing API, but specifying a destination address of `"255.255.255.255"`. eg:
```haxe
socket.send(bytes, 0, bytes.length, "255.255.255.255", port);
```

## Error handling and tests ##

A useful error will be thrown for non-supported targets (not cpp/neko).
> UDP broadcast is not supported on this platform. Supported platforms: cpp, neko

A useful error is also thrown if one attempts broadcast without enabling it:
> Cannot send to broadcast address. Set enableBroadcast to true first.

Unit tests have been provided and tested.
`tests/socket/src/DatagramSocketTest.hx`

## My dev system ##
Arch based Linux
Haxe 4.3.7
OpenFL git develop current
Lime git develop current

## Targets tested ##
- **cpp**: Supported.
- **neko**: Supported.
- **HashLink**: Not supported
- **others**: Untested (but presumed unsupported)

### DIFF
--- a/src/openfl/net/DatagramSocket.hx
+++ b/src/openfl/net/DatagramSocket.hx
@@ -378,17 +378,17 @@ class DatagramSocket extends EventDispatcher
 			
 			if(isEphemeral){
 				socket = new DatagramSocket();
-				socket.bind(new Host(localAddress), 0);
+				socket.bind(address);
 			}
 			
-			socket.setBroadcast(true);			
+			socket.__udpSocket.setBroadcast(true);			
 
 			//TODO: allow reuse if not ephemeral
 			var broadcastAddress = new Address();
 			broadcastAddress.host = new Host(address).ip;
 			broadcastAddress.port = port;
 
-			socket.sendTo(cast bytes, offset, actualLength, broadcastAddress);
+			socket.__udpSocket.sendTo(cast bytes, offset, actualLength, broadcastAddress);
 			
 			if(isEphemeral) {
 				socket.close();
--- a/src/openfl/net/DatagramSocket.hx
+++ b/src/openfl/net/DatagramSocket.hx
@@ -347,15 +347,15 @@ class DatagramSocket extends EventDispatcher
 		@throws RangeError The port is not between 1 and 65535, inclusive,
 			or the offset or length are outside the bounds of the bytes array.
 	**/
-	public static function broadcast(bytes:ByteArray, offset:UInt = 0, length:UInt = 0, port:Int = 0, localAddress:String):Void
+	public static inline function broadcast(socket:DatagramSocket, bytes:ByteArray, offset:UInt = 0, length:UInt = 0, address:String = "255.255.255.255", port:Int = 0):Void
 	{
-		// Validate parameters
+        #if (cpp || neko)
 		if (bytes == null)
 		{
 			throw new ArgumentError("Parameter bytes must be non-null");
 		}
 
-		if (port <= 0 || port > 65535)
+		if (port < 0 || port > 65535)
 		{
 			throw new RangeError("Port must be between 1 and 65535, inclusive");
 		}
@@ -372,18 +372,15 @@ class DatagramSocket extends EventDispatcher
 
 		var actualLength = (length == 0) ? bytes.length - offset : length;
 
-		#if (cpp || neko)
 		try
 		{
-			var socket = new UdpSocket();
 			socket.setBroadcast(true);
 
-			// Bind to the specific local address on a random available port (0)
 			socket.bind(new Host(localAddress), 0);
 
 			// Set up the destination broadcast address
 			var broadcastAddress = new Address();
-			broadcastAddress.host = new Host("255.255.255.255").ip;
+			broadcastAddress.host = new Host(address).ip;
 			broadcastAddress.port = port;
 
 			socket.sendTo(cast bytes, offset, actualLength, broadcastAddress);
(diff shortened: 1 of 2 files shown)


B-Interactive:
I know you've worked closely with the DatagramSocket class @dimensionscape, so I anticipate this one will be for you.

dimensionscape:
> I know you've worked closely with the DatagramSocket class @dimensionscape, so I anticipate this one will be for you.

@B-Interactive Thanks! My only concern here is that it imposes a hidden cost on every send, regardless of whether or not you're trying to broadcast, and this may be adverse since network sends are often implemented in tight loops where this kind of check is going to increase cpu bound usage.

With that said, the api here in OpenFL is not quite complete as you can see below, that the flash api implements it with a static:
https://airsdk.dev/reference/actionscript/3.0/flash/net/DatagramSocket.html#broadcast()

B-Interactive:
Thank you @dimensionscape, these are great pointers.  I hadn't realised Harman had added the broadcast method in 51.0.0.  That'll be a valuable reference implementation.

dimensionscape:
I'm not opposed to doing things in our own way, or in a *haxe* flavor. We could implement this in such a way that it works as a static function, but also takes advantage of Haxe's static extensions. Just a thought. I agree that we should implement this feature though so I am looking forward to any proposals you might have.

B-Interactive:
I think it's sensible to separate the function from the existing `send()` method, in order to keep `send()` as lightweight as possible, so I'm grateful for that tip.

This has been revised now to more closely mimic the AIR API, with the dedicated static `broadcast()` method.  This makes sending a broadcast a fair bit simpler, and adds no baggage to the existing `send()` method.

There is one main caveat however.

In my testing, I was not able to send a successful broadcast, unless the socket was bound to a specified IP.  The default IP of `"0.0.0.0"` fails to assign an interface.  I've played around with some ways to handle this (and you may have some valuable input there too), but the most robust and performant solution in the end _I think_ was to simply require an IP address be passed to the `broadcast()` method.

Solutions that attempted to automatically acquire a local IP were problematic and introduced too much overhead for a static function.  I think that's best left up to each individual implementation.

~David (Bink)

dimensionscape:
> The default IP of "0.0.0.0" fails to assign an interface

have you tried with the latest dev? I recently addressed this here: https://github.com/openfl/openfl/commit/8bf27cbe2de9eeb255b41ff449ab72f49438963e

B-Interactive:
>have you tried with the latest dev? I recently addressed this here: https://github.com/openfl/openfl/commit/8bf27cbe2de9eeb255b41ff449ab72f49438963e

Ah I see.  [`DatagramSocket.representativeHost()`](https://github.com/openfl/openfl/blob/develop/src/openfl/net/DatagramSocket.hx#L362) handles that for DatagramSocket.  I'd been testing directly with UdpSocket (broadcast being a static method), but the solution it provides was similar to one I'd explored (`sock.connect(new sys.net.Host("8.8.8.8"), 53)`).  These were my reservations with that approach:
* Attempting this with no internet, causes an application hang.
* As broadcast is static (if it remains AIR API consistent), it would not access the existing DatagramSocket instances assigned address, instead performing representativeHost()-like test again, which added too much overhead per-broadcast.

The way I've addressed this in my current personal Art-Net project, is to not define broadcast as a static method.  When broadcast is called, it still creates a new UdpSocket, but it efficiently references the instanced DatagramSocket's  `localAddress` to bind to.

Perhaps in a similar way, simply making broadcast a regular instanced method (departing from AIR API), it can simply reference and bind to `this.localAddress`.

With respect to the no-internet - hanging issue, I'd be interested to know if that behaviour is consistent.  I'm testing under Linux, and I can confirm it's currently doing this with the latest dev version of DatagramSocket.  In my attempts to work with the [`sock.connect(new sys.net.Host("8.8.8.8"), 53)`](https://github.com/openfl/openfl/blob/develop/src/openfl/net/DatagramSocket.hx#L368) approach, I hadn't yet found a way to handle that gracefully.

dimensionscape:
> Perhaps in a similar way, simply making broadcast a regular instanced method (departing from AIR API), it can simply reference and bind to this.localAddress.

We can actually permit both by taking advantage of a Haxe's static extension feature. I've moved toward some changes that should permit us to support the AS3 version *and* be used as a regular method by way of static extension. You just need to declare `using DatagramSocket;` like this: 
```hx
...
import openfl.net.DatagramSocket;

using DatagramSocket;

Class SomeClass {
   function new(){
      var socket:DatagramSocket = new Socket();
      ...
      socket.broadcast(...);
   }
}
```

and you should be able to use it as a method as shown, while keeping the AS3 aligned structure without duplication.

> With respect to the no-internet - hanging issue, I'd be interested to know if that behaviour is consistent. I'm testing under Linux, and I can confirm it's currently doing this with the latest dev version of DatagramSocket. In my attempts to work with the [sock.connect(new sys.net.Host("8.8.8.8"), 53)](https://github.com/openfl/openfl/blob/develop/src/openfl/net/DatagramSocket.hx#L368) approach, I hadn't yet found a way to handle that gracefully.

I will look into this.

PR #2793, opened 2025-08-07
Use `gl.bufferSubData()` to update existing buffers.

player-03:
Using `gl.bufferData()` deletes the old buffer and allocates a new one, which isn't necessary if the length stayed the same.

### DIFF
--- a/src/openfl/display3D/IndexBuffer3D.hx
+++ b/src/openfl/display3D/IndexBuffer3D.hx
@@ -30,7 +30,7 @@ import openfl.Vector;
 {
 	@:noCompletion private var __context:Context3D;
 	@:noCompletion private var __id:GLBuffer;
-	@:noCompletion private var __memoryUsage:Int;
+	@:noCompletion private var __memoryUsage:Int = -1;
 	@:noCompletion private var __numIndices:Int;
 	@:noCompletion private var __tempUInt16Array:UInt16Array;
 	@:noCompletion private var __usage:Int;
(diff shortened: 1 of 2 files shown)


joshtynjala:
@player-03 Is there any chance that you can share a project where this change results in a performance boost?

In theory, it seems like using `bufferSubData` to reuse the existing buffer should be faster, of course. However, I'm having trouble coming up with a way to test it properly. I've been trying to tweak BunnyMark in various ways, but FPS is basically the same no matter what I try. In fact, on some devices, I seem to notice FPS drop a bit, which makes me think that I'm not pushing the GPU hard enough, and it might be the CPU that is getting overworked by the larger number of bunnies.

player-03:
I don't have a proof of concept to share right now, but this would be relevant to `LOCAL_DYNAMIC` particles in Away3D. Whenever you change a dynamic property, Away3D reuploads the whole buffer.

I can let you know when I have a demo, but for now, this is just a best practice I noticed we aren't following.

nanjizal:
I reused the buffer when experimenting on rendering triangles to screen in webgl, I think in made a significant difference. I use it in the ply classes here.
https://github.com/nanjizal/kitGL/blob/master/src/kitGL/glWeb/PlyMix.js.hx
there is actually a lime version, but did not implement as much for lime as webgl.
https://github.com/nanjizal/kitGL/blob/master/src/kitGL/glLime/Ply.hx

nanjizal:
I am sure that reusing the buffer when updating every frame makes a difference in raw lime and webgl experiments with thousands of triangles and is significant in some experiments just running. I am not sure what the overhead of checking the length is. You may find there are other aspects which obscure the speed increase, but improving performance in small tweaks is definitely worthwhile.

player-03:
I finally got around to making [a test case](https://gist.github.com/player-03/096d019694649327b3ae3ec6d3a92576) for this, but can't seem to prove there's a difference between the two. In HashLink on a several-year-old Linux machine, it runs at a solid 61 fps (I set it to 60, but `openfl.display.FPS` reports 61) even when uploading 65536*150 floats each frame.

The framerate drops if I increase the number of uploads per frame, but it drops by the same amount regardless of `bufferData()` vs. `bufferSubData()`.

- Perhaps some GL implementations include this optimization in `bufferData()`, despite what the standard says. If so, I'd argue we should merge it because some people might be running a GL version that doesn't optimize it.
- Maybe it isn't optimized, but the allocation is fast enough that you can't tell. (The slowdown could be from the memory copy step, after the allocation is done.) In this case it's still wasting memory, so it wouldn't hurt to merge.

PR #2796, opened 2025-08-19
Fix Bubbles Default Events for openfl_pool_events

Raltyro:
Fixes events with bubbles true default not dispatching such as MouseEvent.MOUSE_UP for TextField not being dispatched for the last object stack


joshtynjala:
Thank you!

PR #2801, opened 2025-09-23
Fix Haxe 5 Compatibility

CCobaltDev:
Due to the `IMap` interface being updated in Haxe 5 (https://github.com/HaxeFoundation/haxe/pull/11961), a new `size` function is needed.


joshtynjala:
It looks like `size()` wasn't yet added in preview.1. That's unfortunate. I wonder if we can tweak the conditional compilation to support preview.1 too.

PR #2805, opened 2025-10-06
Fallback to regular sound if audio streaming isn't supported

ACrazyTown:
When running on a platform with `lime_vorbis` enabled, `Assets.getMusic()` always attempted to stream the music using VorbisFile. However, trying to play something like a WAV resulted in nothing because `Assets.getMusic()` always assumed Vorbis was used


joshtynjala:
Thanks!

PR #2806, opened 2025-10-12
NetStream: Fix video element event listener type

TrentexTech:
Change event listener type of the video element from "durationchanged" to "durationchange" to actually capture the event,
and change the listener method name accordingly.
see: [MDN#HTMLMediaElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/durationchange_event)


joshtynjala:
Thanks!

Issue #2812, opened 2025-10-24
ColorTransform multipliers not working for vector drawn objects using GL Renderer, SWF imported display objects work correctly somehow.


twilson90:
This makes no sense.

I have a graphics drawn red rectangle, and a red rectangle imported as a SWF asset.
The first one renders incorrectly, it appears yellow but it should appear green but the colortransform multipliers do nothing.
The second one renders correctly, appears green.

```
var a = new Sprite();
a.graphics.beginFill(0xFF0000);
a.graphics.drawRect(0,0, 100, 100);
a.graphics.endFill();
a.transform.colorTransform = new ColorTransform(0,0,0,1,0,255,0,0);
trace(a.transform.colorTransform);
addChild (a);

var mc = Assets.getMovieClip("colortransform:");
var b = mc.getChildAt(0);
trace(b.transform.colorTransform);
addChild(b);
```

[draw-bug.zip](https://github.com/user-attachments/files/23122417/draw-bug.zip)

<img width="605" height="453" alt="Image" src="https://github.com/user-attachments/assets/0f9686ad-3d9b-47f5-8bba-2a2695e1d064" />

```
Source/Main.hx:18: (redMultiplier=0, greenMultiplier=0, blueMultiplier=0, alphaMultiplier=1, redOffset=0, greenOffset=255, blueOffset=0, alphaOffset=0)
Source/Main.hx:23: (redMultiplier=0, greenMultiplier=0, blueMultiplier=0, alphaMultiplier=1, redOffset=0, greenOffset=255, blueOffset=0, alphaOffset=0)
```

If I set hardware=false in project.xml it works.

twilson90:
After some further tests, there is certainly an issue when using only drawRect and no other graphics shape commands.

If I do:

```
a.graphics.lineStyle();
a.graphics.beginFill(0xFF0000);
a.graphics.drawRect(0,0, 100, 100);
a.graphics.endFill();
```

or

```
a.graphics.beginFill(0xFF0000);
a.graphics.drawCircle(50,50,100);
a.graphics.endFill();
```

It works.

PR #2813, opened 2025-10-24
Fix GL color transforms: apply world multipliers correctly

twilson90:
ColorTransform multipliers were being ignored when drawing primitives in the GL Renderer.
I think this fixes it.

https://github.com/openfl/openfl/issues/2812


joshtynjala:
I decided to manually merge without extracting the code into a new `__setBaseColorTransform` method.

I feel that PRs can be reviewed more quickly and with less mental overhead when they are focused on one thing. In this case, I'd rather apply the bug fix, and any refactoring to remove duplicate code should be separate for now, to be possibly revisited later.

PR #2815, opened 2025-10-28
Fix: graphics __strokePadding not resetting on lineStyle clear.

twilson90:
Tiny fix for Graphics API, __strokePadding wasn't being reset on clearing lineStyle.

### DIFF
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -1488,6 +1488,8 @@ import js.html.CanvasRenderingContext2D;
 			joints = JointStyle.ROUND;
 		}
 
+		__strokePadding = 0;
+
 		if (thickness != null)
 		{
 			if (joints == JointStyle.MITER)
@@ -1662,13 +1664,15 @@ import js.html.CanvasRenderingContext2D;
 		__bitmap = null;
 
 		#if (js && html5)
-		if (__canvas != null) {
+		if (__canvas != null)
+		{
 			__canvas.width = 0;
 			__canvas.height = 0;
 			__canvas = null;
 		}
 
-		if (__context != null) {
+		if (__context != null)
+		{
 			__context.clearRect(0, 0, 0, 0);
 			__context = null;
 		}


src/openfl/display/Graphics.hx
```diff
@@ -1488,6 +1488,8 @@ import js.html.CanvasRenderingContext2D;
 			joints = JointStyle.ROUND;
 		}
 
+		__strokePadding = 0;
```
joshtynjala, line 1491: Completely clearing the value of `__strokePadding` every time that `lineStyle()` is called feels like it might not be quite the correct solution.

Consider this: If `__strokePadding` is always reset to 0, then the following comparisons will always return true. So the comparisons wouldn't be necessary any more. But then, why were they there? Was it simply a misunderstanding in how Flash behaves, or is there a more complex edge case that this PR will end up breaking?

```
if (joints == JointStyle.MITER)
{
	if (thickness > __strokePadding) __strokePadding = Math.ceil(thickness);
}
else
{
	if (thickness / 2 > __strokePadding) __strokePadding = Math.ceil(thickness / 2);
}
```

If there is a more complex edge cases, I'm guessing it involves a `Graphics` object that contains a mix of multiple line thickness values.

twilson90: Thanks Josh!
I realised this was a problem when working on the other branch (https://github.com/twilson90/openfl/tree/gl-bitmap-fill-matrix) and after getting a better understanding of the way the draw command buffers worked. I should have closed this PR but I've been busy recently, apologies.

twilson90:
Looking at this again and wow it was really stupid of me not to wrap `if (thickness==null) { __strokePadding = 0; }`
I think that's what I meant to do but my brain wasn't working.
Because there is still an issue here in that clearing lineStyle still leads to incorrect bounds being returned.

And that bit you high-lighted is odd, I understand why the __strokePadding is rounded upwards, but the `JointStyle.MITER` check appears to just double the strokePadding to make up for joints between 45(ish)-90 degrees where the miter joint's outer corner can potentially grow outside the bounds. But it doesnt account for very small angles, where the miter joint corner can be quite extreme in its length.
This really ought to be changed and some calculation done in lineTo to inflate the bounds correctly for miter joints.

PR #2816, opened 2025-11-01
CairoGraphics: Respect renderer's smoothing

ACrazyTown:
Just like `CanvasGraphics`, adds a `setSmoothing()` function to `CairoGraphics` which disables smoothing if the renderer doesn't allow it. This fixes an issue where drawing a shape onto a bitmap resulted in a smoothed drawing regardless of whether the smoothing argument was true or not.

### DIFF
--- a/src/openfl/display/_internal/CairoGraphics.hx
+++ b/src/openfl/display/_internal/CairoGraphics.hx
@@ -732,7 +732,7 @@ class CairoGraphics
 		var setStart = false;
 
 		cairo.fillRule = EVEN_ODD;
-		cairo.antialias = SUBPIXEL;
+		setSmoothing(true);
 
 		var hasPath:Bool = false;
 
@@ -1446,7 +1446,7 @@ class CairoGraphics
 					var t1:Float, t2:Float, t3:Float, t4:Float;
 					var dx:Float, dy:Float;
 
-					cairo.antialias = NONE;
+					setSmoothing(false);
 
 					while (i < l)
 					{
@@ -2253,6 +2253,18 @@ class CairoGraphics
 		}
 		#end
 	}
+
+	#if lime_cairo
+	private static function setSmoothing(smooth:Bool):Void
+	{
+		if (!allowSmoothing)
+		{
+			smooth = false;
+		}
+
+		cairo.antialias = smooth ? SUBPIXEL : NONE;
+	}
+	#end
 }
 
 private typedef NormalizedUVT =


joshtynjala:
I tried a few different things, but I'm not sure how to test this change. Can you add some code that reproduces the difference in smoothing?

ACrazyTown:
```haxe
{
	public function new()
	{
		super();

		var shape = new openfl.display.Shape();
		shape.graphics.beginFill();
		shape.graphics.drawCircle(5, 5, 5);
		shape.graphics.endFill();

		var bmd = new openfl.display.BitmapData(10, 10, true, 0);
		bmd.draw(shape);

		var bitmap = new openfl.display.Bitmap(bmd);
		bitmap.scaleX = 10;
		bitmap.scaleY = 10;
		addChild(bitmap);
	}
}
```

The `smoothing` argument in `BitmapData.draw()` is false by default, so the expectation would be that the resulting circle would not have any smoothing applied to it. However, this is not the case:

image

With this patch, Cairo's antialiasing will be updated according to whether smoothing is allowed.

image

Weirdly enough, I could've sworn this was not an issue when targeting HTML5. But now that I've tested it again, I can see that smoothing is applied to the circle regardless of `smoothing` there too.

joshtynjala:
It looks like there may be a separate issue where `<window allow-high-dpi="true"/>` (which is true by default for html5) results in heavy smoothing either way. Everything was coming up super blurry for me on all targets. To properly see the difference, I discovered that I needed to test using `<window allow-high-dpi="false"/>` instead.

With `<window allow-high-dpi="false"/>`, I can see the difference on native targets with your PR. I don't seem to notice much difference on the html5 target, though. I also don't see much difference in Adobe AIR either.

joshtynjala:
Just adding some code that shows the two side by side, just to make it easier to compare:

```haxe
var shape = new openfl.display.Shape();
shape.graphics.beginFill();
shape.graphics.drawCircle(5, 5, 5);
shape.graphics.endFill();

var bmd = new openfl.display.BitmapData(10, 10, true, 0);
bmd.draw(shape, null, null, null, null, true);

var bitmap = new openfl.display.Bitmap(bmd);
bitmap.scaleX = 10;
bitmap.scaleY = 10;
addChild(bitmap);

var shape2 = new openfl.display.Shape();
shape2.graphics.beginFill();
shape2.graphics.drawCircle(5, 5, 5);
shape2.graphics.endFill();

var bmd2 = new openfl.display.BitmapData(10, 10, true, 0);
bmd2.draw(shape2, null, null, null, null, false);

var bitmap2 = new openfl.display.Bitmap(bmd2);
bitmap2.scaleX = 10;
bitmap2.scaleY = 10;
bitmap2.x = 150;
addChild(bitmap2);
```

ACrazyTown:
I've gathered some more info. The reason I thought this worked on HTML5 was likely due to the fact that in my project I had the `image-rendering: pixelated;` CSS style applied to OpenFL's main canvas element, which removes a lot of the smoothing artifacts and results in a much more crisp image.

I've also tried Flash and while it's true that the `smoothing` argument doesn't do much there either, it's possible to get a pixelated drawing by setting `stage.quality = LOW;` right before calling `bmd2.draw()`. Because OpenFL's low stage quality globally disables smoothing, this patch makes that case work on OpenFL native targets as well.

PR #2817, opened 2025-11-08
Fixes & Improvements to Renderers

twilson90:
This is still a WIP, not extensively tested yet, bound to be some unexpected problems.
This isn't ready for merging yet but I will continue to work on it.
As the branch name suggests, this started as a simple fix for bitmapFills but I kept finding things that were broken and needed fixing...

CanvasGraphics:
    - Added draw triangles bitmapFill matrix support (like Flash if no UV supplied)
    - Fixed draw triangles uv mapping issues
    - Added draw triangles shared edge gap filler (inherant problem in the way Canvas draws subpixel diagonals, small gaps between shared edges. Hacky, needs work)
    - Added draw triangles stroke / lineStyle support (like Flash)
    - Tidied up scale9Grid implementation
    - Tidied up render / playCommands / hitTest
    - Improved drawRoundRect scale9Grid implementation
    - Fixed various hitTesting issues (top-left false stroke detection, scale9Grid, drawTriangles, and more probably)

CairoGraphics:
    - Added draw triangles bitmapFill matrix support
    - Fixed draw triangles uv mapping
    - Tidied up scale9Grid implementation
    - Tidied up render / playCommands / hitTest
    - Improved drawRoundRect scale9Grid implementation

Context3DGraphics:
    - Added DRAW_ELLIPSE, DRAW_CIRCLE, DRAW_ROUND_RECT

Context3DDisplayObject:
    - Fixed opaqueBackground + renderer clearing issue (now resembles Flash behaviour)

Todo:
    - scale9Grid clipping and scaling issue with strokes:
    chrome_kXx7L1Z795
    - scale9Grid if width/height reduce by more than scale9Grid width/height then it must revert to regular scaling x/y


joshtynjala:
Just a heads up. It’s likely that we’ll ask you to split these up into separate PRs to make it easier to test and review each thing properly.

twilson90:
Looking into the scale9Grid scaling issue further, I realised Graphics.__bounds is calculated with each draw command, including the stroke thickness.
This doesn't account for scale9Grid (which might be applied after the draw commands) which might affect the thickness of strokes.
__bounds has to be calculated on demand and recalculated with a dirty flag to accomodate scale9Grid.
Something like:
`if drawCommand changes: bounds_dirty = true`
`if scale9Grid && scale changes: bounds_dirty = true`
And then each time bounds is demanded it's recalculated if dirty.
`getBounds(stroke:Bool) { if dirty return __bounds else { // calculate } }`
This would also allow for getRect to return a different result, as it's not supposed to include strokes. Currently they return the same rectangle.

Also, ignoring scale9Grid, I also noticed I get slightly different rendering results for strokes when scaled when compared to Flash:

image

On the left is Flash, on the right is HTML5 Canvas. (again to emphasize, no scale9Grid)
It appears Flash averages the thickness of all strokes in quite an odd way... I have a thickness of 50, but when I set scaleX to 0.5 the overall line thickness becomes 40. Not sure how it's getting that number.

twilson90:
Updated the branch:

-  Graphics bounds calculation overhaul: Graphics bounds are now calculated on demand when a dirty flag is set (whenever commands are added / cleared, and potentially when scale9Grid is set). The original method led to certain inaccuracies, for instance, the bounds of a stroked thick line with CapsStyle/JointsStyle can change depending on whether the path is closed or open, but due to the way the graphics API will automatically resolve paths on render, the end of a fill / closed paths were impossible to determine in this incremental way. Also with the new implementation we can use it for calculating fill / stroke bounds in scale9Grid, when the old method was too limited (still some work to do here). In the new implementation, JointTypes & CapsStyle are now considered, whereas before there was just an arbitrary padding for miter joints.
- Support for DisplayObject.getRect(): New variable __boundsExStroke, which is a separate rectangle for storing the bounds of an object excluding thick strokes. Also added extra parameter to __getBounds() functions, exStroke:Bool.
- Additional fixes to DisplayObject.opaqueBackground for all targets: Originally was not accounting for local bounds x / y. Not exactly the same as Flash now but cloose enough, Flash appears to draw individual opaqueBackgrounds for each child of a DisplayObjectContainer, unless there's overlap, then it draws a rect encompassing the 2 overlapping rects. I've found a couple strange quirks/bugs with Flash's implementation, particularly some issues with the bounds of the opaqueBackground when strokes are applied (especially mitered strokes) it can be very inaccurate. No such issues with this new implementation.
- Path closures and various drawing API inconsistencies updated to match Flash's: Originally I was seeing minor differences with the way OpenFl handled unclosed paths with fills, manually closed paths (where lineTo/curveTo explicitly closes the path), or the way it handles beginFill in the middle of a series of lineTo/curveTo commands. I have rewritten it for Canvas and Cairo so it now matches Flash's behaviour.
- Fixed Gradients for Canvas / Cairo: After struggling to get Gradients working in the Canvas Renderer I took a look at Ruffle and lifted their solution. There's still an issue with stroke gradients which I can't see a reasonable fix for.
- Tidied up the Scale9Grid implementation some more.
- Removed what appears to be an arbitrary 1.5px pad in OpenGLRenderer.__scissorRect.

TODO:
- Fix scale9Grid objects with strokes
- Thoroughly test Graphics API.
- Thoroughly test scale9Grid implementation.

twilson90:
- DisplayObject.scale9Grid implementation simplified + finished. Graphics made dirty when scale9Grid set, Scale9Grid.hx + Scale9GridBounds.hx deleted and merged into Graphics.hx. Stroke thicknesses taken into account.
- DisplayObject.opaqueBackground further improvements + fixes. DisplayObject.cacheAsBitmap now render opaqueBackground. GL renders rotated rect now. Also fixed issue with software render on Canvas + Cairo was also rendering the opaqueBackground to __cacheBitmap in DisplayObject.hx.
- ColorTransform.__isDefault and ColorTransform.__equals altered to allow for very small differences in multipliers. I noticed unnecessary BitmapData.colorTransform() calls when multipliers = 0.99999, which happens after an identity ColorTransform is inverted and combined in the render pipeline.
- Added basic LineScaleMode support (not identical to Flash due to inherant canvas limitations but sort of half-way between openFl and Flash).
- Added Graphics.beginShaderFill matrix parameter support.
- Added Graphics.__readGraphicsData support for LINE_GRADIENT_STYLE, LINE_BITMAP_STYLE + BEGIN_SHADER_FILL.
- Added scale9Grid support for DRAW_QUADS in CairoGraphics & CanvasGraphics.
- Simplified CairoGraphics & CanvasGraphics hitTest function.

Fixes:
- Fixed Graphics bounds calculation.
- Fixed CairoGraphics & CanvasGraphics hitTest.
- Fixed CanvasTextField not rendering.

Test suite demonstrating all the old issues with openFL that I've addressed:
https://github.com/twilson90/openfl-test-suite

Try building in current version of openFl (or dev branch) to see all the original issues as they were, and build again in this branch to see all the fixes.

twilson90:
I think this branch is at a stage now where it should be looked over and tested by someone else, there might be something glaring that I've missed.
When I started using openFl again towards the end of last year I was a bit dismayed by all the issues I immediately encountered... this goes some way towards fixing them.
In the process I've completely gutted the CanvasGraphics and CairoGraphics files, rewritten most of the scale9grid stuff and made the Graphics bounds calculation a lot more accurate (and fixing getRect to return stroke-excluded bounds).
The improved bounds calculation has also meant the subpixel precision of strokes is better.
Theres a load of other little fixes in there which I've tried to list in the commit messages but I've probably missed a few.
It'll be difficult to unstitch this and make into separate PRs, but I suppose it's not impossible.

Again, to see the various fixes I've also updated https://github.com/twilson90/openfl-test-suite
The issues are obvious when you compile openfl-test-suite with the latest version of openFl and contrast/compare when compiled with my branch.

PR #2819, opened 2025-11-10
Allow music streaming on HTML5

ACrazyTown:
Howler.js allows us to [stream audio by using HTML5 audio](https://github.com/goldfire/howler.js?tab=readme-ov-file#streaming-audio-for-live-audio-or-large-files) instead of using the default Web Audio. With this we can implement `Assets.getMusic()` on HTML5. The returned sound will be cached to avoid a re-download if played again via `Assets.getSound()`.

I'm a bit unsure with how this plays with the asynchronous nature of HTML5 and OpenFL's libraries. Loading a library seems to load all sounds normally, so this may only work if the music is stored in an unloaded library.

### DIFF
--- a/src/openfl/utils/Assets.hx
+++ b/src/openfl/utils/Assets.hx
@@ -1,5 +1,6 @@
 package openfl.utils;
 
+import lime.media.AudioBuffer;
 import openfl.utils._internal.Log;
 import openfl.display.BitmapData;
 import openfl.display.MovieClip;
@@ -16,6 +17,9 @@ import lime.utils.Assets as LimeAssets;
 #if lime_vorbis
 import lime.media.AudioBuffer;
 import lime.media.vorbis.VorbisFile;
+#end
+#if lime_howlerjs
+import lime.media.howlerjs.Howl;
 #end
 
 /**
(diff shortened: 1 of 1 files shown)


dimensionscape:
The entire sound file is loaded into memory here and then cached. This undermines the purpose of streaming, which is used to effectively save system RAM for large music files. When we stream, on native, we typically use 2 or 3 tiny buffers and only those buffers that hold small audio segments take up any memory. In the browser, it is much higher level and we don't actually handle chunks of pcm manually. What we need to do is create a unified streaming mechanism that works like VorbisFile in Lime. For html5, If we want to cache the audio to avoid downloading it on subsequent replays, we need to cache it on disk using available browser api. I'm trying to consider the best way to unify audio streaming with native streaming properly. In essence, I imagine we'd have something like `AudioStreamSource` that unifies the layer for close, seek and read(on native). I need to think about this more.

ACrazyTown:
Ah right, I guess it's not proper streaming like on native. However, with this audio playback will start immediately and the rest of the sound will download as it plays, compared to the default behavior where the entire sound needs to be downloaded before it can be played

dimensionscape:
Yea, we would want to play remote audio uri's immediately instead of downloading the entire thing and enable this behavior on both native and html5, with the cache potentially going to disk instead of memory. If you stream a 4 hour podcast, just for example, that's a huge chunk of memory to cache.

In a game, you might have dozens of background music available and we dont want to cache them all in memory when we only play one at a time, and we also dont want to keep downloading each audio file every time, so we need to make sure they can be cached to disk and played. I think we can line up this behavior on native and web to behave the same way. The new SDL_Sound decoding backend that I recently pushed will let us replace the vorbis based streaming and achieve a better result here. I'm working on a potential unified api for Lime, and hopefully we can backfill the music asset internals with it.

Issue #2775, opened 2025-04-08
SharedObject getLocal() inconsistent storage ID naming


IriySoft:
`openfl.net.SharedObject.hx.getLocal(nameStr)` function uses illegal character (":")  to join **path** and **name** if **path** is not passed to the function for html5 and js targets. In this case the function sets **path** to Browser.window.location.pathname(). But if **path** is passed, the function uses "\/" to join it with **name**, that is inconsistent.

Example:

if I call `getLocal("storage_name");` in a JS located at site.com/index.html, the local storage ID will become

"/index.html:storage_name"

If I then try to call `getLocal("storage_name", "/index.html");` the function will try to locate a local storage with this ID and fail (or rather create a new local storage with a new name):

"/index.html/storage_name"

If the same getLocal(nameStr) is then used to load the previously stored data, it's ok, since the same ID with ":" is used again, but it's impossible to access an existing local storage with such ID if **path** is passed to the function, because in this case the function joins path and name using "\/" character and there is no way to do it other way (like to pass "" as **path** and local storage **name** without "\/"  `getLocal("index.html:storage_name", "");`) because ":" is in the illegal characters list that can't be used in **name** and the function returns null.

The code from getLocal:

```
var id = localPath + "/" + name;   // <===== if localPath is set it will stay like this

if (!__sharedObjects.exists(id))
{
	var encodedData:String = null;

	try
	{

	#if (js && html5)
	var storage = Browser.getLocalStorage();

	if (localPath == null)
	{
		// Check old default path, first
		if (storage != null)
		{
			encodedData = storage.getItem(Browser.window.location.href + ":" + name);
			storage.removeItem(Browser.window.location.href + ":" + name);
		}

		localPath = Browser.window.location.pathname;    // <===== if localPath is not set it is set to href
	}

	if (storage != null && encodedData == null)
	{
		encodedData = storage.getItem(localPath + ":" + name);    // <===== Illegal character used here to join path and name
	}
	. . .

```

**To Reproduce**
1. Save something to a local storage using `openfl.net.SharedObject.hx.getLocal("storage")` in an html5 project
2. Locate the local save storage ID (something like "/index.html:storage") in the browser.
3. getLocal("storage", "/index.html") creates a new local storage with "/index.html/storage" ID.

**Expected behavior**
getLocal("storage", "/index.html") should open an existing local storage instead of creating a new one with a different ID.

[...]

PR #2823, opened 2025-11-14
[HTML5] Added force delimiter optional value to SharedObject.getLocal

IriySoft:
Trying to address the issue in `SharedObject.getLocal(name, localPath)` that makes it impossible to `getItem()` previously created when no `localPath` was passed to `getLocal` and then the URL of the app changed because the function uses an illegal character ":" as a delimiter in this case to create the item.

Details here:

https://github.com/openfl/openfl/issues/2775

With this change it becomes possible to use a force delimiter that bypasses illegal character check. At the same time, if no force delimiter is used, the function will work exactly as before making all the existing code work as expected.


IriySoft:
With the latest commit the force delimiter is now saved in a variable to be used in `clear` and `flush` functions.

PR #2825, opened 2025-11-23
Add ETC2 support to TextureBase.hx

B-Interactive:
Using this update to TextureBase.hx, ETC2 is now working in the Android browser in my tests.

The code checks for ETC2 enums (COMPRESSED_RGB8_ETC2, COMPRESSED_RGBA8_ETC2_EAC) directly on the ETC2 extension object if available, or falls back to checking the GL context itself.
This allows both WebGL 1 (via the extension) and WebGL 2 (where I believe ETC2 formats can exist directly on the GL context).

For your testing, I've created a simple web-app that uses an ATF containing _only_ ETC2 content, and plays a running cat when it works.  The asset was created using TexturePacker 6.0.1, not the official `png2atf` tool.

[https://b-interactive.github.io/openfl-etc2-test/](https://b-interactive.github.io/openfl-etc2-test/)


joshtynjala:
Thanks!

PR #2829, opened 2025-12-29
Misc/loaders modular fix

ConfidantCommunications:
Edge case here…
Using Haxe-modular library I discovered that code from the __init__ function in the Loader.hx class was being duplicated into modules created by the haxe-modular library (when no OpenFL classes were necessary), and then causing errors. 

```
var loaders = [new openfl_display__$internal_BitmapDataLoader(),new openfl_display__$internal_ScriptLoader(),new openfl_display__$internal_AssetManifestLoader()];
if(openfl_display_Loader.__registeredLoaders == null) {
	openfl_display_Loader.__registeredLoaders = loaders;
} else {
	openfl_display_Loader.__registeredLoaders = loaders.concat(openfl_display_Loader.__registeredLoaders);
}
```

Per their [documentation](https://github.com/elsassph/haxe-modular/blob/master/doc/start.md) it was a known behaviour of haxe-modular, i.e. "assume that __init__ code could be duplicated in all the bundles, unless you only use calls to static methods."
I [raised the issue](https://github.com/elsassph/haxe-modular/issues/118#issuecomment-3694581546) and Philippe recommended this as a solution, which seems to work.

### DIFF
--- a/lib/flash-externs/src/flash/ui/KeyLocation.hx
+++ b/lib/flash-externs/src/flash/ui/KeyLocation.hx
@@ -3,10 +3,8 @@ package flash.ui;
 #if flash
 #if (haxe_ver >= 4.0) enum #else @:enum #end abstract KeyLocation(Int) from Int to Int from UInt to UInt
 {
-	// #if (flash && !doc_gen)
 	@:noCompletion @:dox(hide) public static inline var D_PAD = 4;
 
-	// #end
 	public var LEFT = 1;
 	public var NUM_PAD = 3;
 	public var RIGHT = 2;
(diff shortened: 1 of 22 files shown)


joshtynjala:
Your PR targets 9.6.0-dev, but you merged in some commits from develop that weren't yet merged into 9.6.0-dev. I tried merging them in and retargeting your PR to the new 9.6.0-dev, and that removed the extra commits, but it still seems to include a merge commit with unrelated changes, for some reason. It looks like I'll need to manually merge this when I get a chance.

ConfidantCommunications:
Thanks and sorry for the complication. I don't do this very often.

joshtynjala:
Merged as fe419aef443e54b879d62215b9ec6975e904d742

PR #2831, opened 2026-01-12
OpenFL9.5.0 This condition can cause issues with custom shader rendering exceptions

rainyt:
If the custom shader does not use the default bitmap shader bitmap parameters, it will cause the shader to fail to render properly.

I noticed that the old version does not have this condition, and there is also a null judgment for the bitmap in the condition, so there should not be any blocking conditions.

### DIFF
--- a/src/openfl/display/_internal/Context3DGraphics.hx
+++ b/src/openfl/display/_internal/Context3DGraphics.hx
@@ -715,7 +715,7 @@ class Context3DGraphics
 						uvDataLength = verticesLength;
 					}
 
-					if (bitmap != null || (uvDataLength == 0 && fill != null))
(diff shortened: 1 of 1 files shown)


Raltyro:
I was wondering why my shaders wasnt rendering

rainyt:
> I was wondering why my shaders wasnt rendering

Are you using the latest version of OpenFL? If so, it is necessary to check if custom texture data has been used. Can you try if my PR is available.

joshtynjala:
@rainyt Does it work if you keep the condition, but add a check for `shaderBuffer`?

```haxe
if (bitmap != null || shaderBuffer != null || (uvDataLength == 0 && fill != null))
```

rainyt:
@joshtynjala Yeah, It can work normally, I have restored it and added new conditions.

PR #2836, opened 2026-01-27
Rectangle should allow negative width/height (flighthq/flight@388519d)

jgranick:
Rectangles can have a negative width and height. Previously, these were treated as if they were empty. I believe these are originally supported as being mirrored, so they extend to the left or above the origin point, instead of to the right and below.

Included are updates to `contains`, `containsRect`, `isEmpty`, and `union` according to my other tests.

I'm submitting as a pull request here for consideration and feedback.

Have a great day!

### DIFF
--- a/src/openfl/geom/Rectangle.hx
+++ b/src/openfl/geom/Rectangle.hx
@@ -251,12 +251,7 @@ class Rectangle
 	**/
 	public function contains(x:Float, y:Float):Bool
 	{
-		var x0 = Math.min(this.x, this.right);
-		var x1 = Math.max(this.x, this.right);
-		var y0 = Math.min(this.y, this.bottom);
-		var y1 = Math.max(this.y, this.bottom);
-
-		return x >= x0 && x < x1 && y >= y0 && y < y1;
+		return x >= this.x && y >= this.y && x < right && y < bottom;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


joshtynjala:
It seems that the modified assertions are failing on Flash/AIR, so maybe the old behavior was correct?

tobil4sk:
The air unit tests are failing, so maybe the current behaviour is consistent with how it worked in flash?

jgranick:
I'm open to feedback. It's also possible that the tests may be running into edge cases, like whether a zero-size rectangle can be considered contained, or whether the edges are exclusive and inclusive, and whether negative is treated differently. I don't have a good way to run or test Flash Player.

tobil4sk:
From the [ci logs](https://github.com/openfl/openfl/actions/runs/21415357818/job/61662413267?pr=2836#step:9:1634), it seems that all of the changed lines are failing now:
```
results: SOME TESTS FAILURES (success: false)
RectangleTest
  test_isEmpty: FAILURE .FF..
    line: 376, expected false
    line: 377, expected false
  test_containsRect: FAILURE .FF......FF...
    line: 258, expected true
    line: 259, expected true
    line: 269, expected true
    line: 270, expected true
```

So changing those tests seems to deviate from the behaviour of the flash api.

jgranick:
I ported the tests I wrote, and ran them using the Harman AIR SDK.

I've provided another commit which updates the behavior to precisely match that of the AIR SDK. All of the methods were performing similarly except two:

 - `containsRect`
 - `toString`
 
The behavior of AIR is to discard a Rectangle which has negative width or height as an invalid object. It is used to represent bounds, so negative bounds is invalid and ignored. Conversely, OpenFL tries to reason with negative rectangles, to try and make some sense of them in `containsRect`. 

I seem to remember some large project triggering negative rectangle sizes. Perhaps this was Papers Please or something else we know. Either this was due to an existing bug elsewhere, or somehow it had a different behavior in an older version of Flash Player?

`toString` in Flash uses `w` and `h` instead of `width` and `height` in the output string. 

I'd either merge this in as-is to change the behavior, or I'd keep the added tests, and document more clearly the difference in `containsRect`. `toString` could probably live without a documenting the deviation.

joshtynjala:
> I don't have a good way to run or test Flash Player.
> ran them using the Harman AIR SDK

If you're interested, Adobe still provides official URLs to download their final versions of Flash Player (not the browser plugin, but the standalone executable) and the AIR SDK. There's no HTML page that lists them anymore, but the download URLs still work for those specific versions (older version downloads were removed). I made a couple of Gists with the details.

Adobe Flash Player 32.0 Standalone Projector Content Debugger for Windows, macOS, and Linux:
https://gist.github.com/joshtynjala/633fbaca5733a889da31498bdbf45dd6

Adobe AIR SDK 32.0 for Windows and macOS:
https://gist.github.com/joshtynjala/76e80fa8422cdc147733f485b97ae60e

PR #2837, opened 2026-01-27
Fix fast scaleX/scaleY transform update path

jgranick:
The original code checks for `__transform.b == 0` or `__transform.c == 0` as a sign that the local transform has not been rotated.

However it is my understanding that if `rotation = 180` or `rotation = -180`, these will return true except we must use a _negative_ scale value in order to have the proper local transform. As a result, in this specific case of 180 degree rotation, scale values will be wrong.

This could be written to defer the local transform calculation, or to handle other cardinal values, however at a minimum, this should probably resolve a bug.

### DIFF
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -2321,7 +2321,7 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if (open
 		{
 			__scaleX = value;
 
-			if (__transform.b == 0)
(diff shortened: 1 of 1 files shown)


joshtynjala:
I got a report from a user that this specific commit broke some if their SWF assets. We may need to revert.

joshtynjala:
It appears that we don't need to revert. The change was correct, but it exposed some other bugs that I was able to fix.

PR #2841, opened 2026-02-24
Event: Use rest arguments in formatToString for Haxe 4.2+

pngdrift:
This matches Flash and allows you to pass more than 5 field names.


joshtynjala:
Thanks! Nice little improvement.

PR #2842, opened 2026-03-01
Fix StageScaleMode and StageAlign behaviors

### LINKED ISSUE #2768: Incorrect scaling with SHOW_ALL scaleMode
**Describe the bug**
When the window is resized with `stage.scaleMode = SHOW_ALL`, the contents of the window get stretched without preserving the original aspect ratio. The letterboxing is also incorrect, and can cover parts of the application.

**To Reproduce**
Minimal repro:
```haxe
{
    public function new()
    {
        super();
        stage.scaleMode = openfl.display.StageScaleMode.SHOW_ALL;

        this.graphics.beginFill(0x777777);
        this.graphics.drawRect(100, 100, 15, 100);
        this.graphics.endFill();

        addChild(new openfl.display.FPS());
    }
}
```

**Expected behavior**
The documentation says that `SHOW_ALL` is supposed to preserve the aspect ratio without distortion, and that the entire application should remain visible, which is also how this scale mode works in Flash.

**Screenshots**

Image

Image

Compare to resizing in the Flash target:

Image

**OpenFL Targets**
HashLink, Neko, Linux

Tutez64:
Hi, on latest OpenFL version (9.5.1), StageScaleMode.SHOW_ALL/EXACT_FIT/NO_BORDER behaves mostly the same as NO_SCALE, except things get blurry when the windows is resized down, and it adds a mouse offset. I tried oldest OpenFL versions and it gave different but still not right results.
Now, SHOW_ALL/EXACT_FIT/NO_BORDER behaves exactly like on Flash.
As for StageAlign, the alignment was previously applied two times for some reason, which was causing it to be incorrect. I also added the "" value which centers horizontally and vertically, like on Flash.
Note: StageAlign still doesn't seem to do anything when used with StageScaleMode.NO_SCALE, which I think isn't correct.

Validated on Windows/C++ target with a minimal repro and game project.
I believe this should fix https://github.com/openfl/openfl/issues/2768

### DIFF
--- a/src/openfl/display/OpenGLRenderer.hx
+++ b/src/openfl/display/OpenGLRenderer.hx
@@ -390,7 +390,17 @@ class OpenGLRenderer extends DisplayObjectRenderer
 	**/
 	public function setViewport():Void
 	{
-		__gl.viewport(__offsetX, __offsetY, __displayWidth, __displayHeight);
+		if (__defaultRenderTarget == null && __context3D != null && __context3D.__stage3D == null && __stage != null && __stage.window != null)
+		{
+			var contextHeight = Std.int(__stage.window.height * __stage.window.scale);
+			var viewportY = contextHeight - __displayHeight - __offsetY;
+			if (viewportY < 0) viewportY = 0;
+			__gl.viewport(__offsetX, viewportY, __displayWidth, __displayHeight);
+		}
+		else
+		{
+			__gl.viewport(__offsetX, __offsetY, __displayWidth, __displayHeight);
+		}
 	}
 
 	/**
@@ -402,6 +412,10 @@ class OpenGLRenderer extends DisplayObjectRenderer
 	{
 		if (__currentShader != null)
 		{
+			if (__defaultRenderTarget == null && __context3D.__state.renderToTexture == null)
+			{
+				setViewport();
+			}
 			if (__currentShader.__position != null) __currentShader.__position.__useArray = true;
 			if (__currentShader.__textureCoord != null) __currentShader.__textureCoord.__useArray = true;
 			__context3D.setProgram(__currentShader.program);
@@ -777,6 +791,7 @@ class OpenGLRenderer extends DisplayObjectRenderer
 			}
 			#end
 			__context3D.setScissorRectangle(__scissorRectangle);
+			setViewport();
 
 			__upscaled = (__worldTransform.a != 1 || __worldTransform.d != 1);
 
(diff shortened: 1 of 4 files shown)


joshtynjala:
I merged in the `StageAlign` string parsing code (and then fell down a bit of a rabbit hole when I discovered that it was actually more complicated than that). I also merged in the undocumented center alignment for empty/invalid strings.

I'd like to point out that, when adding the `-Dcairo` flag to the `lime build ` command for the cpp, neko, or hl targets, stage align should already be working mostly correctly. At the time I last worked on that, I tried to figure out how to get the default OpenGL renderer to behave the same as the Cairo renderer, but as you discovered, there seem to be some bigger changes necessary for OpenGL to work properly.

I think that you're on the right track, but I'm not sure that your solution is exactly what we should merge. I haven't tested everything yet, but I suspect that some of your changes might conflict with `-Dcairo`. We definitely also want to support Neko and HashLink, and not just C++.

I don't have more concrete suggestions yet. I just wanted to get some initial thoughts down.

Tutez64:
Adding this flag appears to fix scaling/alignment issues, but it makes my game project freeze and I believe it would be less performant even if it worked.
I checked neko and hl, they had similar issues as C++. Using my fork (scaleMode=showAll, align=null, with and without the cairo flag), everything seems to be working as expected, but further testing wouldn't hurt.

joshtynjala:
I have committed a slightly different implementation here: 2c8cfed1f3f19eb253660ac8fbf159a4d825b19c

@Tutez64 Can you give it a try, and let me know if it is working well for you?

In particular, I changed it to pass the unscaled window dimensions to `context3D.configureBackBuffer()`. This more closely matches the `stage.stageWidth` and `stage.stageHeight` values used previously, which are unscaled as well. This seems to have allowed the `OpenGLRenderer.__resize()` method to be simplified, and some of your other modified code that accounted for `stage.window.scale` doesn't seem necessary anymore. It's possible that I missed some edge cases that you were testing, though, so I look forward to your feedback so that I can address that, if needed.

I tested with both `<window allow-high-dpi="true"/>` and `<window allow-high-dpi="false"/>` on a HiDPI monitor on macOS. I also tested on Linux and Windows, but they won't properly support HiDPI until we upgrade Lime to use SDL3.

Tutez64:
Yes, it's working well!

joshtynjala:
Great! Thanks for your PR! It was essential to figuring this out, and I added you as a co-author to my commits.

PR #2843, opened 2026-03-03
Fix references to Lime classes when `lime` isn't defined.

player-03:
Several places import a Lime class inside an `#if lime` check, then reference that class without such a check. Or they extend a Lime class if `lime` is defined, then reference an inherited property without checking. This PR fixes most of those, other than in _ApplicationMain.hx_.

But I think the real question should be, do we check `#if lime` at all? Clearly, large portions of OpenFL only work with Lime, and this hasn't been caught in years, so I think it's safe to assume Lime is always available. Unless I'm missing something?


MRK2009:
Yes please bro

PR #2846, opened 2026-03-10
make logging functions dynamic

Vortex2Oblivion:
makes functions dynamic so you can override them with your own behavior


Vortex2Oblivion:
openfl/lime#2040 will also need to be merged

PR #2849, opened 2026-03-17
Invalidate hovered masked cache bitmap textures in the OpenGL cache path

Tutez64:
## Summary

This fixes a visual issue in the OpenGL `cacheAsBitmap` path for masked content.

When a cached display object with masked descendants is re-rendered while hovered, the main cache texture may be reused in an inconsistent state. In practice, this can cause masked content to appear outside its mask.

## Changes

- detect when a cached display object is related to the current hovered display path
- invalidate the main cache texture only in that masked OpenGL cache case
- keep the existing cache buffer reuse behavior unchanged

## Testing

Tested in a native C++ OpenFL project with SWF-heavy UI.

Observed behavior:
- the masked hover glitch no longer reproduces
- no obvious visual regressions were observed in additional UI checks

### DIFF
--- a/src/openfl/display/BitmapData.hx
+++ b/src/openfl/display/BitmapData.hx
@@ -3180,13 +3180,14 @@ class BitmapData implements IBitmapDrawable
 	@:noCompletion private function __drawGL(source:IBitmapDrawable, renderer:OpenGLRenderer):Void
 	{
 		var context = renderer.__context3D;
+		var requireDepthStencil = __requiresRTTDepthStencil(source);
 
 		var cacheRTT = context.__state.renderToTexture;
 		var cacheRTTDepthStencil = context.__state.renderToTextureDepthStencil;
 		var cacheRTTAntiAlias = context.__state.renderToTextureAntiAlias;
 		var cacheRTTSurfaceSelector = context.__state.renderToTextureSurfaceSelector;
 
-		context.setRenderToTexture(getTexture(context), true);
+		context.setRenderToTexture(getTexture(context), requireDepthStencil);
 
 		renderer.__render(source);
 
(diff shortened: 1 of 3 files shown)


joshtynjala:
@Tutez64 In your PRs, can you add a simple project that can reproduce the issue that you are fixing? Or a small code snippet that easy to copy/paste into a blank project?

Before merging, we always want to test that a PR does what it says, and to spend some extra time trying to discover additional edge cases that may not have been considered. If we have a starting point for that, it helps a lot. Thanks!

(EDIT: I saw that some of your PRs mention using SWF assets. You don't necessarily need to share your real-world project assets. Simple geometric "developer art" assets are fine too, if they are enough to reproduce. Or if you can do simple programmatic `Graphics` stuff, that's even better, since managing asset files complicates things.)

Tutez64:
Sure. I wasn't able to easily reproduce the bug without a SWF file, so I stripped down one instead.
[OpenFL cache mask fix.zip](https://github.com/user-attachments/files/26132423/OpenFL.cache.mask.fix.zip)

Tutez64:
This commit fixes VRAM growth issue introduced in the previous commit

Tutez64:
I refined this fix so it no longer disposes the cache texture itself.

For the hovered masked cache case, OpenFL now invalidates only the render-to-texture framebuffer attachments, while keeping the texture alive. This preserves the visual fix without forcing repeated texture recreation.

Short `apitrace` comparisons show a clear reduction in texture churn:

- `glTexImage2D`: `3.905` -> `2.136` per frame
- `glGenTextures`: `1.596` -> `0.707` per frame
- `glDeleteTextures`: `1.006` -> `0.117` per frame

Real texture uploads also dropped from `2687` to `834`.

This change removes most of the extra texture churn introduced by the original fix.

Tutez64:
I refined it again so it now invalidates only the depth/stencil attachments for the affected hovered masked cache case, while keeping both the cache texture and the framebuffer itself.

This preserves the previous reduction in texture churn and also removes most of the framebuffer churn.

Short `apitrace` runs:

- `glGenFramebuffers`: `0.988` -> `0.090` per frame
- `glDeleteFramebuffers`: `0.919` -> `0.019` per frame

Tutez64:
I reduced unnecessary depth/stencil attachment churn for cache redraws.

Previously, `BitmapData.__drawGL()` always rendered to texture with `enableDepthAndStencil = true`, even when the cached content did not use masking or any stencil-dependent path.

This now enables depth/stencil only when the rendered display object tree actually requires it, while preserving the existing hovered masked-cache fix.

`apitrace` comparisons:

- `glRenderbufferStorage`: `0.933` -> `0.623` per frame
- `glDeleteRenderbuffers`: `0.904` -> `0.622` per frame

PR #2851, opened 2026-03-24
Fix stale cacheAsBitmap render when a filtered container loses its last child

Tutez64:
This fixes a stale cacheAsBitmap case for cached display objects.

When a display object has filters, OpenFL enables cacheAsBitmap. If that cached container loses its last child, it may still be marked renderDirty while no longer having children. In that case, __updateCacheBitmap could skip rerendering the cached bitmap, leaving the previous cached image visible.

This patch allows rerendering to proceed when a cached bitmap already exists and the display object is renderDirty, even if it no longer has children.

In practice, this fixes stale/ghost rendering on native targets for filtered containers whose contents are removed or swapped.

I did not prepare a minimal reproduction project because the patch is very small and the issue seemed clear from the code path and observed runtime behavior. If needed, I can provide one.

### DIFF
--- a/src/openfl/display/DisplayObjectRenderer.hx
+++ b/src/openfl/display/DisplayObjectRenderer.hx
@@ -258,7 +258,8 @@ class DisplayObjectRenderer extends EventDispatcher
 			var rect:Rectangle = null;
 
 			var needRender = (displayObject.__cacheBitmap == null
-				|| (displayObject.__renderDirty && (force || (displayObject.__children != null && displayObject.__children.length > 0)))
+				|| (displayObject.__renderDirty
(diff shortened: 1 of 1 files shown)


joshtynjala:
Adding code to reproduce the issue:

```haxe
var container = new Sprite();
container.x = 20;
container.y = 20;
container.filters = [new DropShadowFilter()];
addChild(container);

var child = new Sprite();
child.graphics.beginFill(0xff0000);
child.graphics.drawRect(0, 0, 100, 100);
child.graphics.endFill();
container.addChild(child);

container.addEventListener(MouseEvent.CLICK, event -> {
	container.removeChild(child);
});
```

Without the fix, clicking the container does nothing. With the fix, the child is no longer rendered.

PR #2852, opened 2026-03-29
Avoid bounds-sized BitmapData allocations in CairoGraphics.hitTest

Tutez64:
## Summary

This changes `CairoGraphics.hitTest()` to reuse a tiny shared Cairo surface instead of creating a `BitmapData` sized to the full graphics bounds.

The rendering path is unchanged. This only affects the native Cairo hit-test path used for `cairo.inFill()` / `cairo.inStroke()`.

## Problem

Before this change, `CairoGraphics.hitTest()` allocated a full-size `BitmapData(bounds.width, bounds.height)` for hit-testing.

In native targets, this could cause very large transient allocations when hovering interactive vector `Graphics` with large local bounds, even if the object was visually small on screen because it was scaled down.

Heaptrack showed the hotspot in:

`Stage.__onMouse -> Graphics.__hitTest -> CairoGraphics.hitTest -> BitmapData.__construct`

## Fix

- add a shared 1x1 Cairo surface for hit-testing
- reuse that surface in `CairoGraphics.hitTest()`
- keep the existing path construction and `inFill` / `inStroke` logic unchanged

## Reproduction

This minimal repro [repro-cairo-hit-test.zip](https://github.com/user-attachments/files/26334926/repro-cairo-hit-test.zip) demonstrates the initial native memory spike on first hover.

In a larger real-world project, this transient over-allocation was also causing heavy GC pressure, which led to severe long-term performance degradation and recurring frame drops during gameplay.

### DIFF
--- a/src/openfl/display/_internal/CairoGraphics.hx
+++ b/src/openfl/display/_internal/CairoGraphics.hx
@@ -51,6 +51,8 @@ class CairoGraphics
 	private static var graphics:Graphics;
 	private static var hasFill:Bool;
 	private static var hasStroke:Bool;
+	private static var hitTestBitmap:BitmapData;
+	private static var hitTestCairo:Cairo;
 	private static var hitTesting:Bool;
 	private static var inversePendingMatrix:Matrix;
 	private static var pendingMatrix:Matrix;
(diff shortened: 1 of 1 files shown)


joshtynjala:
It looks like the memory spike on roll over in repro-cairo-hit-test is caused specifically by the `graphics.drawRect()` command in `Main.onAddedToStage()`. The grid of additional sprites in the sample appear to have no affect on the behavior. I can comment out the call to `addTiles()`, and I still observe a spike in memory on roll over.

`drawRect()` with a solid color bypasses the Cairo rendering path and goes through the GL rendering path instead. Drawing a quad with a solid color in GL is easy and fast, so that is a nice little optimization. However, this means that hit testing is happening without going through the Cairo rendering path first. Normally, the Cairo rendering path would have already created a surface before hit testing, so that condition that creates a `BitmapData` would be skipped and no additional surface would need to be created. However, with the GL bypass, there is no surface yet, so it needs to create the `BitmapData`.

I'm going to do a little more testing, but this seems like a good improvement. I'll probably merge it into 9.6.0-dev instead of develop. That branch already has a lot of allocation/garbage-collection improvements that I've been working on, and this fits in perfectly with those as a nice set of improvements for the next release.

joshtynjala:
Merged as 21aaf67336f82850b6225c3f4be445fbce72575a

PR #2854, opened 2026-04-04
Clip oversized OpenGL filter caches to the visible viewport

Tutez64:
## Summary

This adds `__clipFilterCacheBounds` to the OpenGL filter cache path so oversized filtered display objects are cached against the visible viewport instead of their full filter bounds.

This reduces the large initial VRAM cost caused by very large filtered containers/layers whose bounds extend far outside the screen.

## Details

- clip oversized OpenGL filter caches to the visible viewport
- only apply the clipping on the standard render-transform cache path
- track the clipped cache bounds so cached output stays stable when the visible area changes
- clear the tracked bounds when the cache is discarded

The cache-bounds tracking is included to keep the new clipping behavior visually stable during camera/view changes.

## Repro

[OpenFLFilterCacheVRAMRepro.zip](https://github.com/user-attachments/files/26476886/OpenFLFilterCacheVRAMRepro.zip)

It creates several large filtered layers with animated content and tiny off-screen extent markers so the filter bounds become much larger than the visible area.

Before this change, the first filter application was causing a very large VRAM spike because the cache is allocated from the full filtered bounds.
With this change, the initial cache cost is small, bounded by the visible viewport instead.

## Validation

Tested against a game scenario with large grayscale-filtered dungeon layers.

Observed result:
- the initial VRAM spike is greatly reduced
- filtered layers remain visually stable during spectate/camera movement

### DIFF
--- a/src/openfl/display/DisplayObject.hx
+++ b/src/openfl/display/DisplayObject.hx
@@ -968,6 +968,7 @@ class DisplayObject extends EventDispatcher implements IBitmapDrawable #if (open
 	@:noCompletion private var __cacheAsBitmapMatrix:Matrix;
 	@:noCompletion private var __cacheBitmap:Bitmap;
 	@:noCompletion private var __cacheBitmapBackground:Null<Int>;
+	@:noCompletion private var __cacheBitmapBounds:Rectangle;
 	@:noCompletion private var __cacheBitmapColorTransform:ColorTransform;
 	@:noCompletion private var __cacheBitmapData:BitmapData;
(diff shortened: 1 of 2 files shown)


Tutez64:
This extends the visible viewport clipping logic to oversized `cacheAsBitmap` paths that use the current render transform.

Tutez64:
Following https://github.com/openfl/openfl/pull/2856#issuecomment-4270697700:
> You should move this line to #2854 where the `desiredCacheBounds` variable is introduced:
> ```hx
> if (desiredCacheBounds != null) Rectangle.__pool.release(desiredCacheBounds);
> ```

PR #2855, opened 2026-04-04
Optimize software Graphics cache reuse during zoom

Tutez64:
## Summary

This reduces repeated rerasterization/reallocation of software-backed `Graphics` during animated zoom on OpenGL/C++ targets.

The fix separates logical graphics size from allocated raster size, so generic vector `Graphics` can reuse a slightly larger cached surface across nearby zoom levels instead of rebuilding every time the zoom changes a little.

`TextField` stays on an exact render-size path so text does not become thicker or duplicated.

## Repro

[OpenFLGraphicsZoomCacheRepro.zip](https://github.com/user-attachments/files/26484182/OpenFLGraphicsZoomCacheRepro.zip)

It animates zoom over many vector-heavy `Graphics` objects that fall back to the software cache path under OpenGL, with several `TextField`s inside the zooming world to verify text correctness.
On my machine, it went from single-digit FPS during zoom to several hundred FPS after this change.

## Before

- severe FPS drops during zoom
- repeated software rerasterization on small zoom changes

## After

- much more stable FPS during zoom

## Implementation

- add separate raster allocation size tracking for `Graphics`
- reuse larger software cache allocations for non-`TextField` graphics
- keep exact rerasterization behavior for `TextField`
- update Canvas/Cairo software renderers and shape draw paths to use the allocated raster size

### DIFF
--- a/src/openfl/display/DisplayObjectRenderer.hx
+++ b/src/openfl/display/DisplayObjectRenderer.hx
@@ -830,6 +830,15 @@ class DisplayObjectRenderer extends EventDispatcher
 				}
 
 				displayObject.__isCacheBitmapRender = false;
+
+				if (displayObject.__drawableType == TEXT_FIELD && displayObject.__graphics != null)
+				{
+					// Keep TextField software cache state in sync with the freshly
+					// rerasterized cache bitmap, without overriding the broader
+					// object/transform invalidation flow.
+					displayObject.__graphics.__softwareDirty = false;
+				}
+
 			}
 
 			if (updateTransform || needRender)
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -1915,7 +1915,7 @@ import js.html.CanvasRenderingContext2D;
 		}
 	}
 
-	@:noCompletion private function __update(displayMatrix:Matrix, pixelRatio:Float):Void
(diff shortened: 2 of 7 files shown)


Tutez64:
I narrowed the fix.

The initial implementation also changed Cairo/Canvas software rendering behavior, which caused `DisplayObjectTest.test_scrollRect` failures.

The current version limits render-size reuse to the OpenGL software fallback path (`Context3DGraphics`) and keeps Cairo/Canvas on their original exact path.

This keeps the zoom performance win on the target scenario while restoring the software renderer tests.

Tutez64:
Filtered `TextField`s on the OpenGL cache bitmap path could keep `graphics.__softwareDirty` set after rerender, causing repeated software rerasterization and extra memory growth.

This change clears `graphics.__softwareDirty` after a `TextField` cache bitmap rerender.

Result:
- the `Graphics` zoom fix remains
- text rendering stays correct
- memory usage is still slightly above baseline, but the large regression seen before this follow-up is no longer present

Tutez64:
Follow-up to 88a19a008: the cache reuse change exposed a scale9Grid resize issue in the software Graphics path.

Graphics.__update was including the stage/display resize matrix when computing the raster allocation scale. For scale9Grid owners, the final renderer transform already applies that resize scale, so the raster ended up effectively scaled twice and could be clipped during window resizes.

This update skips displayMatrix scaling for __worldScale9Grid owners, keeping the existing behaviour for normal Graphics.

PR #2857, opened 2026-04-18
Expand context3d graphics hardware rendering

Tutez64:
## Summary

This PR expands the `Context3DGraphics` hardware rendering path to reduce unnecessary fallback rendering through Cairo/Canvas.

It includes:
- accepting additional shape commands already supported by the GL path
- ignoring `LINE_STYLE` commands with `thickness == null` in compatibility checks
- rendering simple bitmap-filled line paths in hardware
- reducing collinear bitmap-filled paths to quads
- adding hardware tessellation for a subset of simple color fills
- caching `Graphics` hardware compatibility results until the command stream becomes dirty

The goal is to reduce CPU rasterization, texture uploads, and `graphics-cache` pressure when rendering vector content on native OpenGL targets.

## Motivation

Profiling on native OpenGL showed that a large amount of VRAM churn and upload volume came from `graphics-cache` fallbacks in `Context3DGraphics`, especially for:

- bitmap-filled line paths used by map/ground/UI shapes
- simple color fills built from vector paths
- shapes rejected by `isCompatible()` even though equivalent GL support already existed

By moving these cases onto the hardware path, large software-rendered intermediate `BitmapData` surfaces are avoided.

### Compatibility checks

`Context3DGraphics.isCompatible()` now:
- accepts `DRAW_CIRCLE`, `DRAW_ELLIPSE`, and `DRAW_ROUND_RECT`
- ignores `LINE_STYLE` commands where `thickness == null`

These were previously causing false software fallbacks.

### Bitmap-filled path rendering

Simple bitmap-filled path cases are now rendered through the hardware path, including:
- direct support for simple bitmap-filled line paths
- reduction of collinear bitmap-filled paths to quad geometry

This removes large software fallbacks previously seen for map/ground/UI geometry.

### Color-fill tessellation

A new internal tessellation path handles a subset of simple color-filled vector paths in hardware.
This reduces additional Cairo/Canvas fallback usage for compatible fills.

### Compatibility caching

`Graphics` instances now cache their hardware compatibility result and invalidate it when their command stream becomes dirty, avoiding repeated compatibility analysis every frame.

## Profiling impact

The main result was a large reduction in `graphics-cache` fallback volume and the associated texture uploads / VRAM pressure on OpenGL targets.

## Notes

I also verified this work against a visual regression caused by face culling in tessellated geometry. The correct fix was to use `context.setCulling(Context3DTriangleFace.NONE)` in the tessellated path.

While validating this branch against both `develop` and `9.6.0-dev`, I found a visual regression that can be triggered by the compatibility-caching commit: c4f8c7734d57b37b92dd99e82bacdae6cae22c11
This regression appears to be prevented by the software graphics cache reuse work from #2855, which suggests the visible issue is likely caused by instability in the Cairo/Canvas fallback path during animated transforms rather than by the hardware-path expansion itself.


Tutez64:
Here's a repro: [OpenFLContext3DGraphicsHardwarePathsRepro.zip](https://github.com/user-attachments/files/26862443/OpenFLContext3DGraphicsHardwarePathsRepro.zip)

It contains targeted scenes for the main hardware-path improvements:

- simple bitmap-filled line paths
- collinear bitmap-filled paths reduced to quads
- simple color-filled curve paths handled by the tessellator

On my machine:

| Scene | Upstream develop | This PR only | Full local branch |
| --- | --- | --- | --- |
| Bitmap quad path | ~25 FPS | ~160-240 FPS, ~2-3x lower VRAM | ~260-280 FPS, ~2-3x lower VRAM |
| Bitmap collinear path | ~6 FPS | ~180-220 FPS, ~3-4x lower VRAM | ~240-270 FPS, ~3-4x lower VRAM |
| Color curves | ~70-90 FPS | ~100-120 FPS, ~2x lower VRAM, but with flickering | ~150-180 FPS, ~2x lower VRAM, no artifact |

The “full local branch” numbers include unrelated optimizations from my other PRs, so the important comparison for this PR is `upstream develop` vs `this PR only`.

For scene 3, the flickering seen on the PR-only branch matches the fallback-cache instability mentioned earlier. It is resolved by #2855.

joshtynjala:
> Context3DGraphics.isCompatible() now:
> accepts DRAW_CIRCLE, DRAW_ELLIPSE, and DRAW_ROUND_RECT

Just adding some background about this. At the time that I implemented these, it was intentional to exclude them in `isCompatible`. They require turning on anti-aliasing to look decent, which I'm not sure whether we want to do by default. Additionally, I haven't yet measured if they actually perform better than the Cairo/Canvas path or not. Instead, my main priority was to support a wider range drawing commands for folks who might want to set the `openfl_force_hw_graphics` flag.

Tutez64:
[OpenFLContext3DGraphicsHardwarePathsRepro.zip](https://github.com/user-attachments/files/28179586/OpenFLContext3DGraphicsHardwarePathsRepro.zip)
I updated the repro with 3 additional scenes covering the 2 new commits:
- bitmap-filled compound paths with nested holes
- non-rectangular polygons using beginBitmapFill
- curved paths using beginBitmapFill 

I also fixed the repro’s FPS counter/animation timing. The previous version was tied to stage.frameRate, which could make the displayed FPS and perceived motion misleading depending on the configured frame-rate cap. The repro now uses real delta time for animation.

This version also adds an automated benchmark mode. It runs each scene for a fixed duration with a warmup period, logs frame timing/FPS metrics, and also tries to report per-process VRAM.

With `500` items for measuring FPS, and `2000` for VRAM:

| Scene                       | `develop` | This PR | Local branch | VRAM       |
|-----------------------------|--------------------|-------------|--------------|------------|
| `1` bitmap quad path        | 5 FPS              | 365 FPS     | 640 FPS      | ~5x lower  |
| `2` bitmap collinear path   | 1.5 FPS            | 315 FPS     | 630 FPS      | ~10x lower |
| `3` color curves            | 55 FPS             | 150 FPS     | 375 FPS      | ~2x lower  |
| `4` nested bitmap contours  | 1.5 FPS            | 3 FPS       | 545 FPS      | ~3x lower  |
| `5` polygonal bitmap fills  | 2.5 FPS            | 330 FPS     | 635 FPS      | ~5x lower  |
| `5` bitmap curves           | 3.5 FPS            | 250 FPS     | 635 FPS      | ~3x lower  |

VRAM usage is similar between this PR and my full local branch.

PR #2861, opened 2026-05-10
Fix Cairo scale9 hit test and raster bounds

Tutez64:
Fixes two Cairo `scale9Grid` issues:

- maps Cairo graphics hit testing through `scale9Grid`, so hit areas match scaled graphics
- allocates Cairo software graphics for scale9 content using the local scale used by scale9 drawing, avoiding raster clipping under scaled parents

[openfl-scale9-hit-test.zip](https://github.com/user-attachments/files/27562207/openfl-scale9-hit-test.zip)
[openfl-scale9-raster-clipping.zip](https://github.com/user-attachments/files/27562209/openfl-scale9-raster-clipping.zip)

### DIFF
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -1973,6 +1973,12 @@ import js.html.CanvasRenderingContext2D;
 			}
 		}
 
+		if (__owner.__worldScale9Grid != null)
+		{
+			scaleX = Math.abs(__bitmapScaleX) * pixelRatio;
+			scaleY = Math.abs(__bitmapScaleY) * pixelRatio;
+		}
+
 		#if openfl_disable_graphics_upscaling
 		if (__owner.__worldScale9Grid == null)
 		{
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -1975,8 +1975,8 @@ import js.html.CanvasRenderingContext2D;
 
 		if (__owner.__worldScale9Grid != null)
 		{
-			scaleX = Math.abs(__bitmapScaleX) * pixelRatio;
-			scaleY = Math.abs(__bitmapScaleY) * pixelRatio;
+			scaleX = Math.abs(__owner.scaleX) * pixelRatio;
+			scaleY = Math.abs(__owner.scaleY) * pixelRatio;
 		}
 
 		#if openfl_disable_graphics_upscaling
--- a/src/openfl/display/_internal/CairoGraphics.hx
+++ b/src/openfl/display/_internal/CairoGraphics.hx
@@ -465,6 +465,20 @@ class CairoGraphics
 		{
 			hitTesting = true;
 
+			var scale9Grid:Rectangle = graphics.__owner.__scale9Grid;
+			#if (openfl_legacy_scale9grid && !cairo)
+			var hasScale9Grid:Bool = false;
+			#else
+			// no scale9Grid for masks
+			// no scale9Grid for rotation 0.02 degrees or higher (less than 0.02 is allowed in flash)
+			var hasScale9Grid = scale9Grid != null && !graphics.__owner.__isMask && Math.abs(graphics.__owner.__rotation) < 0.02;
+			#end
+			if (hasScale9Grid)
+			{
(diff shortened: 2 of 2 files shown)


joshtynjala:
> maps Cairo graphics hit testing through scale9Grid, so hit areas match scaled graphics

I confirm that there's a bug here. However, calculating the correct coordinates for `hitTest()` with `scale9Grid` seems to require a different formula than the proposed use of the existing `toScale9Position()` method.

It was easiest for me to see when I drew some lines to mark where the `scale9Grid` is while also ensuring that the mouse cursor changes on roll over:

```haxe
var clip = new Sprite();
clip.graphics.beginFill(0x2E9B57);
clip.graphics.drawEllipse(0, 0, 200, 120);
clip.graphics.endFill();
clip.graphics.lineStyle(1, 0xff0000);
clip.graphics.moveTo(50, 0);
clip.graphics.lineTo(50, 120);
clip.graphics.moveTo(150, 0);
clip.graphics.lineTo(150, 120);
clip.graphics.moveTo(0, 50);
clip.graphics.lineTo(200, 50);
clip.graphics.moveTo(0, 70);
clip.graphics.lineTo(200, 70);
clip.scale9Grid = new Rectangle(50, 50, 100, 20);
clip.scaleX = 2;
clip.scaleY = 2;
clip.buttonMode = true;
clip.useHandCursor = true;
addChild(clip);
```

The hand cursor is displayed at very place than where the lines are drawn. There are also big regions of the fill near the corners where the hand cursor should be displayed, but it isn't.

joshtynjala:
It looks like your solution was almost correct. There needs to be another conversion before calling `toScale9Position()`. The coordinates passed into `hitTest()` are a little funky because they're based on raw `scaleX` and `scaleY` without accounting for `scale9Grid`.

I figured out how to normalize those coordinates passed into `hitTest()`, which can then be passed to `toScale9Position()`. I don't have time to finish up testing today, but I thought I'd mention it so no one else duplicates time looking into this.

joshtynjala:
I just committed 19fd8f28a333e7b7ddf93231f23d09623febaaa6, which adds a method `normalizeHitTestPositionForScale9Grid()` that is called before `toScale9Position()` in `hitTest()`.

It's not perfect yet because it still needs to handle the special case where the scale is small enough that the center region is not rendered and the start and end regions need to be scaled down to fit within the total width or height. It's already a good enough improvement to be worth committing, in my opinion, and that special case can be fixed up later when I have more time to figure out the right math.

twilson90:
I have a fork where I finished all the scale9grid implementation, also completely gutted CanvasGraphics, CairoGraphics and Context3DGraphics in an effort to replicate Flash wherever I noticed a difference.
There's a ton of other changes, but you can maybe cherry pick from mine where it's useful:
https://github.com/twilson90/openfl/tree/gl-graphics

joshtynjala:
Thanks to you both!

Here's the correct (simple!) adjustment for `hitTest()` when there's a valid `scale9Grid`:

```haxe
if (hasScale9Grid)
{
	x *= graphics.__owner.scaleX;
	y *= graphics.__owner.scaleY;
}
```

I'll look into the second issuse and its sample (openfl-scale9-raster-clipping.zip) at a later date.

Tutez64:
Unfortunately, the original repro used zero-based graphics bounds, which accidentally made x *= scaleX appear correct. The updated repro uses negative-origin bounds, as commonly found in button assets with a centered registration point. Current develop fails, while the original toScale9Position() fix from this PR passes. 
[openfl-scale9-hit-test.zip](https://github.com/user-attachments/files/30209371/openfl-scale9-hit-test.zip)

Tutez64:
I retested this fix against the latest develop changes and updated it to
account for 0c80c75db.

The original version used __bitmapScaleX and __bitmapScaleY when calculating
the raster size. These fields are now intentionally updated after
Graphics.__update(), so their values may still be stale during the first
rasterization. In the repro, this caused the unscaled reference buttons to be
clipped at the bottom even though the buttons under a scaled parent were no
longer clipped on the right.

The updated fix reads the current local scale directly from owner.scaleX and
owner.scaleY. This keeps the dirty/managed behaviour introduced by
0c80c75db and correctly sizes the raster on the first render.

joshtynjala:
After my latest commit, develop passes with the new openfl-scale9-hit-test.zip sample.

Tutez64:
This works, thank you.

PR #2863, opened 2026-05-10
Fix dropTarget when dragging over a child of the dragged sprite

Tutez64:
## Summary

Fixes `Sprite.dropTarget` while dragging when the current mouse target is a descendant of the dragged sprite.

Previously, OpenFL ignored the dragged sprite itself when resolving `dropTarget`, but only when `__mouseOverTarget == __dragObject`. If the pointer was over an interactive child of the dragged sprite, `dropTarget` could resolve to that child instead of the object underneath.

This updates the check to treat descendants of `__dragObject` the same as `__dragObject`.

## Repro
[OpenFLDropTargetDraggedChildRepro.zip](https://github.com/user-attachments/files/27571113/OpenFLDropTargetDraggedChildRepro.zip)

### DIFF
--- a/src/openfl/display/Stage.hx
+++ b/src/openfl/display/Stage.hx
@@ -3252,7 +3252,7 @@ class Stage extends DisplayObjectContainer #if lime implements IModule #end
 
 			var dropTarget:DisplayObject = null;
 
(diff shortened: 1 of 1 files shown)


joshtynjala:
Thank you!

PR #2866, opened 2026-05-11
CairoGraphics: match surface size exactly, drop 1.25x upscale margin

AxGord:
## Summary

The Cairo render path allocates `bitmap = width × height × 1.25` when
the previous surface is too small, and reuses any surface that satisfies
`width <= surface.width && height <= surface.height`. After a frame
where bounds grew (or after the legacy 1.25× upscale margin kicked in),
the surface stays oversized; Cairo paints only into the `width × height`
sub-region, and the empty padding renders into the final composited
output as transparency on the right/bottom edges of the shape.

This PR makes the surface match `width × height` exactly: reallocate
on any size mismatch, and drop the 1.25× upscale margin.

## The bug

In `CairoGraphics.render()`, surface reuse and allocation looked like:

```haxe
var needsUpscaling = false;

if (graphics.__cairo != null)
{
    var surface:CairoImageSurface = cast graphics.__cairo.target;

    if (width > surface.width || height > surface.height)
    {
        graphics.__cairo = null;
        needsUpscaling = true;
    }
}

if (graphics.__cairo == null || graphics.__bitmap == null)
{
    var bitmapWidth  = needsUpscaling ? Std.int(width * 1.25)  : width;
    var bitmapHeight = needsUpscaling ? Std.int(height * 1.25) : height;
    ...
}
```

Two related problems:

1. **Reuse with shrinking content.** If the previous frame produced
   `surface = 250×H` and the current frame has `width=200`, the
   `width > surface.width` check is `false`, so we keep the 250×H
   surface. Cairo then paints content into the `200×H` sub-region.
   `Context3DShape` composites the cached bitmap back into the
   framebuffer using `__worldTransform`, which is sized to the bitmap,
   not to the painted sub-region — the right 50 pixels of empty padding
   end up where content should be, visually clipping the shape.

2. **1.25× upscale margin amplifies the problem.** Even on the first
   reallocation after a growth, the surface is intentionally allocated
   25% larger than needed, so the very next frame already has an
   oversized surface and the above clipping pattern starts.

We hit this on a Context3DRenderer-driven workload (TextField caches
and scale-9 rounded-rect Shapes) where bounds oscillate as content
re-flows.

## Fix

```haxe
if (graphics.__cairo != null)
{
    var surface:CairoImageSurface = cast graphics.__cairo.target;

    if (width != surface.width || height != surface.height)
    {
        graphics.__cairo = null;
    }
}

if (graphics.__cairo == null || graphics.__bitmap == null)
{
    var bitmapWidth = width;
    var bitmapHeight = height;
    // existing Graphics.maxTextureWidth/Height clamping retained (per #2847)
    ...
}
```

- `>` → `!=` so a smaller-than-current shape forces reallocation
  instead of painting into an oversized surface.
- 1.25× margin gone; surface is allocated to exactly `width × height`,
  still clamped to `Graphics.maxTextureWidth/Height` from #2847.

## Trade-off

Rapid frame-to-frame resizes now reallocate every frame instead of
every fifth (when bounds grew by ≤25%). In practice
`new BitmapData()` is cheap on the Cairo path — one arena allocation
backing the image surface — and visual correctness on every frame
outweighs the saved allocation. No other code in CairoGraphics depends
on the surface being larger than the current content (verified by
grep for `surface.width` / `surface.height`).

## Test

Type-checked: `haxe --no-output -cp src -lib lime -D openfl=9.5.1 --macro 'include("openfl.display._internal")'` succeeds.

No deterministic repro app — Cairo path activates via specific
composite chains (`BitmapData.draw` with the software renderer,
Context3DShape compositing) rather than the default stage render, so
a standalone demo would need a contrived setup that wouldn't add
clarity beyond the explanation above. The fix was verified by running
our internal Context3DRenderer-driven app on this branch: scale-9
rounded-rect right/bottom edges no longer clip after layout shrinks.
Happy to add a sample if reviewers want one.

## Risk

- Single-render-pass shapes whose bounds never change: identical
  behaviour — `width == surface.width` on every frame after the first.
- Shapes with constantly-changing bounds: one extra `new BitmapData()`
  per frame compared to the 1.25× margin (which amortised over ≤5
  frames). In our profiling this didn't move the needle.
- Interaction with #2847 (`Graphics.maxTextureWidth/Height` clamp):
  unchanged — clamp logic is preserved verbatim.

### DIFF
--- a/src/openfl/display/Bitmap.hx
+++ b/src/openfl/display/Bitmap.hx
@@ -98,6 +98,11 @@ class Bitmap extends DisplayObject
 	#end
 	@:noCompletion private var __bitmapData:BitmapData;
 	@:noCompletion private var __imageVersion:Int;
+	// Painted sub-region (-1 = use full bitmapData dims). Set when this Bitmap
+	// instance is `DisplayObject.__cacheBitmap` and the backing data is sized
+	// larger than the current cache footprint (high-water-mark strategy).
+	@:noCompletion private var __paintedWidth:Int = -1;
+	@:noCompletion private var __paintedHeight:Int = -1;
 
 	#if openfljs
 	@:noCompletion private static function __init__()
--- a/src/openfl/display/BitmapData.hx
+++ b/src/openfl/display/BitmapData.hx
@@ -227,6 +227,8 @@ class BitmapData implements IBitmapDrawable
 	@:noCompletion private var __vertexBufferScaleX:Float;
 	@:noCompletion private var __vertexBufferScaleY:Float;
 	@:noCompletion private var __vertexBufferWidth:Float;
+	@:noCompletion private var __vertexBufferPaintedWidth:Int = -1;
+	@:noCompletion private var __vertexBufferPaintedHeight:Int = -1;
 	@:noCompletion private var __worldAlpha:Float;
 	@:noCompletion private var __worldColorTransform:ColorTransform;
 	@:noCompletion private var __worldTransform:Matrix;
@@ -1656,7 +1658,8 @@ class BitmapData implements IBitmapDrawable
 		@param	context	A Stage3D context
 		@returns	A VertexBuffer3D object for use with rendering
 	**/
-	@:dox(hide) public function getVertexBuffer(context:Context3D, scale9Grid:Rectangle = null, targetObject:DisplayObject = null):VertexBuffer3D
+	@:dox(hide) public function getVertexBuffer(context:Context3D, scale9Grid:Rectangle = null, targetObject:DisplayObject = null,
+			paintedWidth:Int = -1, paintedHeight:Int = -1):VertexBuffer3D
 	{
 		var gl = context.gl;
 
@@ -1667,6 +1670,8 @@ class BitmapData implements IBitmapDrawable
 			|| __vertexBufferContext != context.__context
 			|| (scale9Grid != null && __vertexBufferGrid == null)
 			|| (__vertexBufferGrid != null && !__vertexBufferGrid.equals(scale9Grid))
+			|| __vertexBufferPaintedWidth != paintedWidth
+			|| __vertexBufferPaintedHeight != paintedHeight
 			|| (targetObject != null
 				&& (__vertexBufferWidth != targetObject.width
 					|| __vertexBufferHeight != targetObject.height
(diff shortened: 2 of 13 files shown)


joshtynjala:
Keeping around a larger surface than the content, to allow scaling without reallocating, feels strongly to me like an intentional design decision. Likely one based on real-world experience. If I were to merge a change that completely removes that behavior, I'd want to also add a new define to either opt-in or opt-out, at least for a while. It's not a risk-free change. Often, we discover important edge cases from the community only after releasing an update.

>  In practice new BitmapData() is cheap on the Cairo path

Maybe the allocation alone. However, in general, this is not my experience on any target. Short-lived allocations can lead to a lot of garbage collection that really hurts performance. I've observed tweens/animations involving scaling that don't perform particularly well in OpenFL compared to Flash, and I'd expect those to get worse if this were merged.

Additionally, it's probably worth mentioning that I recently merged #2852 to address an issue with temporary `BitmapData` objects hurting performance. It's not the exact same situation, but it at least adds some context where too much BitmapData churn did affect performance. I've also recently been optimizing other parts of OpenFL that had a lot of temporary object churn, so I'm particularly sensitive to new churn that might cancel out that hard work.

> Context3DShape composites the cached bitmap back into the framebuffer using __worldTransform, which is sized to the bitmap, not to the painted sub-region — the right 50 pixels of empty padding end up where content should be, visually clipping the shape.

Based on description alone, it seems to me like this particular issue could be solved while also keeping the behavior where we still at least sometimes allocate a larger BitmapData than a shape might currently require. But I haven't dived into that part of the code recently.

> a standalone demo would need a contrived setup that wouldn't add clarity beyond the explanation above
> Happy to add a sample if reviewers want one.

Please do. A sample that reproduces the issue almost always adds value. Even if it is contrived. Often, just having a way to reproduce quickly is a huge win for a reviewer. It can be hard to recreate an environment to reproduce from scratch because there are often assumptions that inadvertently get omitted.

Thanks!

AxGord:
@joshtynjala — you were right on both counts. The allocation cost is real and the visual repro request was reasonable. Pushed a follow-up commit that takes the approach you suggested ("solve while keeping the larger BitmapData behavior") and extends it to a second instance of the same pattern I hadn't noticed initially.

**Composite-layer decoupling.** `BitmapData.getVertexBuffer` gains optional `paintedWidth` / `paintedHeight` parameters. The non-scale9 quad uses them for vertex positions, and the UV span is scaled by `paintedWidth / bitmap.width` so the rendered quad covers only the painted sub-region. `Context3DShape.render` / `renderMask` pass `graphics.__width / __height`; `Context3DBitmap.render` passes `bitmap.__paintedWidth / __paintedHeight` (defaults to `-1` = full bitmap for direct Bitmap rendering). With this, an oversize cached surface never leaks transparent padding into world space.

**Second instance of the same pattern.** `DisplayObjectRenderer.__updateCacheBitmap` has the same `>` + `Math.max(width * 1.25, current)` for `__cacheBitmapData` (the cacheAsBitmap / filters cache). Without addressing that, the Cairo-level fix is half a fix. `Bitmap` now carries `__paintedWidth` / `__paintedHeight`, set in `__updateCacheBitmap` whenever the backing data is larger than the current filter footprint.

**High-water on top of the decoupling.** Now that an oversize surface is safe to keep, all four allocation sites (`CairoGraphics`, `CairoTextField`, `DisplayObjectRenderer.__updateCacheBitmap`) use a high-water mark: grow on demand to exactly the new width / height, hold the peak on shrink. Memory plateaus at the actual high-water rather than 1.25x of it, and there is no per-frame reallocation when bounds oscillate.

**Numbers.** Measured on a 240-frame schedule running `scale 1 → 3 → 1` (M-series mac, hxcpp release) via the sample under `tests/cairo-surface-clip/`:

| | upstream (`>` + 1.25x) | exact-fit (first commit in this PR) | final (high-water + decoupled) |
|---|---|---|---|
| `CairoGraphics` allocs | 4 | 6 | **4** |
| `cacheBitmapData` allocs | 2 | 3 | **2** |
| memory headroom | 1.25x | exact | exact |

Allocation counts match the legacy `>` + 1.25x policy without the 1.25x memory overhead.

**Tests and sample.** `tests/graphics/src/CairoSurfaceSizeTest.hx` covers both the Cairo shape surface (3 cases: high-water hold on shrink, growth on grow, opaque right-edge after shrink) and the parallel `CairoTextField` surface (2 cases). Tests drive the software path via `BitmapData.draw` and assert the high-water invariant directly. `tests/cairo-surface-clip/` is a small interactive app that captures the GL framebuffer via `Context3D.drawToBitmapData` and writes per-frame allocation counts behind a `-D openfl_count_surface_allocs` diagnostic flag.

One honest note about the sample: it exercises and verifies the cache state (oversize surface held across shrink, allocation counts) but does **not** reproduce a visible right-edge clip on screen. The oversize cache's transparent overflow against a uniform backdrop is pixel-identical to a correctly-fit cache in a straight on-screen GL render. The clip is analytically derived from the composite path consuming the full bitmap dimensions via `getVertexBuffer`, and is latent until a downstream consumer (off-screen render-to-texture, viewport with non-trivial UV geometry) makes the transparent overflow visible. I went deep on trying to land a visual repro and could not get the synthetic case to crack it; the unit tests assert the invariant directly, which is what actually catches regressions.

PR #2867, opened 2026-05-11
CairoGraphics: reset __bitmapScaleX/Y after the __softwareDirty early-return

AxGord:
## Summary

`CairoGraphics.render()` resets `graphics.__bitmapScaleX/Y` to `1`
(or `__owner.scaleX/Y` for scale9Grid) **before** the
`if (!graphics.__softwareDirty || graphics.__managed) return;`
early-return. On every cache-hit frame the function overwrites the
correct bitmapScale value with `1` and then bails out, breaking
downstream code that depends on the pre-existing scale.

The most visible symptom is **TextFields with filters appearing at 2×
size on Retina/HiDPI screens** after the first cached frame. This PR
moves both the scale9Grid branch and the plain reset *after* the
early-return so they only run when the Cairo commands are actually
re-rendered.

## The bug

In `CairoGraphics.hx` around line 1844 (upstream/develop):

```haxe
if (hasScale9Grid)
{
    graphics.__bitmapScaleX = graphics.__owner.scaleX;
    graphics.__bitmapScaleY = graphics.__owner.scaleY;
}
else
{
    graphics.__bitmapScaleX = 1;   // ← reset here…
    graphics.__bitmapScaleY = 1;
}

graphics.__update(renderer.__worldTransform, pixelRatio);

if (!graphics.__softwareDirty || graphics.__managed)
{
    CairoGraphics.graphics = null;
    return;                         // …but the reset already happened
}
```

Why this matters for text:

1. `CairoTextField` renders a TextField's glyphs to its `__cairo`
   bitmap at `pixelRatio` resolution (2× on Retina) and writes
   `graphics.__bitmapScaleX = pixelRatio` so the composite path knows
   the bitmap is "1 pixel per device pixel, not per stage point".
2. `CairoShape` later composites that bitmap back into the framebuffer
   with `1 / __bitmapScaleX` applied to the transform — yielding the
   intended 1:1 device pixel mapping.
3. Every subsequent render frame, `CairoGraphics.render()` runs first
   (called from the regular graphics path) and, on a cache hit, the
   above block writes `__bitmapScaleX = 1` and then early-returns.
   The composite step now multiplies by `1/1 = 1` and the 2×-pixel
   bitmap renders at 2× its intended size.

The bug only manifests once a cached frame is reused — i.e. after the
first dirty render of a TextField with filters or `cacheAsBitmap=true`.

## Fix

Move both the scale9Grid branch and the reset branch *after* the
early-return so they only execute when the Cairo commands are
re-rendered:

```haxe
graphics.__update(renderer.__worldTransform, pixelRatio);

if (!graphics.__softwareDirty || graphics.__managed)
{
    CairoGraphics.graphics = null;
    return;
}

// __bitmapScaleX/Y must only be reset when we are actually re-rendering
// the Cairo commands. ...
if (hasScale9Grid) { ... } else { graphics.__bitmapScaleX = 1; ... }
```

Behaviour on the re-render path is unchanged — `__bitmapScaleX/Y` end
up at the same values they did before the reorder.

## Tests

New `tests/graphics/src/CairoTextBitmapScaleTest.hx` (registered in
`Tests.hx`) with two cases:

- `test_cacheHitPreservesNonOneBitmapScale` — sets bitmapScale to 2.0,
  marks `__softwareDirty = false`, calls `CairoGraphics.render()`,
  asserts bitmapScale is still 2.0.
- `test_managedGraphicsAlsoPreservesBitmapScale` — same with
  `__managed = true` instead (the other half of the early-return
  condition).

Both use an empty `Graphics` (no draw commands → `__bounds == null`),
which makes `Graphics.__update` return early without invalidating
`__softwareDirty`, isolating the bitmapScale-reset behaviour from
unrelated update-path side effects.

Verified the tests catch the regression: reverting CairoGraphics.hx
to upstream/develop and running the suite produces 4 failures (both
test methods, two assertions each). With the patch applied, all 9
test classes (69 assertations) pass in ~2ms on neko.

## Visual repro (not part of the diff)

A minimal native repro is a Sprite/TextField with a filter and
`cacheAsBitmap = true` on a HiDPI window, e.g.:

```haxe
var tf = new TextField();
tf.defaultTextFormat = new TextFormat("_sans", 24, 0xFFFFFF);
tf.text = "Retina text — should stay 24pt";
tf.filters = [new GlowFilter(0xFF0000, 1, 4, 4, 1)];
tf.cacheAsBitmap = true;
addChild(tf);
```

On `develop`: frame 1 renders at the correct 24pt with a red glow;
from frame 2 onward (cache hit) the text is composited at 2× and
appears ~48pt and blurry. On this branch: text stays 24pt across
all frames.

(I tried producing matching before/after PNGs from a `lime run mac`
demo, but the native app blocks the run process and won't quit on
`openfl.system.System.exit()` from a background launch — happy to
provide screenshots if a reviewer wants them.)

## Risk

- Single-render path: no behaviour change — values end up identical.
- Cache-hit path: previously incorrectly clobbered `__bitmapScaleX/Y`,
  now preserved.
- Public API: unchanged.
- Affected: any code that writes a non-1 `__bitmapScaleX/Y` and relies
  on the value surviving across cached `render()` calls. In
  upstream OpenFL the only documented writer is `CairoTextField`
  (text caching path), which is the bug's source.

### DIFF
--- a/tests/graphics/src/CairoTextBitmapScaleTest.hx
+++ b/tests/graphics/src/CairoTextBitmapScaleTest.hx
@@ -86,7 +86,7 @@ class CairoTextBitmapScaleTest extends Test
 {
 	public function test_skippedOnNonCairo()
 	{
-		Assert.isTrue(true);
+		Assert.pass();
 	}
 }
 #end
--- a/tests/graphics/src/CairoTextBitmapScaleTest.hx
+++ b/tests/graphics/src/CairoTextBitmapScaleTest.hx
@@ -80,9 +80,13 @@ class CairoTextBitmapScaleTest extends Test
 	}
 }
 #else
+import utest.Assert;
 import utest.Test;
 class CairoTextBitmapScaleTest extends Test
 {
-	public function test_skippedOnNonCairo() {}
+	public function test_skippedOnNonCairo()
+	{
+		Assert.isTrue(true);
+	}
 }
 #end
--- a/tests/graphics/src/CairoTextBitmapScaleTest.hx
+++ b/tests/graphics/src/CairoTextBitmapScaleTest.hx
@@ -1,0 +1,88 @@
+package;
+
+#if (!flash && lime_cairo)
+import lime.graphics.cairo.Cairo;
+import openfl.display.BitmapData;
+import openfl.display.CairoRenderer;
+import openfl.display.Sprite;
+import openfl.display._internal.CairoGraphics;
+import openfl.geom.Matrix;
+import utest.Assert;
+import utest.Test;
+
+/**
+	Regression test for the Retina text-doubling bug in CairoGraphics.render():
+	when a TextField (via CairoTextField) writes a non-1 bitmapScale (= pixelRatio,
+	e.g. 2 on Retina) and the graphics object then renders with `__softwareDirty=false`
+	(a cache hit), the early-return must NOT have reset `__bitmapScaleX/Y` to 1
+	beforehand — otherwise CairoShape composites the high-DPI text bitmap without
+	the corresponding 1/pixelRatio downscale and the text visually doubles.
+
+	The test uses an empty graphics (no draw commands → `__bounds == null`), so
+	`Graphics.__update` returns early without touching `__dirty` / `__softwareDirty`.
+	That isolates the bitmapScale-reset behaviour from the unrelated update path.
+**/
+class CairoTextBitmapScaleTest extends Test
+{
+	private function buildRenderer(pixelRatio:Float):CairoRenderer
+	{
+		var bmd = new BitmapData(10, 10, true, 0);
+		var cairo = new Cairo(bmd.getSurface());
+		var renderer = @:privateAccess new CairoRenderer(cairo);
+		@:privateAccess renderer.__pixelRatio = pixelRatio;
+		@:privateAccess renderer.__worldTransform = new Matrix();
+		return renderer;
+	}
+
+	public function test_cacheHitPreservesNonOneBitmapScale()
+	{
+		var sprite = new Sprite();
+		// Force-create the Graphics instance via the getter, but draw nothing
+		// so __bounds stays null and Graphics.__update early-returns without
+		// invalidating __softwareDirty.
+		sprite.graphics;
+
+		var g = @:privateAccess sprite.__graphics;
+
+		// Simulate CairoTextField setting a Retina pixelRatio on the cached bitmap.
+		@:privateAccess g.__bitmapScaleX = 2.0;
+		@:privateAccess g.__bitmapScaleY = 2.0;
+
+		// Simulate a cache hit: nothing in the graphics changed, so render() must
+		// early-return without touching bitmapScale.
+		@:privateAccess g.__softwareDirty = false;
+		@:privateAccess g.__managed = false;
+
+		@:privateAccess CairoGraphics.render(g, buildRenderer(2.0));
+
+		// With the fix: the early-return preserves the 2.0 we set.
+		// Without the fix: render() resets bitmapScaleX/Y to 1 before the early-return.
+		Assert.equals(2.0, @:privateAccess g.__bitmapScaleX);
+		Assert.equals(2.0, @:privateAccess g.__bitmapScaleY);
+	}
+
+	public function test_managedGraphicsAlsoPreservesBitmapScale()
+	{
+		// Same scenario, but __managed=true (the other half of the early-return condition).
+		var sprite = new Sprite();
+		sprite.graphics;
+		var g = @:privateAccess sprite.__graphics;
+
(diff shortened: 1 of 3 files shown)


joshtynjala:
I tried testing the same fix on CanvasGraphics, and when a filter is applied, the TextField renders at half the expected size. But with cacheAsBitmap = true and no filters, the TextField renders at the correct size. Weird. There seems to be some kind of difference between Cairo and Canvas renderers.

I need to switch to my Mac to test Cairo because HiDPI doesn't work well on my Linux machine for native/cairo targets.

joshtynjala:
All looks good with Cairo on my Mac. Tested with allow-high-dpi=true/false, with/without filters, and with/without cacheAsBitmap (no filters).

So I guess that there's something funky with the Canvas renderer, possibly related to filters.

tests/graphics/src/CairoTextBitmapScaleTest.hx
```diff
@@ -80,9 +80,13 @@ class CairoTextBitmapScaleTest extends Test
...
 }
 #else
+import utest.Assert;
 import utest.Test;
 class CairoTextBitmapScaleTest extends Test
 {
-	public function test_skippedOnNonCairo() {}
+	public function test_skippedOnNonCairo()
+	{
+		Assert.isTrue(true);
```
joshtynjala: I discovered a while back that utest offers an `Assert.pass()` method that is designed for this exact situation where you don't have any real assertions.

https://github.com/haxe-utest/utest/blob/8e99ed9/src/utest/Assert.hx#L845-L853

AxGord:
Quick follow-up on the Canvas half-size symptom. Reproduced the html5 path on macOS + Chromium with emulated `devicePixelRatio = 2` (override before lime boots), four TextField variants on stock `origin/develop`: plain, `cacheAsBitmap=true`, `GlowFilter` only, `GlowFilter + cacheAsBitmap=true`.

Stock upstream renders all four at the expected size at both DPR=1 and DPR=2. No half-size on the text. The glow halo radius does differ across DPR (looser at DPR=2, tighter at DPR=1, which makes sense given how filter offsets feed into the cache-bitmap allocation), but nothing dramatically clipped.

Reading your comment again — *"I tried testing the same fix on CanvasGraphics, … the TextField renders at half the expected size"* — it sounds like the half-size showed up only after porting the reorder to Canvas. Which tracks structurally: upstream `CanvasGraphics.render()` has no `__bitmapScale = 1` reset block, so there is nothing to move past an early-return; inserting an analogous reset would clobber the value `CanvasTextField` writes (`__bitmapScale = pixelRatio` at the end of its render) before `CanvasShape` consumes `1 / __bitmapScale` at composite time.

Tentative read: this PR's fix is correctly Cairo-local, and a literal port to Canvas isn't the right shape. If you'd like, I can publish the minimal html5 repro project (project.xml + ~50-line `Main.hx`) so any future Canvas-side investigation has a fixed starting point — happy to file it as a separate issue too.

joshtynjala:
> upstream CanvasGraphics.render() has no __bitmapScale = 1 reset block, so there is nothing to move past an early-return

I think you missed it here:

https://github.com/openfl/openfl/blob/79b06e1/src/openfl/display/_internal/CanvasGraphics.hx#L2004-L2013

AxGord:
@joshtynjala You're right — apologies, I missed that block. The same pre-`softwareDirty`-gate `__bitmapScaleX/Y` reset exists on Canvas at the lines you pointed to. My earlier "no reset block" claim was wrong, and my observation that html5 looked correct on stock was a different mechanism than I described.

**What actually happens on stock Canvas:** the reset clobbers `__bitmapScale` to `1` every frame; on the next call `CanvasTextField:99-105` detects `__bitmapScale != pixelRatio` and forces `__softwareDirty = true`, which triggers a full glyph re-render *every* frame and restores the value. So stock html5 is visually correct but pays an avoidable per-frame redraw.

**Why a literal port of this PR's reorder gives half-size on Canvas:** Canvas's composite path uses `drawImage(canvas, 0, 0, W, H)`, which already encodes source-pixel-density compression in the `(W, H)` dest-size args — unlike Cairo's `cairo.paint()` which is 1:1 source-to-dest through the current matrix. On Canvas, `__bitmapScale` must be `1` at composite time for non-scale9-grid paths; the stock reset was load-bearing for that, and removing it without further changes leaves `__bitmapScale = pixelRatio` from `CanvasTextField`, so `scale(1/pixelRatio)` halves the rendered size.

I opened #2869 with a Canvas-side fix that uses the `__managed` contract instead (matching the convention `CairoTextField`/`CairoGraphics` already share). Scoped to the Canvas (2D) renderer only — OpenGL is untouched, same as this PR. It also keeps `test_scale9Grid` green and ships a structural + visual integration test pair.

Sorry for the confusion on the first read.

PR #2870, opened 2026-05-20
Fix OpenFL window fps being ignored with Lime frame timing

Tutez64:
## Summary

OpenFL overrides Lime's `createWindow()` path and constructs `openfl.display.Window` directly, so it bypassed Lime's new `__seedFrameConfiguration(attributes)` call (https://github.com/openfl/lime/commit/a5d2a49dc1ff8279be31689e2fc94ddfe09acc17). As a result, the main loop stayed at Lime's default 60 FPS even when `project.xml` specified another window FPS.

## Changes

- Return early when the created window is invalid, matching Lime's default `__createWindow()` behavior.
- Seed application frame timing from the generated window attributes in OpenFL's `createWindow()` override.

## Repro

Any OpenFL project with a custom `fps` flag value will work.

### DIFF
--- a/src/openfl/display/Application.hx
+++ b/src/openfl/display/Application.hx
@@ -60,6 +60,8 @@ class Application #if lime extends LimeApplication #end
 	public override function createWindow(attributes:WindowAttributes):Window
 	{
 		var window = new Window(this, attributes);
+		if (window.id == -1) return null;
+		__seedFrameConfiguration(attributes);
 
 		#if (!flash && !macro)
 		if (Lib.current.__loaderInfo.width == -1)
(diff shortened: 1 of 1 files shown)


Tutez64:
I updated the fix to be gated behind `#if (lime >= "8.4.0")`.

dimensionscape:
Thanks for tracking this down.

I’d rather not make OpenFL responsible for calling Lime’s internal frame configuration hook directly. That would make this a compatibility requirement in OpenFL, and older OpenFL releases would still be broken against the newer Lime behavior.

I fixed this on the Lime side instead: https://github.com/openfl/lime/commit/ff8ad734074c4c964b8b25fb317ee06bbbc29b14

Lime now seeds the frame timing when a window is created directly, so OpenFL’s existing window creation path keeps working without needing to know about Lime internals. I also verified the specific case with an OpenFL project using `<window fps="23" />`.

Closing this in favor of the Lime fix.

PR #2871, opened 2026-05-23
Fix TextField selection stuck when target stops mouseUp propagation

Tutez64:
## Summary

Fixes a Flash/AIR parity issue where `TextField` text selection could remain in a "mouse drag" state after the button was released.

When application code adds a target-phase `mouseUp` listener on a `TextField` and calls `stopImmediatePropagation()` (a pattern used in Flash UIs to isolate clicks from gameplay), OpenFL's `stage_onMouseUp` handler never ran because it was registered on the stage in the **bubble** phase. The stage `MOUSE_MOVE` listener was never removed, so moving the pointer continued to extend the selection without `buttonDown`.

AIR did not show this behavior with the same listener setup.

## Fix

Register and remove `stage_onMouseUp` with `useCapture = true` so selection teardown runs during the capture phase, before target listeners can stop propagation.

## Repro

[OpenFLTextFieldMouseUpCaptureRepro.zip](https://github.com/user-attachments/files/28170360/OpenFLTextFieldMouseUpCaptureRepro.zip)

### DIFF
--- a/src/openfl/text/TextField.hx
+++ b/src/openfl/text/TextField.hx
@@ -3381,7 +3381,7 @@ class TextField extends InteractiveObject
 
 		stage.removeEventListener(Event.ENTER_FRAME, this_onEnterFrame);
 		stage.removeEventListener(MouseEvent.MOUSE_MOVE, stage_onMouseMove);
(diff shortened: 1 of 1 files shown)


joshtynjala:
Thank you!

PR #2875, opened 2026-06-14
Add tags to haxelib

Geokureli:
OpenFL doesn't show up when you search "openfl" in lib.haxe.org: https://lib.haxe.org/t/openfl/


dimensionscape:
Thanks

PR #2878, opened 2026-07-03
Context3D: remove redundant anisotropy extension checks on html5

pngdrift:
The checks did not work as expected and caused the extension to remain null. `Reflect.hasField(extension, "MAX_TEXTURE_MAX_ANISOTROPY_EXT")` always returns false because it on html5 is transpiled to `Object.prototype.hasOwnProperty.call(extension, "MAX_TEXTURE_MAX_ANISOTROPY_EXT")` but the extension object exposes constants via the prototype instead of own properties.

### DIFF
--- a/src/openfl/display3D/Context3D.hx
+++ b/src/openfl/display3D/Context3D.hx
@@ -324,10 +324,8 @@ import lime.math.Vector2;
 			var extension:Dynamic = gl.getExtension("EXT_texture_filter_anisotropic");
 
 			#if (js && html5)
-			if (extension == null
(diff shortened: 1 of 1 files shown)


pngdrift:
To check, here is what the following code outputs
```js
var canvas = document.createElement("canvas");
var gl = canvas.getContext("webgl");

var ext = gl.getExtension("EXT_texture_filter_anisotropic");
console.log(Object.prototype.hasOwnProperty.call(ext, "MAX_TEXTURE_MAX_ANISOTROPY_EXT"));
console.log(ext);
```
in Firefox ESR 140:
Firefox output
Edge 149:
Edge output

PR #2880, opened 2026-07-22
Clip Context3D software fallback graphics to the visible viewport

Tutez64:
## Problem

`Context3DGraphics` falls back to Cairo or Canvas when a `Graphics` instance
contains commands that are not supported by the hardware path.

The software renderer currently rasterizes the complete graphics bounds, even
when most of the geometry is outside the visible viewport. Large transparent
or otherwise invisible geometry can therefore create an enormous intermediate
surface and increase memory usage by one or more GiB.

This differs significantly from Flash, where the same content does not require
a full-size offscreen allocation.

## Fix

This PR limits software-rendered graphics to the intersection of:

- the graphics bounds
- the visible renderer viewport
- the active clip rectangle, when present

The renderer-space clipping bounds are converted back to the graphics local
coordinate space before rasterization. The resulting origin is preserved when
the bitmap is rendered, and the software cache is invalidated whenever its
clipped bounds change.

Graphics using scale9 grids retain the previous behavior because their
coordinate transformations require separate handling.

## Reproduction
[OpenFLContext3DSoftwareFallbackClippingRepro.zip](https://github.com/user-attachments/files/30249787/OpenFLContext3DSoftwareFallbackClippingRepro.zip)

It creates a single `Graphics` instance containing:

- a transparent `21920 × 21080` outlined rectangle that forces the software
  fallback
- a visible panel confirming that the same graphics instance rendered
  successfully

Before this change, OpenFL attempts to allocate a software surface covering the
complete graphics bounds, causing a very large memory increase and potentially
a slow or failed startup.

After this change, the software surface is limited to the visible viewport and
memory usage remains bounded by the window size.

## Testing

The graphics test suite passes with:

- 90 assertions
- 84 successes
- 0 errors
- 0 failures

Additional coverage verifies:

- clipping to local graphics bounds
- transformed viewports
- empty intersections
- preservation of the raster origin
- software-cache invalidation when clipping bounds change

### DIFF
--- a/src/openfl/display/Graphics.hx
+++ b/src/openfl/display/Graphics.hx
@@ -74,6 +74,7 @@ import js.html.CanvasRenderingContext2D;
 	@:noCompletion private var __renderTransform:Matrix;
 	@:noCompletion private var __shaderBufferPool:ObjectPool<ShaderBuffer>;
 	@:noCompletion private var __softwareDirty:Bool;
+	@:noCompletion private var __softwareRenderBounds:Rectangle;
 	@:noCompletion private var __strokePadding:Float;
 	@:noCompletion private var __transformDirty:Bool;
 	@:noCompletion private var __triangleIndexBuffer:IndexBuffer3D;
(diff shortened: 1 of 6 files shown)


Tutez64:
Closing this in favor of #2879.

Clipping the software fallback bounds avoids excessively large bitmap allocations, but changes to the clipped bounds may invalidate the cache and trigger repeated CPU rasterization, causing a significant runtime performance regression.

#2879 addresses the original case without this trade-off by keeping overlapping `drawRect` fills on the Context3D path while preserving the `evenOdd` fill rule exactly.

PR #2883, opened 2026-07-27
Allocate DisplayObject render transforms lazily

Tutez64:
## Summary
`DisplayObject` eagerly allocates `__worldTransform` and `__renderTransform` in its constructor, even when the object is never updated or rendered.

This removes those eager allocations and lets the existing transform update path create the matrices when they are first needed.

Additional safeguards handle objects that reach a renderer before their first display-list update, including objects added between the update and render phases. Internal cache bitmaps are also initialized before their transform matrices are accessed.

The constructor-state check used by `Sprite.__bind()` now relies on `__worldColorTransform`, since a null world matrix is valid after construction.

## Tests

Added coverage verifying that:

- newly created display objects keep both matrices null;
- coordinate conversion initializes the complete parent chain;
- rendering an object before its first update initializes its transforms;
- internal cache bitmaps are created with usable transform state.

The display object test suite passes on Neko and C++ with 1,475 assertions on each target.

## Real-world observation

These values should be considered indicative rather than a controlled benchmark.

| Approximate checkpoint | Before | After | Difference |
|---|---:|---:|---:|
| 160 seconds | 60,643 matrices | 43,647 matrices | -16,996 (-28.0%) |
| 400 seconds | 76,648 matrices | 54,009 matrices | -22,639 (-29.5%) |

The later checkpoint represents approximately 1.38 MiB less shallow `Matrix` object memory.

Total GC time remained effectively unchanged in these runs (0.812% before and 0.807% after).

### DIFF
--- a/src/openfl/display/CairoRenderer.hx
+++ b/src/openfl/display/CairoRenderer.hx
@@ -183,6 +183,7 @@ class CairoRenderer extends DisplayObjectRenderer
 	@:noCompletion private function __renderDrawable(object:IBitmapDrawable):Void
 	{
 		if (object == null) return;
+		if (object.__drawableType != BITMAP_DATA) __ensureRenderTransform(cast object);
 
 		switch (object.__drawableType)
 		{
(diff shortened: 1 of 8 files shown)


Tutez64:
Follow-up:
  `__getWorldTransform` stopped before the Stage, so the first pending mouse hit-test could combine against null
  `__worldTransform` / `__renderTransform` before the first render. That caused intermittent native startup
  crashes on Windows (not Linux?):  access violation / exit `0xffffffff` in `__updateTransforms`.
  This includes the Stage in the update chain, defensively initializes null parent transforms in
  `__updateTransforms`, and adds a regression test.

PR #2886, opened 2026-08-23
Add SecurityPanel and a no-op Security.showSettings

Tutez64:
- Add `openfl.system.SecurityPanel` (and the matching Flash extern).
- Implement `Security.showSettings` as a no-op on non-Flash targets, which is the AIR behavior.

### DIFF
--- a/lib/flash-externs/src/flash/system/SecurityPanel.hx
+++ b/lib/flash-externs/src/flash/system/SecurityPanel.hx
@@ -1,15 +1,15 @@
 package flash.system;
 
 #if flash
-extern class SecurityPanel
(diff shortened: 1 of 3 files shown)


Tutez64:
Follow-up: `SecurityPanel` is now an enum abstract (same pattern as `StageAlign` / `TouchscreenType`), and `showSettings` takes `SecurityPanel` instead of `String`.
The first version used `String` constants. That compiled on native, but AIR uses the Flash externs where `showSettings` is `Null<flash.system.SecurityPanel>`, so `Security.showSettings(SecurityPanel.SETTINGS_MANAGER)` failed with
`String should be Null<flash.system.SecurityPanel>`.

PR #2887, opened 2026-08-23
Fix LocalConnection rest args and stub non-Windows native targets

Tutez64:
- Align `allowDomain` / `allowInsecureDomain` / `send` with AS3 rest-argument types.
- Keep Win32 named pipes on `cpp && windows`.
- On other native targets, provide a stub so `-D draft` compiles. `isSupported` remains false; `send` dispatches `StatusEvent` with `level` `"error"` (same as no receiver).


Tutez64:
Follow-up regarding the Win32 named-pipe implementation:
`Lib.notImplemented` is `notImplemented(?posInfo:PosInfos)`, so passing a `String` did not work. `send` also used `BytesBuffer` without importing it.

PR #2889, opened 2026-08-23
Default Stage.align to Flash's empty string (centered)

Tutez64:
Flash Player and AIR default `stage.align` to `""`, which centers content. OpenFL initialized `TOP_LEFT`, so `SHOW_ALL` (and similar scale modes) letterboxed to the top-left until callers assigned `stage.align = ""`.
  - Default `Stage.align` to `""` (native `null`, HTML5 empty string), matching that runtime behavior.
  - Treat `""` as centered on HTML5 the same way `null` already is on native.


Tutez64:
See https://github.com/openfl/lime/pull/2092 for the remaining AIR failure.

PR #2893, opened 2026-09-05
Filters: flash parity

m0rkeulv:
This PR contains the missing GradientBevel and  GradientGlow filters, and it replaces the blur filter with one that is much closer to what the flash player /Adobe Air would produce. there are now also software implementations of all filters.

It also fixes a couple of bugs/issues i've found along the way.

- hdpi scaling would affect the rendering result creating too small / weak effects, this  PR adds  multiplying with pixelRatio to get the correct the render size and intensity ( I've only tested this on windows 11, 125%, 150%, 175, 225%, scaling)

- There were issues with BitmapData.draw when using hardware rendering as a cached  hardware render would not be used by  BitmapData.draw and it resulted in a blank/empty result, so i have added a check to make sure the cache is the same as the renderer before use.

A small note about the filter image size and quality. increasing the quality for  blur based filters results  in larger render surfaces, i have with the help of Claude tried to determine how this works but have not been able to find any formula, after lots of measuring of flash renderings it just became simpler to hardcode the 16 possible values as a lookup table. it looks like this is what Ruffle also landed on so i guess maybe flash don't have a formula for this either. it starts off as a curve but then switches to something else after quality 10.

image

i don't know what the policy is for commenting code, i can probably remove them but they have been useful form me to  remember what's going on in a few places, i also renamed the blur filter to BoxBlur to make it easier to  see that its a box blur filer and not a gaussian blur without looking trough the shader code. i also did  this as  i at one point considered making it opt in with conditional compilations and defines, but  i dropped that since  i needed box blur for the other filters to render correctly.

with the help of Claude i got a report  generated where it captured the images of different targets and filter variations it should show how similar the filters now are to  Flash/Air.
https://claude.ai/code/artifact/120ed30d-3354-4123-815c-c2f4b0d978ba

ex.
image

One thing that is not part of this PR and that may  or may not slightly affect the results is that i also worked on this in a branch where i have added  FXAA super-sampling for native, and MSAA for web.

### DIFF
--- a/src/openfl/display/BitmapData.hx
+++ b/src/openfl/display/BitmapData.hx
@@ -382,12 +382,12 @@ class BitmapData implements IBitmapDrawable
 
 		if (filter.__preserveObject)
 		{
-			bitmapData3.copyPixels(this, rect, destPoint);
+			bitmapData3.copyPixels(sourceBitmapData, sourceRect, destPoint);
 		}
 
-		var lastBitmap = filter.__applyFilter(bitmapData2, this, sourceRect, destPoint);
+		var lastBitmap = filter.__applyFilter(bitmapData2, sourceBitmapData, sourceRect, destPoint);
 
-		if (filter.__preserveObject)
+		if (filter.__preserveObject && !filter.__softwareComposite)
 		{
 			lastBitmap.draw(bitmapData3, null, null);
 		}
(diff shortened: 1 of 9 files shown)


m0rkeulv:
just saw there where a pullrequest for  fixing something related to alpha channel blur pass  (https://github.com/openfl/openfl/pull/2872). 
As far as i can tell there should not be any issues with alpha channel blur in my PR.
| AIR | HashLink software | HashLink Hardware |
image

PR #2895, opened 2026-09-12
revert multi-rect batching in isCompatible(), fixes corrupted textures

mrchaoss1:
714751053 let `isCompatible()` accept multiple non-intersecting `DRAW_RECT`
commands per fill on the hardware path. But `DRAW_RECT`'s bitmap/shader-fill
rendering only ever handles one rect per fill - no per-rect UV mapping - so a
second rect samples the wrong part of the bitmap, causing corrupted/black
textures.

Reverts `isCompatible()` to the original one-shape-per-fill check.

Tested: two non-overlapping bitmap-filled rects rendered corrupted before,
correctly after.

Before:
image

After:
image

### DIFF
--- a/src/openfl/display/_internal/Context3DGraphics.hx
+++ b/src/openfl/display/_internal/Context3DGraphics.hx
@@ -41,7 +41,6 @@ class Context3DGraphics
 	private static var tempIndicesVector:Vector<Int> = new Vector<Int>();
 	private static var tempUvtVector:Vector<Float> = new Vector<Float>();
 	private static var tempScale9VerticesVector:Vector<Float>;
-	private static var tempRects:Array<Rectangle> = [];
 
 	private static function buildBuffer(graphics:Graphics, renderer:OpenGLRenderer):Void
 	{
@@ -514,31 +513,6 @@ class Context3DGraphics
 		Matrix.__pool.release(tileTransform);
 	}
 
-	private static function cleanupTempRects():Void
-	{
-		for (rect in tempRects)
-		{
-			Rectangle.__pool.release(rect);
-		}
-		#if haxe4
-		tempRects.resize(0);
-		#else
-		tempRects.splice(0, tempRects.length);
-		#end
-	}
-
-	private static function intersectsTempRects(rect:Rectangle):Bool
-	{
-		for (other in tempRects)
-		{
-			if (other.intersects(rect))
-			{
-				return true;
-			}
-		}
-		return false;
-	}
-
 	private static function isCompatible(graphics:Graphics):Bool
 	{
 		#if (openfl_force_sw_graphics || force_sw_graphics)
@@ -554,15 +528,9 @@ class Context3DGraphics
 
 		var data = new DrawCommandReader(graphics.__commands);
 		var hasColorFill = false, hasBitmapFill = false, hasShaderFill = false;
-
-		// for each fill, allow drawing only shapes with no intersection because
-		// that would require a cutout. fall back to software for intersections
-
-		// for drawRect(), remember the bounds of each rectangle
-		cleanupTempRects();
-
-		// drawTriangles() or drawQuads(): for simplicity, we'll allow only one
-		var hasDrawnComplex = false;
+		// only one filled shape per fill: shapes may intersect and need a cutout, and
+		// DRAW_RECT's bitmap fill only maps UVs for a single rect. fall back otherwise.
+		var hasDrawnFilledShape = false;
 
 		for (type in graphics.__commands.types)
 		{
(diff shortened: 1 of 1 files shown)


joshtynjala:
Instead of completely reverting, it should probably be incompatible with those types of fills only. Multiple rects of solid colors were working with that commit.

mrchaoss1:
I've been testing it by using AMD GPU, and that caused issues

joshtynjala:
@mrchaoss1 Does your project render correctly with OpenFL 9.5.2? Because that version of OpenFL (and many before it) already allowed multiple rectangles to be rendered with Context3D, whether they are intersecting or not. That's why I'm confused about your issue. Commit 7147510530500409192e998fca5f270b3dcd0174 should be restoring behavior that was temporarily disabled, but already worked properly for many years.

I tried reproducing with the following code that uses `beginBitmapFill()` with multiple `drawRect()` calls. However, it renders correctly on my machine.

```haxe
var grid = new BitmapData(10, 10, false, 0xff666666);
grid.fillRect(new Rectangle(0, 0, 5, 5), 0xff111111);
grid.fillRect(new Rectangle(5, 5, 5, 5), 0xff111111);

var sprite = new Sprite();
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 0, 20, 120);
sprite.graphics.drawRect(20, 0, 20, 120);
sprite.graphics.drawRect(40, 0, 20, 120);
sprite.graphics.drawRect(60, 0, 20, 120);
sprite.graphics.drawRect(80, 0, 20, 120);
sprite.graphics.drawRect(100, 0, 20, 120);
sprite.graphics.endFill();
sprite.x = 10;
sprite.y = 10;
addChild(sprite);
```

Can you share a simplified project that demonstrates the issue? Just enough to demonstrate the issue without anything extra that is not relevant to the issue. I'd like to fix it instead of reverting, so being able to reproduce the issue on my machine will be important. Thank you!

joshtynjala:
@mrchaoss1 Quick followup! Are you calling `disposeImage()` on your `BitmapData`? If OpenFL needs to draw in software because it encounters intersecting rectangles, this may be the cause of your black rectangles.

I can reproduce black rectangles with the following code:

```haxe
var grid = new BitmapData(10, 10, false, 0xff666666);
grid.fillRect(new Rectangle(0, 0, 5, 5), 0xff111111);
grid.fillRect(new Rectangle(5, 5, 5, 5), 0xff111111);
grid.disposeImage();

var sprite = new Sprite();
sprite.graphics.beginBitmapFill(grid);
// these rectangles intersect, so they must be drawn in software
sprite.graphics.drawRect(0, 0, 120, 20);
sprite.graphics.drawRect(0, 0, 20, 120);
sprite.graphics.drawRect(0, 100, 120, 20);
sprite.graphics.drawRect(100, 0, 20, 120);
sprite.graphics.endFill();
sprite.x = 10;
sprite.y = 10;
addChild(sprite);
```


If I comment out the `disposeImage()` call, it renders correctly.

```haxe
// grid.disposeImage();
```

I should mention that OpenFL is behaving correctly in my code sample above. It _should_ display black rectangles because the underlying `Image` has been disposed, but it needs to draw in software due to the intersections of the rectangles. Since the texture is hardware-only, it falls back to the solid black color.

joshtynjala:
It may also be worth trying the `openfl_force_hw_graphics` define, if you don't want to remove calls to `disposeImage()`. Not all vector shapes can be drawn in hardware, but if you're drawing bitmaps to rectangles only, that limitation shouldn't affect you.

mrchaoss1:
I've checked them and here's what I got:

- Keeping bitmaps readable: fixes it, but memory goes ~200MB -> ~3GB (we
  free CPU pixels after GPU upload on purpose) and some shaders break.
- openfl_force_hw_graphics: also fixes it, bitmaps stay GPU-only this time,
  but memory still goes up to ~1.3-2GB and shaders still break.

Both point at software fallback as the trigger like you said, but neither
works for us without new regressions. That commit fixes it, untill we dont find an idea how to fix it

joshtynjala:
> openfl_force_hw_graphics: also fixes it, bitmaps stay GPU-only this time,
but memory still goes up to ~1.3-2GB and shaders still break.

Are you sure that you remembered to add your `disposeImage()` calls back in when you tested the `openfl_force_hw_graphics` define? This define basically just forces `Context3DGraphics.isCompatible()` to always return `true`. It ensures that _nothing_ gets drawn in software, and forces it all to be drawn in hardware. I would not expect such a major increase in memory.

joshtynjala:
Another possible fix for your project is separating your drawRect() calls into multiple fills.

Using my example code from above, instead of something like this:

```haxe
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 0, 120, 20);
sprite.graphics.drawRect(0, 0, 20, 120);
sprite.graphics.drawRect(0, 100, 120, 20);
sprite.graphics.drawRect(100, 0, 20, 120);
sprite.graphics.endFill();
```

You can do this:

```haxe
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 0, 120, 20);
sprite.graphics.endFill();
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 0, 20, 120);
sprite.graphics.endFill();
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 100, 120, 20);
sprite.graphics.endFill();
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(100, 0, 20, 120);
sprite.graphics.endFill();
```

Or you could include some rects together in a single fill, as long as you know that they won't intersect.

```haxe
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 0, 120, 20);
sprite.graphics.drawRect(0, 100, 120, 20);
sprite.graphics.endFill();
sprite.graphics.beginBitmapFill(grid);
sprite.graphics.drawRect(0, 0, 20, 120);
sprite.graphics.drawRect(100, 0, 20, 120);
sprite.graphics.endFill();
```

joshtynjala:
@mrchaoss1 There is one thing that I'm not quite understanding about your PR. You revert commit 7147510530500409192e998fca5f270b3dcd0174, but you appear to restore the changes from commit 238fa39928fef8d169f0e8653ea7675ac1910ff4. That commit also forces falling back to software, but it does that every time that there is more than one `drawRect()` call in a single fill. Even if they don't intersect. It should be _more likely_ to cause your black rectangles from disposed images. I would have expected you to need to revert both commits 7147510530500409192e998fca5f270b3dcd0174 and 238fa39928fef8d169f0e8653ea7675ac1910ff4 to prevent the black rectangles.

